PR #88, opened 2021-03-05
chore: improve warning message

posva:
I think the message can have markdown but I am not sure.


johnsoncodehk:
As I know markdown not working here. Look good, thanks!

PR #149, opened 2021-04-22
fix: don't complete attribute value for `v-else`

g-plane:
Before:

https://user-images.githubusercontent.com/17216317/115713186-4dcc4d00-a3a8-11eb-9751-928f8bb3b068.mp4

After:

https://user-images.githubusercontent.com/17216317/115713221-5886e200-a3a8-11eb-87cf-a809ddaa3c60.mp4


johnsoncodehk:
Also maked some related changes. Works good, thank you!

PR #151, opened 2021-04-23
fix: don't complete attribute value for `setup`

g-plane:


johnsoncodehk:
Look good, thank you!

PR #154, opened 2021-04-24
feat: support complete attribute without v-bind

g-plane:
This is useful and common for those attributes can accept static string as values.

图片


johnsoncodehk:
Thanks to try the improvement! But this is UX design, I want avoid too many items in completion list, user can obviously to see all valid props.

螢幕截圖 2021-04-24 下午1 17 37

g-plane:
Is there another way? When I use a attribute with static string as value, complete that item from list then I must move my cursor back and delete the colon, which is so inconvenient.

johnsoncodehk:
Merged with some patching.

PR #155, opened 2021-04-24
feat: support adding breakpoints (close #107)

### LINKED ISSUE #107: Support for debugging in .vue
I want to use 'Debugger with for Filefox', but I can't add breakpoints.

g-plane:
Closes #107 

图片

@odex21 There may be a bug from the "Debugger for Firefox" VS Code extension, and adding breakpoints may not have effects, so you may need "Debugger for Chrome" VS Code extension and use Chrome to debug your code.

### DIFF
--- a/package.json
+++ b/package.json
@@ -36,6 +36,11 @@
 				"configuration": "./languages/vue-language-configuration.json"
 			}
 		],
+		"breakpoints": [
+			{
+				"language": "vue"
+			}
+		],
 		"configuration": {
 			"type": "object",
 			"title": "Volar",


johnsoncodehk:
I'm not sure how vscode breakpoint working, I thought we need to use `getBreakpointStatementAtPosition` api from typescript.

I will check it later, thank you!

johnsoncodehk:
A problem is when I add a breakpoint before start debugger, the breakpoint is change to unexpected line.
if I add a breakpoint after started debugger it's works good.

But I think it may a compiler source map problem not IDE plugin problem, merge it.

Issue #198, opened 2021-05-22
[Bug] File icons are emptied when importing


wenfangdu:
When importing, file icons are emptied when selected across:

![volar](https://user-images.githubusercontent.com/28700378/119218590-b4fa3000-bb13-11eb-8dc7-029d74380266.gif)

PR #210, opened 2021-05-26
fix: 🐛 https://github.com/johnsoncodehk/volar/issues/198

IWANABETHATGUY:
✅ Closes: #198

details lost extension after resolved by ts-service, only happend when resolve `tsx` and ts
image

after this pr: 

https://user-images.githubusercontent.com/17974631/119636543-52709f00-be47-11eb-82c1-d538afc9cc93.mp4


johnsoncodehk:
Thanks to check the reason! We can't simply replace detail, otherwise resolved detail message will be missing.

PR #238, opened 2021-06-01
feat: 🎸 finish 235

### LINKED ISSUE #235: [FR] Setting to add file extension for `.js`/`.jsx` files in import path intellisense
Currently, no file extension is shown in the path intellisense for `.js`/`.jsx` files like `.vue` files do, would be great if we have the option to enable that.

Please note: the desired behavior should be completing what's shown in the intellisense, e.g. `Foo.ts`/`Foo.tsx` being shown as `Foo`, and completed as `Foo`, `Foo.js`/`Foo.jsx` being shown as they are, and completed as they are.

image

image

Repro: https://github.com/wenfangdu/volar-jsx-import-vue-repro

IWANABETHATGUY:
resolved #235

### DIFF
--- a/package.json
+++ b/package.json
@@ -104,6 +104,11 @@
 					"default": false,
 					"description": "Auto-complete Ref value with `.value`. (Experimental)"
 				},
+				"volar.completion.alwaysShowFileExtension": {
+					"type": "boolean",
+					"default": false,
+					"description": "always show file extension of path completion"
+				},
 				"volar.tsPlugin": {
 					"type": "boolean",
 					"enum": [
--- a/packages/vscode-server/src/configs.ts
+++ b/packages/vscode-server/src/configs.ts
@@ -1,4 +1,4 @@
-import { codeLensOptions } from 'vscode-vue-languageservice';
+import { codeLensOptions, completionOptions } from 'vscode-vue-languageservice';
 import type * as emmet from 'vscode-emmet-helper';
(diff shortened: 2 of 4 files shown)


IWANABETHATGUY:
https://user-images.githubusercontent.com/17974631/120356657-279ab500-c337-11eb-8803-39bfd55e2d40.mp4

wenfangdu:
Thanks for the PR! However, for `.ts` and `.tsx` files, I think this setting should bypass them since when doing so, the following will happen:

image

image

IWANABETHATGUY:
https://user-images.githubusercontent.com/17974631/120433897-5acc5b00-c3ae-11eb-9d98-5b3ef61f85f0.mp4

this pr doesn't complete the extension, just add the extension to the label.
cc @johnsoncodehk

wenfangdu:
Oh, sorry, didn't notice that. The desired behavior should be completing what's shown in the intellisense, e.g. `Foo.tsx` being shown as `Foo`, and completed as `Foo`, `Foo.js`/`Foo.jsx` being shown as they are, and completed as they are.

johnsoncodehk:
Thanks for the PR!

But TS already has `typescript.preferences.importModuleSpecifierEnding` setting and I expect to reuse it. I will check how [typescript-language-features](https://github.com/microsoft/vscode/tree/main/extensions/typescript-language-features) handle all the TS settings and try to make same achieve in volar. (Maybe do it in the next weekend)

PR #281, opened 2021-06-25
Fix the error of asynchronous component

yangdan8:


johnsoncodehk:
See https://github.com/johnsoncodehk/volar/issues/280#issuecomment-868271203.

yangdan8:
Currently, the 0.25.24 that you have modified and released causes the template error to be unrecognized

johnsoncodehk:
See https://github.com/johnsoncodehk/volar/issues/280#issuecomment-871852424.

PR #379, opened 2021-08-17
feat: support sfc named css modules

fnlctrl:
There doesn't seem to be existing tests, so none was added. I don't know how to test with a dev extension yet so this might not work.
Only saw the rewrite warning after finishing the PR 😂  Well, hope it at least serves as a reminder for this feature


johnsoncodehk:
Made some modify, thanks!

Issue #378, opened 2021-08-17
useSetupSugar in vue3.2  error  Cannot access '__default__' before initialization


klwfwdk:
in doc https://v3.vuejs.org/api/sfc-script-setup.html#usage-alongside-normal-script
the script setup should below normal script tag. if not, the error will log in console
`Cannot access '__default__' before initialization`

PR #380, opened 2021-08-17
fix: useSetupSugar blocks order error (#378)

klwfwdk:
Adjusted the code block insertion order generated by useSetupSugar. Now the block like `<script setup>` is below the `<script>`


johnsoncodehk:
Merge to consistent with document, but blocks order should not throw error, please report the problem to upstream, thanks.

Close #378

PR #395, opened 2021-08-20
chore: sync previewer.ts with vscode

Amour1688:
https://github.com/microsoft/vscode/blob/main/extensions/typescript-language-features/src/utils/previewer.ts#L170


johnsoncodehk:
LGTM, why closed it? 🤔

PR #401, opened 2021-08-21
[proof of concept] Make typescript optional in initializationOptions

predragnikolic:
> Note: This is a proof of concept. This PR is just meant to give some ideas, 
so feel free to push commits to this PR or create a new PR that is better. :)

This PR makes the `typescript` option optional in `initializationOptions`. 

**Why?**
Each lsp clients need to reinvent the way of how to find tsserver path. 
By moving that logic to the server, clients can get rid of that code.
Of course someone can still pass in the tsserver path in the `initializationOptions` if there is a need for that. (this change should be backwards compatible)

Inspiration taken from theia-typescript-server: 
https://github.com/typescript-language-server/typescript-language-server/blob/3c457111969899d7cd72291db1e6dddb5ac68cf6/src/lsp-server.ts#L68

Except that theia-typescript-server is more bullet proof, and it will try to:
- load tsserver if yarn pnp is used
- else load tsserver in workspace folder
- else load the globally installed tsserver
- else load tsserver bundled with a extension

This PR only adds:
- load tsserver in workspace folder
- else load tsserver bundled with a extension

### DIFF
--- a/packages/server/src/index.ts
+++ b/packages/server/src/index.ts
@@ -5,6 +5,7 @@ import * as vscode from 'vscode-languageserver/node';
 import { updateConfigs } from './configs';
 import { createServicesManager } from './servicesManager';
 import * as tsConfigs from './tsConfigs';
+import * as path from 'path';
 
 const connection = vscode.createConnection(vscode.ProposedFeatures.all);
 const documents = new vscode.TextDocuments(TextDocument);
@@ -58,18 +59,18 @@ function onInitialize(params: vscode.InitializeParams) {
 async function onInitialized() {
 
 	connection.onRequest(shared.PingRequest.type, () => 'pong' as const);
-
+	const loadedTs = loadTs()
 	if (options.languageFeatures) {
 		const servicesManager = createServicesManager(
 			options,
-			loadTs,
+			loadedTs,
 			connection,
 			documents,
 			folders,
 		);
 
 		(await import('./features/customFeatures')).register(connection, documents, servicesManager);
-		(await import('./features/languageFeatures')).register(loadTs().server, connection, documents, servicesManager, options.languageFeatures);
+		(await import('./features/languageFeatures')).register(loadedTs.server, connection, documents, servicesManager, options.languageFeatures);
 		(await import('./registers/registerlanguageFeatures')).register(connection, options.languageFeatures, vue.getSemanticTokenLegend());
 
 		connection.client.register(vscode.DidChangeConfigurationNotification.type, undefined);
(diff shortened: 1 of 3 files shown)


packages/server/src/index.ts
```diff
@@ -99,8 +100,39 @@ async function onInitialized() {
...
+
+function findTsserverPath(): string {
+	if (options.typescript?.serverPath) {
+		return options.typescript.serverPath;
 	}
+
+	const tsServerPath = path.join('typescript', 'lib', 'tsserver.js');
+
+	// look into node_modules of workspace root
+	let executable = findPathToModule(folders[0], tsServerPath);
```
predragnikolic, line 119: I used the first workspace folder `folders[0]` and I know that this is not right.
But I do not know how to handle multiple workspace folders... 

Any suggestions are welcome. :)

johnsoncodehk:
Thanks for trying to improve, but for some reasons, not automatically using the typescript module of workspace is an intentional behavior:

1. Considering the bundle size, plugin should use IDE TS module first. If the IDE does not have a TS module, the plugin should bundle its own TS module.

> Each lsp clients need to reinvent the way of how to find tsserver path.

Since the method of extracting TS is different for each IDE, each LSP client must be re-invented.

2. Without the user's knowledge, the user may have different TS behaviors because of installing different TS versions in different projects. By default, we expect users to have consistent TS behavior in the same IDE and plugin version, unless users want to select a specific TS version for individual projects.

3. We cannot confirm whether the node_modules directory of the workspace is downloaded from a reliable source. If it is not from a reliable source, node_modules may contain malicious code. We should never automatically load the workspace code without the user's consent, otherwise this is a security hole. (This is why VSCode has Workspace Trust)

predragnikolic:
> Since the method of extracting TS is different for each IDE, each LSP client must be re-invented.

I don't believe that the way of extracting TS should be different for each IDE.

I would prefer convention over configuration. 
For example, try to use the most sane default option that will cover 90% of use-cases. (in this case try to use the tsserver from node modules of the project, but because of security issues, maybe this is not the most sane default option)
And if that is not enough, provide that option to be configurable. (fallback to the bundled tsserver).

> We cannot confirm whether the node_modules directory of the workspace is downloaded from a reliable source. If it is not from a reliable source, node_modules may contain malicious code. 

While `Workspace Trust` will prevent code from running automatically.

The end user is still the one who will say `"yes, this is safe"`, or `"no, this is not safe"`. 
I am a human, not a malware detection program. :)
Most likely I would initially say `"no, this is not safe"` and later change my mind to `"yes, this is safe"`, but that still doesn't change the fact that maybe there is a npm dependency that is not safe in that project, or that a npm dependency will get a malware added in a future release.

But yes, security is a problem in node.js.

predragnikolic:
I am going to close this PR.

In favor of implementing the to finding of the tsserver on the client, in python, insecurely. :)

johnsoncodehk:
I mean if the user clicks "yes, this is safe", or the user config to use the workspace module, and the user is attacked, this is the user's responsibility. But if the user just downloads a project from an unknown source and is attacked when the extension automatically loads the project module when opened in the IDE, this is the responsibility of the IDE and the extension.

I think the method of extracting TS from IDE is different for each IDE, but we already have a common method for extracting TS from webspace: https://github.com/johnsoncodehk/volar/blob/master/packages/shared/src/ts.ts#L25-L40

Use `getWorkspaceTypescriptPath('node_modules/typescript/lib', [WORKSPACE_FOLDER])`.

Issue #334, opened 2021-07-27
Use Inline Completions instead of Snippet for `.value` auto-complete


johnsoncodehk:
Issue: https://github.com/microsoft/vscode/issues/124024
Sample: https://github.com/microsoft/vscode-extension-samples/tree/main/inline-completions

cah-jamesarnold:
Will this solve the annoyance of `.value` appending to the end of every ref type? I honestly had to uninstall Volar because it was driving me insane fighting the auto complete for every variable.

johnsoncodehk:
@cah-jamesarnold I think yes, and you can disable it. See https://github.com/johnsoncodehk/volar/discussions/324.

cah-jamesarnold:
> @cah-jamesarnold I think yes, and you can disable it. See #324.

Thank you!

PR #405, opened 2021-08-22
fix: support isRefType for @vue/composition-api, close #404

### LINKED ISSUE #404: Support unwrapping Ref<T> from @vue/composition-api in template
Repro: https://github.com/antfu/repro-volar-vue2-template-unref

image

I recently made [`vue2-script-setup-transform`](https://github.com/antfu/vue2-script-setup-transform) which enables `<script setup>` in Vue 2 apps. When using with Volar, it get almost everything working perfect despite this one.

As I can see that Volar is unwrapping the type from Vue 3's, while `@vue/composition-api` has another. Is that possible to check the present of `@vue/composition-api` deps and add an additional check when it's the case? I believe this will make many Vue 2 developers start switching to Volar.

Thanks

antfu:
Not sure if it's the fix, just browse the code and thought this might be the point (not tested). Feel free to close if it's wrong.

`@vue/composition-api` use `dist/index.d.ts` for the file name, and I have changed it to `dist/vue-composition-api.d.ts` since v1.1.3 for this.


johnsoncodehk:
There only affect the `.value` auto-complete feature (#334). 😂

It is true that this feature does not support `@vue/composition-api` currently. I have not test this modify yet, will check it later.

PR #431, opened 2021-08-28
chore: Add "end" offset to shared.SfcBlock

yaegassy:
Could you add an offset for `end: number` to the SfcBlock interface?

I would like to use "shared.SfcBlock" etc. in coc-volar's `volar.action.splitEditor`, but I would also like an offset for "end".


johnsoncodehk:
LGTM, thanks!

PR #638, opened 2021-10-27
feat: improve pug auto-complete

### LINKED ISSUE #618: no component prop suggestions in Pug .vue template
## Details
I've recently moved from Vetur to Volar for better Pug support and it's working great for me except for the lack of intellisense suggestions on component props. 

I've noticed that in normal html mode volar will pick up the available props and it's visible in intellisense.
image

However when working with the same component in pug you will get no suggestions.
image

If you know the name and start typing it, then intellisense starts to kick in, but only if you already know what you are looking for.
image

Interestingly, when you manually add the prop and hover it, you can see that the prop/type information is properly being passed to the IDE.
image

## Reproduction
I created a fresh Quasar project using the Quasar cli create command but I assume this issue would be visible in any vue project.

## Versions
Volar v0.28.7
Vue 3

johnsoncodehk:
Close #618


jdheeter:
Works great for me! 🙌

Issue #741, opened 2021-11-28
Vue instance type extending in Vue 2 project does not work


rchl:
The types declared in the `types` array in `tsconfig.json` don't extend the Vue instance as they supposed to.

Here is a simple Vue 2 project with enabled Volar support: https://github.com/rchl/volar-vue2-test

The type of the `context` argument in `pages/index.vue` should be a `Context` and not `any`.

![Screenshot 2021-11-28 at 12 37 50](https://user-images.githubusercontent.com/153197/143766104-41504ce0-2b55-4f96-8067-dfe8ca43a106.png)

 It works with Vetur.

sethidden:
I guess this is a similar issue to https://github.com/nuxt/typescript/issues/83#issuecomment-524563103, where @nuxt/types adds the types correctly to the Vue 2 component types ([here](https://github.com/nuxt/nuxt.js/blob/dev/packages/types/app/vue.d.ts#L30) but that Vue 2 component type is not being used. I assume Volar needs to use the Vue 3-ish (hence dependency on @vue/runtime-dom) interface which @nuxt/types does not inject to

rchl:
Sounds plausible. But whatever the outcome, it would be good to document the situation because as it is now, it's not really practical to use Volar for Vue 2 projects. I wonder if this is something that can and will be addressed or not really. In the latter case, there is Vetur still, and Volar should probably not bother with Vue 2.

johnsoncodehk:
This is by design. vetur internally wraps the options in `Vue.extends`, but I think predictability is more important than out-of-the-box. Your code should be same behavior in `<script>` and *.ts, so volar avoids hacking `<script>` block.

You can change it very easily to get the behavior of vetur.
```diff
+ import Vue from 'vue';

- export default {
+ export default Vue.extend({
  async asyncData(context) {
    context.error({message: 'err'})
  }
- }
+ )}
```

rchl:
What sort of serious project that relies on type checking would not want that? When using any external Vue plugin or just Nuxt, which itself extends Vue a lot, basically requires that or otherwise the user will see type errors everywhere.

If you insist on keeping this behavior for some reason, then it would be at least document this behavior in appropriate section in the documentation.

Also, the documentation states:

> Template type-checking do not support with Vue.extend, you can use composition-api, vue-class-component, or export default { ... } instead of export default Vue.extend.

So does that mean that your suggestion will also make type checking in the template cease to work?

johnsoncodehk:
[...]

Issue #21, opened 2020-11-10
Methods not found when using mapGetters


LeanderD:
First I would like to say that I like this extension. It's working well!

In my code I'm using mapGetters to well.. map the getters from my Vuex store.
Unfortunately, this is not working with Volar. It says it can't find these methods.

A "solution" would be to not use the mapGetters of course, but to me that's not really an option.
Hope this can be fixed!

johnsoncodehk:
Thanks for your attention and report! I didn't use vuex, but I searched and tested projects using mapGetters from github, and it seems to work. Can you give me a reproduction project?

LeanderD:
I am using [vue-class-component](https://github.com/vuejs/vue-class-component/). So that probably has something to do with it. 
The project is also set up as a mono-repo. I did have some issues with that before (with other extensions and resolving paths etc), but those have been resolved.

A simple component looks like this:
```
<template>
    <div>{{ getSomething(param) }}</div>
</template>

<script lang="ts">
import { Options, Vue } from 'vue-class-component';
import { mapGetters } from 'vuex';

@Options({
    computed: {
        ...mapGetters(['getSomething']),
    },
})
export default class SimpleComponent extends Vue {}
`

In this case it would tell me that it can't find `getSomething` (the one in the template).
It will build and work.. it just shows me an error in the editor.

Hope this is clear enough. Thanks for making this amazing extension!

johnsoncodehk:
I am not familiar with vue-class-component. I asked someone familiar with it. The problem is not from mapGetters, but because @Options decorator has no effect on the type. Decorator is runtime only. You need to use mixins instead.
I'm not sure if this is the best solution, so I suggest you propose it here: https://github.com/vuejs/vue-class-component/issues

LeanderD:
Thank you for your answer! That does make sense. I will take a look at it!

johnsoncodehk:
Thanks for the update!

PR #750, opened 2021-11-30
fix(docs): extend vue 2 info regarding types

rchl:
I know this puts the instructions upside-down since you've been suggesting to remove `Vue.extend` while I've changed it to recommend `Vue.extend`. The rationale is that, based on testing and as stated in the comments in #741, `Vue.extend` is really the way to go to fix the inevitable type issues in Vue 2 projects and I haven't really found any downsides to it.

If you have any specific examples where `Vue.extend` would makes things worse type-wise then I can tweak the wording or change it completely.

Note that when working with Vetur, I'm used to having to help the type system in many situations by adding manual JSDoc comments (when using JS) so I would expect this to be the case with Volar also. But that shouldn't change the fact that `Vue.extend` is required in the first place.

Related #741


johnsoncodehk:
If you don't care about the component props / events type-check and auto-complete in template, `Vue.extend` is fine.

Otherwise you need to migrate to defineComponent from `@vue/composition-api` to support these features.

rchl:
> If you don't care about the component props / events type-check and auto-complete in template,

I'm trying to say that this already **works**. Check the simple https://github.com/rchl/volar-vue2-test project for example.

I know that using Composition API works better in terms of type support but I just can't understand why do you keep saying that `Vue.extend` breaks auto-complete in the template. This is simply not true. I'd need you to explain this clearly before I adjust the text.

johnsoncodehk:
Sorry for not making it clear, I am talking about cross-component props auto-complete and type-checking.

It should have auto-complete for `:test`.
螢幕截圖 2021-12-04 上午2 15 18

It should have type error for `:test`.
螢幕截圖 2021-12-04 上午2 16 05

rchl:
But that's not relevant to whether someone is using `Vue.extend` or not.

And even so, it works in the my test project above any way. Try adding a `test` property to `Tutorial` component in `pages/index.vue`.

johnsoncodehk:
Without `Vue.extend` it should works. But actually I planned to abandon support for cross-component props type-checking with `export default { ... }`, so I hope remove mentioned `export default { ... }` in this pr.

image

So user only have 3 option in future:
- `export default Vue.extend({ ... })`: support type-check in `<script>` and `<template>`, but not support cross-component type-check.
- `export default defineComponent({ ... })`: support type-check in `<script>` and `<template>` and with cross-component
- class component: support type-check in `<script>` and `<template>` and with cross-component with additional code. (#21)

rchl:
Seems to be working with `Vue.extend` also.

Screenshot 2021-12-03 at 20 27 21

So to conclude:
 - cross-component props type-checking seems to be working with both `Vue.extend` and without
 - Type-checking within the component requires `Vue.extend` to work properly.

With that said, is there anything you want to change me in this PR?

johnsoncodehk:
@rchl emmmm it's vetur report. :S

rchl:
🤦  Sorry. That's my default setup and forgot to switch for testing.

But now that I've switched to Volar, it doesn't seem to work with either `export default {}` or export default Vue.extend({})`.

Screenshot 2021-12-03 at 20 33 34

johnsoncodehk:
Maybe `Tutorial.vue` is still `Vue.extend({ ... }) `.

rchl:
No, you can try yourself on the project. Pushed Vue.extend removal

johnsoncodehk:
Sorry, also need `lang="ts"`.

rchl:
OK, can reproduce that case.

That complicates things since now we have more variables to potentially document.

As far as using plan JS, `Vue.extend` still seems like a preferred option.

As far as TS, are you saying that it will work only with `Vue.extend` (based on [your comment](https://github.com/johnsoncodehk/volar/pull/750#issuecomment-985755033)? In which case, doesn't my point still stands that using `Vue.extend` is recommended?

johnsoncodehk:
I think we don’t need to advise users what to use, just state the facts, which methods are supported and which are not.

|                                              | Component options type-checking in `<script>`                | Interpolation type-checking in `<template>` | Cross-component props type-checking     |
|----------------------------------------------|--------------------------------------------------------------|---------------------------------------------|-----------------------------------------|
| `export default { ... }` with JS             | Not Support                                                  | Not support                                 | Not support                             |
| `export default { ... }` with TS             | Not Support                                                  | Support but planed to abandon in future     | Support but planed to abandon in future |
| `export default Vue.extend({ ... })` with JS | Not Support                                                  | Not support                                 | Not support                             |
| `export default Vue.extend({ ... })` with TS | Limited (Support `data` types but not support `props` types) | Not support                                 | Not support                             |
| `export default defineComponent({ ... })`    | Support                                                      | Support                                     | Support                                 |
| Class component                              | Support                                                      | Support with additional code (#21)          | Support with additional code            |

rchl:
Is this a typo?

Screenshot 2021-12-05 at 20 35 08

`Vue.extend` is exactly what is supposed to make those work.

johnsoncodehk:
`Vue.extend` only working with js but not working with ts on my tests.

You can try:

```js
// test-js.js
import Vue from 'vue'

Vue.extend({
	data: () => ({
		foo: 123
	}),
	mounted() {
		this.foo // any
	},
})
```

```ts
// test-ts.ts
import Vue from 'vue'

Vue.extend({
	data: () => ({
		foo: 123
	}),
	mounted() {
		this.foo // number
	},
})
```

rchl:
I would swear that JS was working before but it doesn't seem like it. Maybe I was confused due to my Vetur running again...

In that case I would say that Options API Vue2 support is not in a state that is usable in a real world projects so maybe it shouldn't be supported/advertised at all?

Having to add `Vue.extend` everywhere is one major obstacle. I don't think that it's feasible to do that everywhere in an existing, potentially large projects.

The other issue is that type checking is spotty at best so many of the issues wouldn't be caught. Vetur is not ideal either, mind you, but certainly more complete.

johnsoncodehk:
> I would swear that JS was working before but it doesn't seem like it. Maybe I was confused due to my Vetur running again...

I think this has not relate to vetur/volar because it is reproduced in vanilla js/ts.

> In that case I would say that Options API Vue2 support is not in a state that is usable in a real world projects so maybe it shouldn't be supported/advertised at all?

With defineComponent from `@vue/composition-api` I think the support for vue2 is already quite good.

> Having to add `Vue.extend` everywhere is one major obstacle. I don't think that it's feasible to do that everywhere in an existing, potentially large projects.

Users only need to continue to use vetur. If they don't want to modify the code, they don't have to switch to volar.
btw the method I recommend is defineComponent instead of Vue.extend.

> The other issue is that type checking is spotty at best so many of the issues wouldn't be caught. Vetur is not ideal either, mind you, but certainly more complete.

Are you talking about using Vue.extend? If yes, it is because vetur has done some work outside of Vue.extend to support more types. But like I said, doing so destroys predictability, so Volar won't do it. The component type will be completely controlled by the vue library. If Vue.extend is missing the props type, the user should not get the props type in the template, otherwise it is a bug for volar.

rchl:
> With defineComponent from `@vue/composition-api` I think the support for vue2 is already quite good.

But I'm talking about the Options API specifically. Can `definedComponent` be used with Options API? If so then I'm fine with just including that info the readme (the table you've posted).

johnsoncodehk:
> Can `definedComponent` be used with Options API?

Yes, it is comprehensively better than Vue.extend.

yangHeavy:
> I think we don’t need to advise users what to use, just state the facts, which methods are supported and which are not.
> 
> Component options type-checking in `<script>`	Interpolation type-checking in `<template>`	Cross-component props type-checking
> `export default { ... }` with JS	Not support	Not support	Not support
> `export default { ... }` with TS	Not support	Support but planed to abandon in future	Support but planed to abandon in future
> `export default Vue.extend({ ... })` with JS	Not support	Not support	Not support
> `export default Vue.extend({ ... })` with TS	Limited (Support `data` types but not support `props` types)	Limited	Not support
> `export default defineComponent({ ... })`	Support	Support	Support
> Class component	Support	Support with additional code (#21)	Support with additional code

hello, i have seen this property table for prop type check.
First, i am Very exciting volar supports the vue2.0 type check.
But what about `vue-property-decorator` use `@prop` ? Can it have the type check and auto-complete? It seems belong to class component type.
See the example in this repository: [https://github.com/MAG-Christian-Birkl/volar-vue2-test-mag](https://github.com/MAG-Christian-Birkl/volar-vue2-test-mag)
Although I have add `static components: { HelloWorld: typeof HelloWorld }`, it still failed with type check.

johnsoncodehk:
@yangHeavy it's a example what you can define in class component for volar.

```ts
export default class Foo extends Vue {
  static components: {
    MyComp: typeof MyComp
  }
  $props!: {
    optionalProp?: string;
    requiredProp: string;
  }
  $emit!: {
    (e: 'my-event-1'): void
    (e: 'my-event-2', arg1: string, arg2: number): void
  }
  $scopedSlots!: { // $scopedSlots for vue2, $slots for vue3
    'my-slot-1': () => VNode[]
    'my-slot-2': (bindings: { arg1: string, arg2: number }) => VNode[]
  }
}
```

---

In App.vue of the repo, you need to define `static components: { HelloWorld: typeof HelloWorld }` in the class component.

yangHeavy:
Thanks for answer for how to use class component with volar. But there some problems too.
1. I have to use declarations, otherwise I get an error

2. so `vue-property-decorator` use `@prop` can't  check?

johnsoncodehk:
@yangHeavy yes you need to define it repeatedly. decorator == no TS support :)

yangHeavy:
I see. But there are some problems when I have defined the variable. 
I define it string, use it with number, but volar doesn't get an error.
image

johnsoncodehk:
@yangHeavy in this case you need to define `$props!: { what?: string }` additionally.

yangHeavy:
@johnsoncodehk  Sorry, I found it. I try it success!! I will share Volar with my colleagues and weigh whether we need to use variable type checking in cases where we need to define it twice. Thank you very much.

johnsoncodehk:
@yangHeavy as a workaround, you could shim the component type to auto define `$props`.

In this workaround, you need to write `readonly` additionally to distinguish this is a prop not component data.

```ts
@Component({})
class HelloWorld extends Vue {
  @Prop() readonly requiredProp!: string;
  @Prop() readonly optionalProp?: string;
  nonProp?: string;
}

export default HelloWorld as ShimComponentProps<typeof HelloWorld>;
```

helper types:
```ts
type ShimComponentProps<T extends new (...args: any) => any> = new () => InstanceType<T> & {
  $props: ComponentProps<InstanceType<T>>
};

type ComponentProps<T> =
  { [K in keyof Pick<T, ReadonlyKeys<T> & OptionalKeys<T>>]?: T[K] }
  & { [K in keyof Pick<T, ReadonlyKeys<T> & RequiredKeys<T>>]: T[K] };

type Equal<X, Y> = (<T>() => T extends X ? 1 : 2) extends <T>() => T extends Y ? 1 : 2 ? true : false;

type ReadonlyKeys<
  T,
  U extends Readonly<T> = Readonly<T>,
  K extends keyof T = keyof T
  > = K extends keyof T ? Equal<Pick<T, K>, Pick<U, K>> extends true ? K : never : never;

type OptionalKeys<T> = keyof {
  [K in keyof T as T[K] extends Required<T>[K] ? never : K]: T[K]
};

type RequiredKeys<T> = keyof {
  [K in keyof T as T[K] extends Required<T>[K] ? K : never]: T[K]
};
```

yangHeavy:
@johnsoncodehk `ShimComponentProps` is a nice typescript coding example! But it still has some problems after i test it.
image
`ShimComponentProps` doesn't extends Vue! And this will cause @Components() can't identify `HelloWorld`.
image

rchl:
I've applied your suggestion with some minor changes. Let me know if it looks better now.

johnsoncodehk:
LGTM, thank you!

hexf00:
> @yangHeavy it's a example what you can define in class component for volar.
> 
> ```ts
> export default class Foo extends Vue {
>   static components: {
>     MyComp: typeof MyComp
>   }
>   $props!: {
>     optionalProp?: string;
>     requiredProp: string;
>   }
>   $emit!: {
>     (e: 'my-event-1'): void
>     (e: 'my-event-2', arg1: string, arg2: number): void
>   }
>   $scopedSlots!: { // $scopedSlots for vue2, $slots for vue3
>     'my-slot-1': () => VNode[]
>     'my-slot-2': (bindings: { arg1: string, arg2: number }) => VNode[]
>   }
> }
> ```
In vue2 project, the type of scope parameter is deduced as vnode [] | undefined, I want to know what happened in the middle, or does it not have this feature?

hexf00:
> > @yangHeavy it's a example what you can define in class component for volar.
> > ```ts
> > export default class Foo extends Vue {
> >   static components: {
> >     MyComp: typeof MyComp
> >   }
> >   $props!: {
> >     optionalProp?: string;
> >     requiredProp: string;
> >   }
> >   $emit!: {
> >     (e: 'my-event-1'): void
> >     (e: 'my-event-2', arg1: string, arg2: number): void
> >   }
> >   $scopedSlots!: { // $scopedSlots for vue2, $slots for vue3
> >     'my-slot-1': () => VNode[]
> >     'my-slot-2': (bindings: { arg1: string, arg2: number }) => VNode[]
> >   }
> > }
> > ```
> 
> In vue2 project, the type of scope parameter is deduced as vnode [] | undefined, I want to know what happened in the middle, or does it not have this feature?

https://github.com/johnsoncodehk/volar/tree/master/extensions/vscode-vue-language-features

I followed the steps prompted in the document and found that the type overload of Vue needs to be deleted. The slot parameter type prompt of volar can take effect

aziztitu:
@johnsoncodehk I was able to follow your instructions and get the type checking for Props and ScopedSlots working, but for the Emit, it seems like there might be a type mismatch.

image

In the Vue types package, it looks like $emit expects a Function type. So when using an object with a bunch of functions as it's type, it complains due to the conflict.

image


Should we override a different variable other than $emit instead?

johnsoncodehk:
@aziztitu Vue 2 do not support typed emit, if you do want to, you can try omit `$emit` from `Vue` in interface.

aziztitu:
@johnsoncodehk Gotcha. I was able to at least get the Event suggestions working by hacking it like this:

image
image


This stops any errors from being thrown, but the only issue is there's no actual type checking happening. It seems to think that all events now take 2 args :D 
image


But like you said, it probably is just due to the fact that it's Vue 2. I can just remove the args from the Emit type, and live with it for now :)

johnsoncodehk:
@aziztitu please track #1855

PR #817, opened 2021-12-25
chore: add coc-volar to monorepo

johnsoncodehk:
Since coc-volar is relying on `@volar/server` and `@volar/shared`, I hope to reduce the maintenance cost of coc-volar when updating `@volar/server` and `@volar/shared`.

I am exploring the maintenance method of adding coc-volar to monorepo by git submodule. When I change the code of `@volar/server` and `@volar/shared`, if it affects coc-volar, I can get TS error in volar project, and then Submit a PR or commit to coc-volar in time.

If we use this method, you can choose to update tsconfig to adapt to monorepo (https://github.com/johnsoncodehk/coc-volar/commit/60f4cc2c9909d36a25f30b996460889e6c7fc124), or keep the original tsconfig and let volar monorepo adapt it.

@yaegassy Do you expect this method of cooperation?

### DIFF
--- a/.gitmodules
+++ b/.gitmodules
@@ -1,0 +1,3 @@
+[submodule "packages/coc-volar"]
+	path = packages/coc-volar
+	url = https://github.com/yaegassy/coc-volar.git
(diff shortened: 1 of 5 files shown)


yaegassy:
I'm sorry for the late reply. Thanks for caring about `coc-volar`. 🙇

It's not a particular burden at the moment, so it's fine. :)

PR #822, opened 2021-12-27
feat: support extracting used components from @Options/@Component 

SniperJoe:
Issue: https://github.com/johnsoncodehk/volar/issues/753
This issue is already closed, but the current solution forces users to specify components twice, in the decorator and as class property. Also users will have to rewrite all of their components and add new property in order to make volar work properly. This code will allow users to leave their code as-is and still get proper components IntelliSense. Autoimport works too. 

Also I had to remove ```NonNullable<typeof __VLS_component_ts.components>``` as it's incompatible with the new code and no longer actually needed. More than that, it breaks global components if ```static components``` property is not specified.


johnsoncodehk:
Thanks for try to improve! But I thought about it for a long time, I think this feature is violates volar design principles. All types of behavior should can predicted based on TS. When use decorators, it shouldn't affect component type.

PR #839, opened 2022-01-04
fix: support string type  in v-for

Teemwu:


johnsoncodehk:
Hi @Teemwu, could you give me an example of what is the problem to solve?

Teemwu:
Hi @johnsoncodehk,  when I use a string in v-for, volar will warn, but it can be rendered by vue.
```vue
<template>
  <div v-for="item in message" :key="item">{{ item }}</div>
</template>

<script setup lang="ts">
import { ref } from 'vue'

const message = ref('Foo')
</script>
```
image

johnsoncodehk:
LGTM, thanks!

PR #851, opened 2022-01-09
fix: resolve dependency issues

merceyz:
**What's the problem this PR addresses?**

- `@volar/shared` depends on `vscode-html-languageservice` but doesn't declare it as a dependency.
- The patch that `vue-tsc` applies to `tsc` doesn't pre-resolve the path to the proxy which makes it rely on hoisting to place `typescript` and `vue-tsc` next to each other which isn't guaranteed to happen.

**How did you fix it?**

- Move `vscode-html-languageservice` from `devDependencies` to `dependencies`.
- Inject the absolute path to the proxy instead of an identifier.

### DIFF
--- a/packages/shared/package.json
+++ b/packages/shared/package.json
@@ -14,12 +14,12 @@
 	},
 	"dependencies": {
 		"upath": "^2.0.1",
+		"vscode-html-languageservice": "^4.2.1",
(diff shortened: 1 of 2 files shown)


johnsoncodehk:
LGTM, thanks!

PR #866, opened 2022-01-17
fix(schemas): missing "vue-jsconfig.schema.json"

yaegassy:
It seems that `vue-jsconfig.schema.json` does not exist in the schemas directory. When editing `jsconfig.json`, the json language server will output an error.

fix-volar-vue-jsconfig-schemas

Changed `jsconfig.json` related "jsonValidation" to `vue-tsconfig.schema.json`.

### DIFF
--- a/extensions/vscode-vue-language-features/package.json
+++ b/extensions/vscode-vue-language-features/package.json
@@ -116,7 +116,7 @@
 			},
 			{
 				"fileMatch": "jsconfig.json",
-				"url": "./schemas/vue-jsconfig.schema.json"
+				"url": "./schemas/vue-tsconfig.schema.json"
 			},
 			{
 				"fileMatch": "jsconfig.*.json",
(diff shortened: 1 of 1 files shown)


yaegassy:
It seems that the error is also output when editing `tsconfig.json`.

It seems that `tsconfig.schema.json` does not exist in the schema directory.

fix-volar-vue-tsjsconfig-schemas

I have never developed and released a VSCode extension, so I can't pinpoint the cause, but maybe `.vscodeignore` is affected. <https://github.com/johnsoncodehk/volar/blob/master/extensions/vscode-vue-language-features/.vscodeignore>

johnsoncodehk:
LGTM, thanks!

angelhdzmultimedia:
Same here. I solved this temporarily by duplicating `jsconfig.schema.json` and renaming it to `tsconfig.schema.json`.

The fix is on the way! 🎉👀

Ragura:
It appears it's still showing the error after the fix, because of these lines (starts at line 57) still referencing the version of the schema file without the vue- prefix which doesn't exist on disk. Deleting these lines in favour of the next in line (with vue- prefix) solves the issue.

```json
{
  "fileMatch": "tsconfig.json",
  "url": "./schemas/tsconfig.schema.json"
},
```

PR #870, opened 2022-01-17
Add emacs volar integration by jadestrong

sethidden:
See https://github.com/jadestrong/lsp-volar

It also has multi server support as can be seen in https://github.com/jadestrong/lsp-volar/blob/main/lsp-volar.el (CTRL+F volar-api, volar-doc, volar-html)


johnsoncodehk:
Thanks for the update!

Issue #781, opened 2021-12-13
Can we have basic snippets along with volar extension? (Just like vetur)


anburocky3:
Vetur has some basic snippets built in like `template`, `script` & `style`. Can we have those with both `vue` & `vue-ts` snippets?

sethidden:
Volar doesn't have this. You're supposed to use VS Code's snippets instead.
The maintainer's stance on this is here https://github.com/johnsoncodehk/volar/issues/183
See https://github.com/johnsoncodehk/volar/blob/master/extensions/vscode-vue-language-features/README.md?plain=1#L126

anburocky3:
So this means, volar doesn't have default vue snippets and we have to create it on our own? Am i right?

sethidden:
Yes — you need to use https://code.visualstudio.com/docs/editor/userdefinedsnippets

anburocky3:
Ok thanks, I'll make my own. Does the @johnsoncodehk accept PR related to basic vue snippets with this extension? If so, I can contribute to that.

johnsoncodehk:
You can submit pr and I will test it to see if it applies.

Issue #183, opened 2021-05-18
[FR] Customizable Scaffold Snippets using `.vue` files


wenfangdu:
I haven't found Vetur's [Customizable Scaffold Snippets](https://vuejs.github.io/vetur/guide/snippet.html#customizable-scaffold-snippets) equivalent in Volar, I think its main advantage is:

> You can use everything that's allowed in [VS Code Snippet Syntax](https://code.visualstudio.com/docs/editor/userdefinedsnippets). The good thing is, you write them in `.vue` files instead of `.json` files, and you don't need to escape special characters.

Could you please consider adding it?

johnsoncodehk:
Thanks for your suggestion, but I think plugin specific things is not good for development ecology, I think it is simpler and more appropriate to build a .vue -> .json converter.

wenfangdu:
@johnsoncodehk I've gone ahead and created a [snippet generator extension](https://github.com/wenfangdu/vscode-snippet-generator) for this. It supports code to JSON conversion.

johnsoncodehk:
Great work! Added to readme.

I hope there is an online converter for two-way conversion, which can help edit and preview existing VSCode snippets.

PR #884, opened 2022-01-21
docs: suggest snippets from the VSCode Marketplace

jgaehring:
Hi, I'd like to propose a tiny addition to the docs if you're accepting.

I think this would be helpful to folks migrating over to Volar from Vetur who, like myself, don't really care about customizing their snippets, but just want some quick, ready-made snippets they can install with a click. Sarah Drasner's snippets are relatively popular, with Vue 3 support, and she's a member of Vue's core team so I suspect they will be well maintained. I read through #183 and #781 and appreciate the rationale not to include built-in snippets, but this seems like an easy alternative.

Happy to amend or close if you're not accepting PR's of this sort. Thanks!


johnsoncodehk:
LGTM, thanks for the update!

anburocky3:
@jgaehring  Looks fine, but the templates are generated in this order.

```html
<template>

<script>

<style>
```

And according to vue3, we should be sticking with the following order.

```html
<script>

<template>

<style>
```

PR #890, opened 2022-01-23
chore: add "-" to the html completion trigger characters

yaegassy:
In "coc-volar", starting from `v0.30.2` of `@volar/server`, typing `-` in `<template>` causes completion to be canceled...

- **REF**: <https://github.com/yaegassy/coc-volar/issues/110>

I considered making adjustments on the `coc-volar` side, but that would have led to other problems, so I made adjustments on the language server side.

If adding `-` causes problems, don't worry about it, just close this PR. 🙇

### DIFF
--- a/packages/vscode-vue-languageservice/src/services/completion.ts
+++ b/packages/vscode-vue-languageservice/src/services/completion.ts
@@ -21,7 +21,7 @@ export function getTriggerCharacters(tsVersion: string) {
 	return {
 		typescript: ts2.getTriggerCharacters(tsVersion),
 		jsdoc: ['*'],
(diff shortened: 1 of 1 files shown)


johnsoncodehk:
Hi @yaegassy, I do not mind to add this. Just to clarify, how do coc html plugin resolve this problem?

yaegassy:
Thanks for the quick reply. I was surprised at how fast you replied!

I don't think `coc-html` does anything different, but the completion of attributes containing `-` is not cancelled.

johnsoncodehk:
I would like to investigate why coc-html is working fine at this point. 🤔

yaegassy:
Thanks for commenting on the issue on the `coc-volar` side.

- <https://github.com/yaegassy/coc-volar/issues/110#issuecomment-1019448547>
- <https://github.com/yaegassy/coc-volar/issues/110#issuecomment-1019453356>

This PR is closed because it might be caused by a problem on the `coc.nvim` side.

yaegassy:
A fix on the `coc.nvim` side solved this problem. 🙇

PR #955, opened 2022-02-11
fix: globalProperties not ShallowUnwrapRef on the template

### LINKED ISSUE #954: Incorrect globalProperties types on the template
[GlobalProperties aka ComponentCustomProperties ](https://vuejs.org/guide/typescript/options-api.html#augmenting-global-properties) are not property shallowUnref on the template type.

```tsx
import { Ref } from "vue";

declare module "vue" {
  interface ComponentCustomProperties {
    $myVar: Ref<number>;
  }
}

// Comp.vue
<template>
  {{ $myVar /* type: Ref<number>; Prints: number*/}}
</template>
```

Found on https://github.com/vuejs/core/pull/5389

pikax:
fix #954

This will shallow unwrap the `__VLS_ctx` type

### DIFF
--- a/packages/vscode-vue-languageservice/src/utils/localTypes.ts
+++ b/packages/vscode-vue-languageservice/src/utils/localTypes.ts
@@ -94,7 +94,7 @@ export type ExtractComponentProps<T> =
 	: T extends FunctionalComponent<infer P2> ? P2
 	: T
 
-export type ExtractRawComponents<T> = { [K in keyof T]: ExtractRawComponent<T[K]> };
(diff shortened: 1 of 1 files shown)


johnsoncodehk:
Hi @pikax, I think you should add `ShallowUnwrapRef` to https://github.com/vuejs/core/blob/cea82cffa3963d6d7f82f490ba0dba50a1d202db/packages/runtime-core/src/componentPublicInstance.ts#L206 instead of here. Otherwise it only works for volar but not works for webstorm.

PR #1008, opened 2022-03-04
Document recursive components

zigomir:
As encouraged in https://github.com/johnsoncodehk/volar/issues/644#issuecomment-1059143082

Sorry about larger diff, I removed all `>`s from Vetur note. Let me know if you prefer the quote to stay as is.


johnsoncodehk:
LGTM, thanks!

PR #1046, opened 2022-03-15
fix: use `console.warn` instead of `console.log`

### LINKED ISSUE #1043: Bug: vim-lsp fails to parse headers sent from volar when using @volar/vue-language-server with `--stdio` flag
For example, when the user don't set `volar.config.js`, volar v0.33.0 or later calls `console.log` for logging. But, when using volar with `--stdio`, texts output from `console.log` also included to messages sent to vim-lsp. Then, it breaks parsing function in vim-lsp and I cannnot use volar... After I replaced `console.log` to `console.warn`, which means the message goes to stderr, both volar and vim-lsp look work fine.
https://github.com/johnsoncodehk/volar/blob/3852ee85c8d7a8e20e0bca084d2b73cf11491f93/packages/vue-language-service/src/languageService.ts#L519

Also, following code could cause same problem.
https://github.com/johnsoncodehk/volar/blob/271cdc7c3ad9f93725b422ce7e4086775b09929b/packages/shared/src/ts.ts#L142

tsukkee:
fix #1043 

This PR replaces `console.log` to `console.warn` where their output affects to LSP clients.

### DIFF
--- a/packages/shared/src/ts.ts
+++ b/packages/shared/src/ts.ts
@@ -139,7 +139,7 @@ function resolveVueCompilerOptions(rawOptions: {
 		try {
 			result.experimentalTemplateCompilerOptions = require(templateOptionsPath).default;
 		} catch (error) {
-			console.log('Failed to require "experimentalTemplateCompilerOptionsRequirePath":', templateOptionsPath);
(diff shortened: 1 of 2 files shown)


johnsoncodehk:
Thank you!

Issue #992, opened 2022-02-28
[Bug]: False positive type check on force binding attribute with `vue-tsc`


ahnpnl:
Hi,

I ran into some false positive type check errors with my Vite + Vue tsc when using force binding attribute template syntax
<img width="561" alt="image" src="https://user-images.githubusercontent.com/27772165/156058301-656d613c-924b-4561-a4ad-1a64343bcbd0.png">

Here is the repo to reproduce the issue https://github.com/ahnpnl/vite-vue-ts

### Step to reproduce

- Check out the repo
- Install deps with `yarn`
- Run `yarn production` or `yarn ts-check`
- Observe the error like the screenshot.

johnsoncodehk:
Hi @ahnpnl, I think template AST is broken by this syntax, is it valid in vue? Could you share document link if it have?

ahnpnl:
Hi, I use https://vuejs.org/api/built-in-directives.html#v-bind 

I also tried the syntax like the docs related to attribute and prop force binding `.`/`^` I also ran into similar issue.

PR #1047, opened 2022-03-15
fix(code-gen): support attribute binding syntax `:<name>.attr`

ahnpnl:
### Summary

Added missing support for attribute binding syntax `:<name>.attr`.  Reference https://vuejs.org/api/built-in-directives.html#v-bind

For example
```
<div :someProperty.attr="someString"></div>
```

related to #992 

The other way of binding attribute
```
<div ^someProperty="someString"></div>
```
I don't know how to fix yet. When debugging, `prop.type` for this syntax is `CompilerDOM.NodeTypes.ATTRIBUTE` while the above syntax is `CompilerDOM.NodeTypes.DIRECTIVE`. Not sure if that is Vue SFC compiler issue or not. 

IMO, since both syntaxes are equivalent, their `prop.type` should be the same. Probably you have more insight with Vue team about that.

### Test plan

- Pick up any projects using `vue-tsc`
- Create a SFC file with attribute binding syntax `:<name>.attr` in HTML part
- Run `vue-tsc` against that SFC file and verify if no error occurs.

### DIFF
--- a/packages/vue-code-gen/src/generators/template.ts
+++ b/packages/vue-code-gen/src/generators/template.ts
@@ -393,8 +393,8 @@ export function generate(
 						? prop.arg.content
 						: prop.arg.loc.source
 
-					if (prop.modifiers.some(m => m === 'prop')) {
-						propName = propName.substr(1);
+					if (prop.modifiers.some(m => m === 'prop' || m === 'attr')) {
+						propName = propName.substring(1);
 					}
 
 					if (prop.name === 'bind' || prop.name === 'model') {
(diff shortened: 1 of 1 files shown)


ahnpnl:
cc @johnsoncodehk

johnsoncodehk:
I will test it later, thank for the fix!

johnsoncodehk:
LGTM! but I forgot why not just use `prop.arg.loc.source` instead of `prop.arg.content.substring(1)`, will add some tests for record in future.

> I don't know how to fix yet. When debugging, `prop.type` for this syntax is `CompilerDOM.NodeTypes.ATTRIBUTE` while the above syntax is `CompilerDOM.NodeTypes.DIRECTIVE`. Not sure if that is Vue SFC compiler issue or not.

I think this is SFC compiler issue, because `^someProperty` should `:someProperty.attr` be same in AST in intuition.

PR #1136, opened 2022-04-02
Include backtick to surroundingPairs

gcoda:


johnsoncodehk:
Thanks for the PR! But we following this config with https://github.com/microsoft/vscode/blob/24e73f3551a59166d34d0deecbb623c24af67645/extensions/html/language-configuration.json#L19-L26, could you explain what problem will be resolve with adding "`"?

gcoda:
`.vue` files have `<script>` section and bits of javascript in directives and `:attributes=""`. 
Backticks\Template literals are working just fine inside vue templates, it is kinda surprising, to select some text, hit ` and watch it disappear. 

try rewriting

```vue
<th class="table-header">
```

to

```vue
<th :class="`table-header-colspan-${computedColspan}`"
```

johnsoncodehk:
Add ```{ "open": "`", "close": "`" }``` to `vue-language-configuration.json` cannot resolve this problem to me.

If you typing ``` ` ``` at `<th :class=" | ">` it will be working, because the embedded language is `typescript` so it's already using `javascript-language-configuration.json`.

螢幕截圖 2022-04-07 13 41 03

But with `<th :class="|">` it's not working anyway, it may a intentional vscode behavior.

johnsoncodehk:
I think I found the reason, it may need to fix in https://github.com/johnsoncodehk/volar/blob/master/extensions/vscode-vue-language-features/syntaxes/vue.tmLanguage.json, nice to have a issue for tracking.

PR #1528, opened 2022-06-30
Push binding if importClause if default export

IIFelix:
Fixes https://github.com/johnsoncodehk/volar/issues/1524

If importClause is default export binding won't be added. 
@johnsoncodehk pls review because i dont know if this break anything. Tests are running fine.


johnsoncodehk:
This problem is tricky 🤦, maybe have to big change the implement to resolve. (btw admire you found the code, I know how that hard to read the code. :))

IIFelix:
Thanks 😄

Issue #1332, opened 2022-05-23
[Feature Request] Support using nightly TypeScript


tjx666:
As we know, there is an extension [JavaScript and TypeScript Nightly](https://github.com/microsoft/vscode-typescript-next) which enables the nightly build of TypeScript (typescript@next) as VS Code's built-in TypeScript version used to power JavaScript and TypeScript IntelliSense.

![image](https://user-images.githubusercontent.com/41773861/169733077-a520b0d3-990e-4694-beee-dbdc341c8551.png)

But now volar only support toggle typescript version between builtin and workspace version:

![image](https://user-images.githubusercontent.com/41773861/169733033-579ad042-0951-4442-ae4e-4274bc57ea10.png)

It would be helpful to support using the nightly typescript version. For example, I can using the latest `got to source definition` function comes with the latest typescript.

![image](https://user-images.githubusercontent.com/41773861/169733265-b3b7bbff-5f6e-4ab0-b58d-2bcfecd3e838.png)

tjx666:
Seems not work in latest version. @johnsoncodehk

johnsoncodehk:
@tjx666 What behavior are you expecting and it doesn't work?

tjx666:
For TS project, I can easily toggle to ts4.8 with extension: [JavaScript and TypeScript Nightly](https://github.com/microsoft/vscode-typescript-next).


https://user-images.githubusercontent.com/41773861/179938524-9a10443c-5fcc-4f23-b435-72fe6a1d48c1.mov

But volar can only select 4.7:

![image](https://user-images.githubusercontent.com/41773861/179938978-60709f8b-f34c-470d-ba1d-89dcf2636950.png)

tjx666:
@johnsoncodehk So, how can I easily toggle ts version to 4.8. Seems I have to install 4.8 as dependency now.

yaegassy:
@tjx666 This "JavaScript and TypeScript Nightly" extention goes into the extension folder installed by the user.

Try setting `typescript.tsdk` in `.vscode/settings.json`. This is an example of my environment.

```jsonc
{
  "typescript.tsdk": "/Users/yaegassy/.vscode/extensions/ms-vscode.vscode-typescript-next-4.8.20220719/node_modules/typescript/lib",
}
```

You will probably see more items available for selection in "Use Workspace Version". Perhaps, but it is also 
possible to use typescript/lib installed at an arbitrary location.

[...]

PR #1629, opened 2022-07-24
feat: support for "JavaScript and TypeScript Nightly"

tjx666:
#1332


extensions/vscode-vue-language-features/src/features/tsVersion.ts
```diff
@@ -158,6 +149,17 @@ function getWorkspaceTsPaths(useDefault = false) {
 }
 
 function getVscodeTsPaths() {
+	const nightly = vscode.extensions.getExtension('ms-vscode.vscode-typescript-next');
```
tjx666, line 152: It seems that user can only select nightly version when user install `ms-vscode.vscode-typescript-next` extension. They can't select the builtin typescript version.

johnsoncodehk: It's for consistent with VSCode behavior, because seems VSCode cannot select it too.

PR #1639, opened 2022-07-26
feat(vue-component-meta): extract the "exposed" metadata

elevatebart:
When a component uses `expose` or `defineExpose()` on its methods or data, it needs to be documented.

This PR adds documentation for the method.


johnsoncodehk:
LGTM, thanks!

Issue #533, opened 2021-09-25
Support other preprocessors


phil294:
It would be great if we also had support for less-popular preprocessors like e.g. SLM (HTML) or Elm, CoffeScript (JS).

Surely it is a lot of work natively integrating and supporting these, so I suggest a LSP interface for custom preprocessors. Say, a configuration object
```json
"preprocessors": {
  "coffee": {
    "lspPath": "/some/path"
  }
}
```
And now, whenever Volar encounters a unsupported lang type as `<something lang="coffee">`, it simply functions as a proxy between client and a spun-up third party lsp instance at `/some/path`.

Of course, this falls short for special Vue syntax, like autocompletion for `v-if` inside a custom html preprocessor. But it's better than having no intellisense at all. In the case of Vue 3 composition api, things look bright anyway in comparison to Vue 2, where we needed [magic under the hood](https://github.com/vuejs/vetur/blob/master/server/src/services/typescriptService/bridge.ts) to get TS integration.

Vetur's approach to preprocessors [is to ignore them](https://github.com/vuejs/vetur/issues/1184) and instead wait for the specific language to consider Vue syntax individually, like in the case of [ESLint](https://github.com/vuejs/eslint-plugin-vue). Volar seems to aim at integrating preprocessors into the codebase, if I understand the intent correctly. I think both approaches (if taken to the extreme) are wrong. Programming languages should arguably have higher precedence than frameworks, but language-to-framework is a a many-to-many relationship after all, [just like tooling-to-ide is](https://code.visualstudio.com/assets/api/language-extensions/language-server-extension-guide/lsp-languages-editors.png), so let's try and reuse existing servers [whereever possible](https://microsoft.github.io/language-server-protocol/implementors/servers/).
  
Congrats on becoming the recommended extension for Vue 3!

johnsoncodehk:
Hi @phil294, this is the purpose of [Plugin API](https://github.com/johnsoncodehk/volar/issues/185).

For Elm / CoffeScript, it needs to be completely independent of TS processing, there is not much to talk about.

For template preprocessors, we need to implement a mapping protocol with template AST to support type flow analysis with v-if / v-for, and language support in interpolations.

Currently Volar does this internally for pug, but I planed migrate to plugin API in future.

phil294:
ok I understand, thanks. Looking forward to this!

PR #1687, opened 2022-08-10
feat: Support for Plugin API

### LINKED ISSUE #185: Support for Plugin API
See https://github.com/johnsoncodehk/volar/issues/127#issuecomment-815337186

Tasks:
- [ ] Pluginization embedded file parsing

johnsoncodehk:
Closes #185

Please note that the API interface is not stable, it may change before v1.0.

## Usage

- `tsconfig.json`

```jsoncc
{
  "vueCompilerOptions": {
    "plugins": ["./your-plugin"]
  }
}
```

Example: https://github.com/johnsoncodehk/volar-starter/blob/master/vue-i18n.js

### DIFF
--- a/packages/alpine-language-core/src/index.ts
+++ b/packages/alpine-language-core/src/index.ts
@@ -1,5 +1,5 @@
 import * as vue from '@volar/vue-language-core';
-import useHtmlFilePlugin from './plugins/file-html';
+import * as useHtmlFilePlugin from './plugins/file-html';
 
 export type LanguageServiceHost = vue.LanguageServiceHost;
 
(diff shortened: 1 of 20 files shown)


phil294:
Hi, thanks for your work on this! This should make adding e.g. a custom HTML parser a breeze. However, am I understanding the plugin syntax correctly, in that it is *not yet possible* to integrate an external LSP for scripts (#533)? So that go-tos, autocomplete etc. all can work with JS preprocessor languages like CoffeeScript or Elm.

johnsoncodehk:
@phil294 You can convert CoffeeScript code to TypeScript code with mapping in `resolveEmbeddedFile` API, it can provide the same level language capabilities as TypeScript script blocks in theory. But currently implement cannot reuse https://github.com/johnsoncodehk/volar/blob/master/packages/vue-language-core/src/plugins/vue-tsx.ts code-gen logic with JS preprocessor, so yes this way not without significant effort.

You can open a feature request (and best to attach sample repro, so I can see what I actually need to solve), I need to find a suitable implementation.

PR #1707, opened 2022-08-11
Extract component feature

IIFelix:
Add new command for extracting selected text into new component.
Also add new docs entry.


johnsoncodehk:
Thanks! I also have to investigate how to implement auto import component after extracted.

johnsoncodehk:
After a long consideration, I decided not to add this feature for the following reasons. 🙏

- Extract component feature is not suitable for everyone. Some people may expect to only extract the template without making any script changes; some may expect to refactor the used properties into props at the same time.
- For implement props extraction, language server needs to add a new request API, and I want to avoid language server scope bigger.

PR #1851, opened 2022-09-09
fix: no binding, no slot

elevatebart:
If a slot is not bound with any value, `vue-component-meta` does not detect it. 

I believe and hope this is an easy fix.

I took the liberty of writing a test.


johnsoncodehk:
Hi @elevatebart, seems test don't fail, could you update it?

johnsoncodehk:
Just merge it because this is not a breaking change.

PR #1964, opened 2022-10-09
feat: Support RFC 436 (Experimental)

johnsoncodehk:
RFC: https://github.com/vuejs/rfcs/discussions/436

## Usage

```jsonc
{
  "vueCompilerOptions": {
    "jsxTemplates": true,
    "experimentalRfc436": true
  }
}
```

Please note that `extends any` is required, otherwise JSX syntax will be broken.

```html
<script setup lang="ts" generic="T extends any">
defineProps<{ msg: T }>()
</script>
```

### DIFF
--- a/vue-language-tools/vue-language-core/schemas/vue-tsconfig.schema.json
+++ b/vue-language-tools/vue-language-core/schemas/vue-tsconfig.schema.json
@@ -90,6 +90,11 @@
 						"never"
 					],
 					"markdownDescription": "https://github.com/johnsoncodehk/volar/issues/1038, https://github.com/johnsoncodehk/volar/issues/1121"
+				},
+				"experimentalRfc436": {
+					"type": "boolean",
+					"default": false,
+					"markdownDescription": "https://github.com/vuejs/rfcs/discussions/436"
 				}
 			}
 		}
--- a/vue-language-tools/vue-language-core/src/generators/script.ts
+++ b/vue-language-tools/vue-language-core/src/generators/script.ts
@@ -245,7 +245,11 @@ export function generate(
 					0,
 					{ diagnostic: true },
 				]);
-				codeGen.push('export default await (async () => {\n');
+				codeGen.push('export default await (async ');
+				if (vueCompilerOptions.experimentalRfc436 && sfc.scriptSetup.generic) {
+					codeGen.push(`<${sfc.scriptSetup.generic}>`);
+				}
+				codeGen.push('() => {\n');
 			}
 
 			codeGen.push('const __VLS_setup = async () => {\n');
@@ -377,12 +381,25 @@ export function generate(
 
 			writeTemplate();
 
+			codeGen.push(`return {} as typeof __VLS_Component`);
 			if (htmlGen?.slotsNum) {
-				codeGen.push(`return {} as typeof __VLS_Component & (new () => { ${getSlotsPropertyName(vueVersion)}: ReturnType<typeof __VLS_template> });\n`);
+				codeGen.push(` & (new () => { ${getSlotsPropertyName(vueVersion)}: ReturnType<typeof __VLS_template> })`);
 			}
-			else {
-				codeGen.push(`return {} as typeof __VLS_Component;\n`);
+			if (vueCompilerOptions.experimentalRfc436 && sfc.scriptSetup.generic) {
+				codeGen.push(` & (new <${sfc.scriptSetup.generic}>() => {\n`);
+				if (scriptSetupRanges.propsTypeArg) {
+					codeGen.push(`$props: `);
+					addVirtualCode('scriptSetup', scriptSetupRanges.propsTypeArg.start, scriptSetupRanges.propsTypeArg.end);
+					codeGen.push(`,\n`);
+				}
+				if (scriptSetupRanges.emitsTypeArg) {
+					codeGen.push(`$emit: `);
+					addVirtualCode('scriptSetup', scriptSetupRanges.emitsTypeArg.start, scriptSetupRanges.emitsTypeArg.end);
+					codeGen.push(`,\n`);
+				}
+				codeGen.push(`})`);
 			}
+			codeGen.push(`;\n`);
 
 			codeGen.push(`};\n`);
 			codeGen.push(`return await __VLS_setup();\n`);
--- a/vue-language-tools/vue-language-core/src/sourceFile.ts
+++ b/vue-language-tools/vue-language-core/src/sourceFile.ts
@@ -440,6 +440,7 @@ export class VueSourceFile implements SourceFile {
 				endTagStart: block.loc.end.offset,
 				content: block.content,
 				lang: block.lang ?? 'js',
+				generic: typeof block.attrs.generic === 'string' ? block.attrs.generic : undefined,
 			} : null;
 
 			if (self.sfc.scriptSetup && newData) {
--- a/vue-language-tools/vue-language-core/src/types.ts
+++ b/vue-language-tools/vue-language-core/src/types.ts
@@ -30,6 +30,7 @@ export interface ResolvedVueCompilerOptions {
 	experimentalTemplateCompilerOptions: any;
 	experimentalTemplateCompilerOptionsRequirePath: string | undefined;
 	experimentalResolveStyleCssClasses: 'scoped' | 'always' | 'never';
+	experimentalRfc436: boolean;
 }
 
 export type VueLanguagePlugin = (ctx: {
@@ -66,7 +67,10 @@ export interface Sfc {
 	script: (SfcBlock & {
 		src: string | undefined;
 	}) | null;
-	scriptSetup: SfcBlock | null;
+	scriptSetup: SfcBlock & {
+		// https://github.com/vuejs/rfcs/discussions/436
+		generic: string | undefined;
+	} | null;
 	styles: (SfcBlock & {
 		module: string | undefined;
 		scoped: boolean;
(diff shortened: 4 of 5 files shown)


edimitchel:
Answering https://github.com/johnsoncodehk/volar/pull/1964

edimitchel:
Ping https://github.com/vuejs/rfcs/discussions/436

StepanMynarik:
Full of excitement I tried it right away in my multiselect component.
When passing a template into the multiselect item slot, I still see the item as the base 'Item' class.

In my multiselect component I did this: `<script setup lang="ts" generic="T extends Item">`
Also: `const {
  items,
  selectedItem,
} = defineProps<{
  items: T[];
  selectedItem: T[];
}>();`

When using it, I made sure to pass nested 'ExtendedItem' class instances to both params.
In the slot however I still only see it as the base 'Item' class.

Are there any extra steps required?

NOTE: I made sure to first update Volar extension and vue-tsc to 1.0.3

johnsoncodehk:
@StepanMynarik I think the current design is not enough to support generic typed slots, we may need a new `defineSlots` api like this.

```ts
const { items, selectedItem, } = defineProps<{ items: T[]; selectedItem: T[]; }>();
defineSlots<{ default(args: { item: T }): VNode[] }>();
```

It's better to present your use case in an RFC so peoples can discuss with it.

StepanMynarik:
@johnsoncodehk Oh, I must have got it wrong. What is the current intended use case then?

I will make sure to post my use case to the RFC, good point.

johnsoncodehk:
@StepanMynarik I'm checking if this might working without add new API, will let you know later.

johnsoncodehk:
@StepanMynarik please track #1987

johnsoncodehk:
@colinj you should move interface to `<script>` block.

```html
<script lang="ts">
interface Props<T> {
  msg: T;
}
</script>

<script setup lang="ts" generic="T extends string | number">
const props = defineProps<Props<T>>(); 
</script>
```

You can use `Show Virtual Files` command to inspect codegen result.

johnsoncodehk:
This is `withDefaults` problem, you can use this workaround.

```ts
<script lang="ts">
interface Props<T> {
  msg?: T;
}
</script>

<script setup lang="ts" generic="T extends string | number">
const props = withDefaults(defineProps<Props<T>>() as Props<string | number>, {
  msg: "Fred", // TS Error
  // Type 'string' is not assignable to type 'InferDefault<Readonly<Props<T>>, NotUndefined<T>>'.ts(2322)
});
</script>
```

colinj:
Yes, I started to look inside `withDefaults` to see how it works but I think this workaround is just what is needed to be done. 
Thanks again @johnsoncodehk!

colinj:
I am also getting eslint (no-undef) errors for `T` (`'T' is not defined`) so I have added the following entry to my `.eslintrc.cjs` file to ignore `T`.

```js
module.exports = {
  ...
  rules: { ... },
  globals: {
    T: "readonly",  // naming convention used to define the type on generic Vue components
  },
};
```

Obviously, this means I would need to use `T` to avoid linting errors when defining the generic type for these components.

grindpride:
Enabling `vueCompilerOptions` for this feature causes template refs type errors

```
error TS2344: Type '() => Omit<JSX.Element, "props" | "children"> & Omit<{ $: ComponentInternalInstance; $data: {};
 $props: Partial<{}> & Omit<Readonly<ExtractPropTypes<...>> & { ...; } & VNodeProps & AllowedComponentProps & ComponentCustomProps, never>; ... 10 more ...; $watch(source: string | Function, cb: Function, options?: WatchOp...' does not satisfy the constraint 'abstract new (...args: any) => any'.
  Type '() => Omit<Element, "props" | "children"> & Omit<{ $: ComponentInternalInstance; $data: {}; $props: Partial<{}> & Omit<Readonly<ExtractPropTypes<...>> & { ...; } &
 VNodeProps & AllowedComponentProps & ComponentCustomProps, never>; ... 10 more ...; $watch(source: string | Function, cb: Function, options?: WatchOption...' provides no match for the signature 'new (...args: any): any'.

const dateNavigationRef = ref() as Ref<InstanceType<typeof DateNavigation>>

```

johnsoncodehk:
@grindpride See https://github.com/johnsoncodehk/volar/pull/1987#issuecomment-1275359287

grindpride:
@johnsoncodehk Can you also see this issue https://github.com/johnsoncodehk/volar/issues/1405#issuecomment-1279954491 ?

geoffgscott:
Type checking works amazing and passes but I do get this message in VSCode anytime a generic is used inside defineProps:
```
The inferred type of 'default' references a type with a cyclic structure which cannot be trivially serialized. A type annotation is necessary.
```

```vue
<script lang="ts" setup generic="TC extends readonly string[] | string[]">
defineProps<{
    options: TC;
}>()
</script>
```

ThaDaVos:
I noticed this working amazingly during dev builds - but prod builds are broken as it cannot seem to be able to resolve the file (te name seems to include the generic part) - but that may also be caused because I am using `Laravel Vite`

johanninos:
> @grindpride See [#1987 (comment)](https://github.com/johnsoncodehk/volar/pull/1987#issuecomment-1275359287)

InstanceType<typeof BButton<T>> can pass generic.
but ReturnType can't untill typescript 4.7.0

but vue has bug with this ts new feature

https://github.com/vuejs/core/issues/7161

StepanMynarik:
> Type checking works amazing and passes but I do get this message in VSCode anytime a generic is used inside defineProps:
> 
> ```
> The inferred type of 'default' references a type with a cyclic structure which cannot be trivially serialized. A type annotation is necessary.
> ```
> 
> ```
> <script lang="ts" setup generic="TC extends readonly string[] | string[]">
> defineProps<{
>     options: TC;
> }>()
> </script>
> ```

Same here with ```generic="T extends Item"```.
But only when I use T directly in params, as a single item.
When I declare ```T[]``` prop, then no error :D

dajpes:
For those using vue 3, you need to paste the following in tsconfig.app.json:

```
"vueCompilerOptions": {
    "jsxTemplates": true,
    "experimentalRfc436": true
  }
```

damafeez:
> Type checking works amazing and passes but I do get this message in VSCode anytime a generic is used inside defineProps:
> 
> ```
> The inferred type of 'default' references a type with a cyclic structure which cannot be trivially serialized. A type annotation is necessary.
> ```
> 
> ```
> <script lang="ts" setup generic="TC extends readonly string[] | string[]">
> defineProps<{
>     options: TC;
> }>()
> </script>
> ```

I get the same error as well. Does anyone know how to resolve it?

damafeez:
> Type checking works amazing and passes but I do get this message in VSCode anytime a generic is used inside defineProps:
> 
> ```
> The inferred type of 'default' references a type with a cyclic structure which cannot be trivially serialized. A type annotation is necessary.
> ```
> 
> ```
> <script lang="ts" setup generic="TC extends readonly string[] | string[]">
> defineProps<{
>     options: TC;
> }>()
> </script>
> ```

I get the same error, any knows how to get rid of it?

rodrigocfd:
Is it possible to use this with Vite?

jmoore914:
> Is it possible to use this with Vite?

This is working for me with Vite.

I am running into the cyclic type error as others have mentioned though.

dakt:
This is freaking amazing. The one thing that kept React better suited for large project was this. Please, oh please make this official fast, please...I typed our ui lib with this and it boosted DX 10 fold

Paul7Peterson:
A nice example if you want to type the `modelValue` of a select component based on the given options:

```vue
<script setup lang="ts" generic="VueGen extends AppSelectOption<string | number>">
  export interface AppSelectOption<T extends string | number> {
    value: T,
    text: string;
    disabled?: boolean;
  }

  export interface Props<T extends AppSelectOption<string | number>> {
    text: string;
    modelValue: T['value'] | undefined;
    options: readonly T[];
    disabled?: boolean;
  }

  defineProps<Props<VueGen>>();

  const emits = defineEmits<{
    (e: 'update:modelValue', modelValue: VueGen['value']): void;
  }>();

  function onChange (e: Event): void {
    const target = e.target as HTMLSelectElement;
    emits('update:modelValue', target.value);
  }
</script>

<template>
  <label>
    {{ text }}
    <select
      :value="modelValue"
      :disabled="disabled"
      @change="onChange"
    >
      <option
        v-for="option in options"
        :key="option.value"
        :value="option.value"
        :disabled="option.disabled"
      >
        {{ option.text }}
      </option>
    </select>
  </label>
</template>
```

This also allows to set the value as undefined in case you don want any pre-selected option.

But I'm also facing the  `The inferred type of 'default' references a type with a cyclic structure which cannot be trivially serialized. A type annotation is necessary. ts(5088)` TypeScript error and `@ts-ignore` doesn't help. Any suggestion on how to address it?

yyx990803:
To everyone running into the `cyclic structure cannot be trivially serialized` issue, could you please provide a reproduction, or at least the Vue / Volar / TypeScript versions? I cannot reproduce this locally.

Paul7Peterson:
@yyx990803 Here's a [minimum repo](https://github.com/Paul7Peterson/RFC-issue-reproduction/tree/init) with the error reproduction with the Select example I posted. 

Using TypeScript 4.9.5 locally from global.

h-sigma:
I'm unable to provide a repro ATM. I'll at-least confirm my versions: Typescript 4.9.5, Volar v1.2.0, Vue 3.2.45, Vue-Tsc 1.2.0.

Paul7Peterson:
> 

Thanks! Yeah, this really solved the issue! 💚

dakt:
Types lost when compiled as library.

In project:
image

As npm lib:
image

damafeez:
> * Upgraded typescript to 5.0.2 : Cyclic error still occurs.
> * Switching Volar + vue-tsc to pre-release version 1.3.4: Does get rid of the cyclic error in combination with the typescript version above. (My life is now complete)


I just upgraded typescript to `^5.0.2` (both globally and in-project), I upgraded volar and vue-tsc to `^1.3.4` and still get the  `The inferred type of 'default' references a type with a cyclic structure which cannot be trivially serialized. A type annotation is necessary.` error, sadly.

PR #1976, opened 2022-10-10
docs(vue-component-meta): updated README

jd-solanki:
As per our discord [chat](https://discord.com/channels/793943652350427136/1027819645677350912) I have updated `vue-component-meta` readme according to the bast of my knowledge.

I hope you like it.

### DIFF
--- a/vue-language-tools/vue-component-meta/README.md
+++ b/vue-language-tools/vue-component-meta/README.md
@@ -64,25 +64,6 @@ When you extract the component meta and extract the `description` property of th
 
 You can use it to document your component as you build your project without writing additional documentation.
 
-### Defining type in JavaScript projects
-
-You can use JSDoc's `@type` to define the prop type in JavaScript projects.
-
-```ts
-props: {
-  /**
-   * Array of texts to render as a message
-   * @type {import('vue').PropOptions<string[]>}
-   */
-  texts: {
-    type: Array,
-    required: false,
-  }
-}
-```
-
-`vue-component-meta` meta will extract the prop definition as "Array of texts to render as a message" and its type as `string[]`.
-
 ## Pitfalls 👀
 
 As `vue-component-meta` uses static code analysis, it can't extract the dynamic prop definition.
--- a/vue-language-tools/vue-component-meta/README.md
+++ b/vue-language-tools/vue-component-meta/README.md
@@ -1,10 +1,166 @@
 # vue-component-meta
 
-## Usage
+`vue-component-meta` allows you to extract the meta-data like props, slots, events, etc from your components via static code analysis. You can even generate description for your props from your source code. This helps document your components via automation. Please refer to the [reference](#reference) section for references.
 
-See https://github.com/johnsoncodehk/volar/blob/master/vue-language-tools/vue-component-meta/tests/index.spec.ts.
+## Guide 📗
 
-## Sponsors
+First of all, you need to create a component meta checker using `createComponentMetaChecker`:
+
+```ts
+import type { MetaCheckerOptions } from 'vue-component-meta'
+import { createComponentMetaChecker } from 'vue-component-meta'
+
+const checkerOptions: MetaCheckerOptions = {
+  forceUseTs: true,
+  schema: { ignore: ['MyIgnoredNestedProps'] },
+  printer: { newLine: 1 },
+}
+
+const tsconfigChecker = createComponentMetaChecker(
+  // Write your tsconfig path
+  resolve(__dirname, 'path-to-tsconfig'),
+  checkerOptions,
+)
+```
+
+Now, you can extract the component meta using `getComponentMeta` method of checker:
+
+```ts
+const componentPath = path.resolve(__dirname, 'path-to-component');
+const meta = checker.getComponentMeta(componentPath);
+```
+
+This meta contains really useful stuff like component props, slots, events and more. You can refer to its [type definition](https://github.com/johnsoncodehk/volar/blob/master/vue-language-tools/vue-component-meta/src/types.ts) for more details.
+
+### Extracting prop meta
+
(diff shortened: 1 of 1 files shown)


vue-language-tools/vue-component-meta/README.md
```diff
@@ -1,10 +1,166 @@
...
+
+const checkerOptions: MetaCheckerOptions = {
+  forceUseTs: true,
+  schema: { ignore: ['MyIgnoredNestedProps'] },
+  printer: { newLine: 1 },
+}
+
+const tsconfigChecker = createComponentMetaChecker(
+  // Write your tsconfig path
+  resolve(__dirname, 'path-to-tsconfig'),
```
rchl, line 21: Maybe also import `resolve` to make the example complete.

vue-language-tools/vue-component-meta/README.md
```diff
@@ -1,10 +1,166 @@
...
+
+You can use it to document your component as you build your project without writing additional documentation.
+
+## Pitfalls 👀
+
+As `vue-component-meta` uses static code analysis, it can't extract the dynamic prop definition.
+
+### default value
+
+`vue-component-meta` won't able to extract default value for prop as props can't be analyzed.
```
rchl, line 63:
```suggestion
`vue-component-meta` won't be able to extract default value for prop as props can't be analyzed.
```

vue-language-tools/vue-component-meta/README.md
```diff
@@ -1,10 +1,166 @@
...
+    variant: {
+      default: 'light',
+    },
+  }),
+}
+```
+
+### description
+
+Same as above scenario you might have issue with description not generating when prop definition is dynamic. In this case write prop description can be tricky.
```
rchl, line 101:
```suggestion
Same as above scenario you might have issue with description not generating when prop definition is dynamic. In this case writing prop description can be tricky.
```

rchl (commented):
Very nice!

rchl:
I haven't had a chance to use `vue-component-meta` yet but I'm assuming it's able to handle cases where prop is of `Object` type and more specific type is provided through JSDoc's `@type`? Would it be worth including an example for that?

jd-solanki:
It will be great for completion, but first, let me check if it's working fine in Anu as I don't have any playground to test it 😉 

Will be back soon.

Thanks for your suggestions 👍🏻

rchl:
Just an addendum to that... I meant an example using `@type` as that is what one would use in JS code. If inside TS code then it would be done with `as Foo` instead, I presume. You seem to be already assuming TS as you have one example using `as`. It might be beneficial to include examples for both TS and JS.

jd-solanki:
> Just an addendum to that... I meant an example using `@type` as that is what one would use in JS code. If inside TS code then it would be done with `as Foo` instead, I presume. You seem to be already assuming TS as you have one example using `as`. It might be beneficial to include examples for both TS and JS.

You mean like this:
```ts
props: {
  /*
   *
   * Array of texts to render as a message
   * @type {string[]}
  */
  texts: {
    type: Array,
    required: false,
  }
}
```

Is this enough?

Sorry if I didn't get your idea.

rchl:
> You mean like this:

Yes, something like this although I believe the correct type to use would be be:

```js
props: {
  /**
   * Array of texts to render as a message
   * @type {import('vue').PropOptions<string[]>}
   */
  texts: {
    type: Array,
    required: false,
  }
}
```

jd-solanki:
Thanks! Update the md.

vue-language-tools/vue-component-meta/README.md
```diff
@@ -1,10 +1,195 @@
...
+### Defining type in JavaScript projects
+
+You can use JSDoc's `@type` to define the prop type in JavaScript projects.
+
+```ts
+props: {
+  /*
+   * Array of texts to render as a message
+   * @type {import('vue').PropOptions<string[]>}
+  */
```
rchl, line 76: Sorry to be a pain but proper JSDoc comment would be formatted like:

```suggestion
  /**
   * Array of texts to render as a message
   * @type {import('vue').PropOptions<string[]>}
   */
```

rchl: Also I wonder if generated types are correct in this case....

jd-solanki: I haven't tried it as I am on TS projects always.

Do you want me to check on JS, Maybe I can do that tomorrow we will be leaving office soon.

Sorry for that JSDoc style.

rchl: I'm sorry but this doesn't seem to work. I've tried it on tests in this repo and for prop defined this way:

```
		/**
		 * Default function Array
		 * @type {import('vue').Prop<[1, 2, 3]>}
		 */
		arrayDefault: {
			type: Array,
			default: () => [1, 2, 3]
		},
```

gotten results like:

```
    {
      "name": "arrayDefault",
      "global": false,
      "description": "Default function Array",
      "tags": [
        {
          "name": "type",
          "text": "{import('vue').Prop<number[]>}"
        }
      ],
      "required": false,
      "type": "unknown[] | undefined",
      "schema": {
        "kind": "enum",
        "type": "unknown[] | undefined",
        "schema": [
          "undefined",
          {
            "kind": "array",
            "type": "unknown[]",
            "schema": [
              "unknown"
            ]
          }
        ]
      },
      "default": "[1, 2, 3]"
    },
```

So the resulting type is `unknown[]` which means the JSDoc annotation is not taken into account.

(I've used `Prop` type here as `PropOptions` doesn't seem to be exported in Vue 3.x anymore.)

So it's best to just remove this case from the readme. Sorry for the noise.

jd-solanki:
@rchl Done 😇

johnsoncodehk:
Great work, thank you!

PR #1987, opened 2022-10-11
feat: Support generic typed template slots for RFC 436

johnsoncodehk:
Update codegen for script setup to change export component type from `DefineComponent` to functional component.

For following SFC code:

```html
<template>
  <slot name="default" :test="props.msg"></slot>
</template>

<script lang="ts" setup generic="T">
const props = defineProps<{ msg: T }>()
</script>
```

Output type:

```ts
<T>(__props: { msg: T }): {
	props: typeof __props,
	children: { default(_: { test: T }): any },
}
```

`vue-component-meta` tests is broken because we not yet implement type extract for functional component in `vue-component-meta`.

### DIFF
--- a/vue-language-tools/vue-language-core/src/generators/script.ts
+++ b/vue-language-tools/vue-language-core/src/generators/script.ts
@@ -242,7 +242,7 @@ export function generate(
 					0,
 					{ diagnostic: true },
 				]);
-				codeGen.push('export default ');
+				codeGen.push('export default (');
 			}
 			if (vueCompilerOptions.experimentalRfc436 && sfc.scriptSetup.generic) {
 				codeGen.push(`<${sfc.scriptSetup.generic}>`);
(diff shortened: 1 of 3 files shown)


StepanMynarik:
Updated to 1.0.4 and the thing works PERFECTLY.
Well done!

StepanMynarik:
I noticed one issue that stems from this I believe.

In the following code, 'typeof BButton' is highlighted with error: "Type '(__VLS_props: { text: string; preset?: Preset | undefined; size?: Size | undefined; disabled?: boolean | undefined; active?: boolean | undefined; }) => Omit<Element, "props" | "children"> & Omit<...> & { ...; }' provides no match for the signature 'new (...args: any): any'."
```typescript
const referenceRef = $ref<InstanceType<typeof BButton> | null>(null);
```

johnsoncodehk:
@StepanMynarik You should use `ReturnType<typeof BButton>` instead of `InstanceType<typeof BButton>` for components gen with `experimentalRfc436` enabled.

colinj:
### Generic-typed slots in child components do not work when inside a parent component that is also using generic types.

I have a generic list component that has a default slot that exposes the `item: T` property in a scoped slot. This works perfectly when the parent component is a non-generic component. 
When I hover over `v-slot` it displays the type defined by the `items` prop. eg `{ name: string; age: number; }`
```vue
<MyList :items="listOfPeople" v-slot="{ item }">
  {{ item.name }} is {{ item.age }} years old.
</MyList>
```

But when I try to use the same generic component inside of another generic component, TypeScript displays an error over the scoped-slot in the child component.

```vue
<script setup lang="ts" generic="T extends string | Record<string, unknown>">
  defineProps<{ items: T }>();
  const display =  (item: T): string => { return ... };
....
</script>
<template>
  <MyList :items="items" v-slot="{ item }"> <!-- TypeScript error here -->
    {{ display(item) }}
  </MyList>
</template>
```

> Element implicitly has an 'any' type because expression of type '"default"' can't be used to 
> index type `Record<string, any> | {} | { default: { item: T; }; }`. 
> Property 'default' does not exist on type `Record<string, any> | {} | {  default: { item: T; }; }`.

I'm not sure why, but it seems that the slot is incorrectly typed and it only happens when the parent component includes the `generic` attribute in the `script setup`.
When comparing to the union type, TypeScript says the `default` property does not exist on the type which I think is due to the inclusion of the empty object `{}`.

johanninos:
> @StepanMynarik You should use `ReturnType<typeof BButton>` instead of `InstanceType<typeof BButton>` for components gen with `experimentalRfc436` enabled.

`InstanceType<typeof BButton<T>>` can pass generic.
but `ReturnType<typeof BButton<T>>` wasn't supported untill typescript 4.7.0

but vue has bug with this ts new feature

https://github.com/vuejs/core/issues/7161

how to deal with that?

johanninos:
```
  const F = () => BButton <string>('');
  const btnRef = ref<ReturnType<typeof F> | null>(null);
```
this works, but a bit wasteful

jaulz:
Does this actually also work with `defineComponent`?

PR #2056, opened 2022-10-24
chore: replace Math.pow with (<<) for base 2

Fuzzyma:
The PR of this tweet:

https://twitter.com/johnsoncodehk/status/1584181852595974144

I replaced Math.pow with left shifts because they are substantially faster


johnsoncodehk:
That is correct but you're late. 😝
Duplicate of #2053

PR #2187, opened 2022-12-09
fix(v-bind): allow duplicate keys via v-bind

sawmurai:
This PR might fix https://github.com/johnsoncodehk/volar/issues/2166. It does fix the error that entries of `v-bind` collide with already bound attributes.

An example of where this is wanted would be a slot that v-binds a name property:

```vue
<slot name="the-slot-name" v-bind="{name: 'Some name'}"></slot>
```

I would not rule out that I missed something obvious, so feel free to just close this PR in favor of a better fix.


johnsoncodehk:
Fixed by https://github.com/johnsoncodehk/volar/commit/275a95f8a357f4247fa8023eed17364691ecbd63, and thanks. :)

PR #2293, opened 2023-01-08
refactor: move framework core out of repo

johnsoncodehk:
Before this PR, this repo was refactored into three main components: Framework, Core Plugins, and Vue Tools. Including these in a single repo helps with rapid development, but leads to unclear focus of code management, which may deter some potential contributors.

Secondly, splitting the project also helps to better think about how to organize and write tests, which is the main missing part of us in the past.

This PR does something very similar to https://github.com/vitejs/vite/pull/11158, so you can take a look at the other benefits that this PR points out.

## Tasks

- [x] Move pug-language-tools to https://github.com/volarjs/pug-language-tools
- [x] Move move packages and plugins to https://github.com/volarjs/framework
- [x] ~~vue-language-tools -> packages~~


johnsoncodehk:
I decided not to move vue-language-tools to packages for now, because I plan to move the repo to vuejs in the future, and keep this change until then to avoid links that can break the documentation twice.

PR #2338, opened 2023-01-26
chore: should be using Lerna-Lite schema instead of Lerna schema

ghiscoding:
since you're using Lerna-Lite, it would be better to use Lerna-Lite's JSON Schema instead of Lerna's Schema which is slightly different. Also note that I'm the maintainer of Lerna-Lite ;)

### DIFF
--- a/lerna.json
+++ b/lerna.json
@@ -1,5 +1,5 @@
 {
-	"$schema": "https://raw.githubusercontent.com/lerna/lerna/main/core/lerna/schemas/lerna-schema.json",
+	"$schema": "https://raw.githubusercontent.com/lerna-lite/lerna-lite/main/packages/cli/schemas/lerna-schema.json",
 	"npmClient": "pnpm",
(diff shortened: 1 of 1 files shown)


ghiscoding:
@johnsoncodehk side note, I recently added a new, and optional, [`lerna watch`](https://github.com/lerna-lite/lerna-lite/tree/main/packages/watch#readme) command in Lerna-Lite which can be helpful in a monorepo. It could be useful if you want to run build, update tsc types or run test on certain packages and and propagate to their dependents and/or dependencies (see the multiple [examples](https://github.com/lerna-lite/lerna-lite/tree/main/packages/watch#examples)). It's also ok to ignore my message ;)  

Happy Chinese New Year ✨🧧

johnsoncodehk:
LGTM, thank you! I will try the new command for a while.

And also thanks for your work on `lerna-lite`, it's really great. Happy New Year!

PR #2371, opened 2023-02-09
fix: `genConstructorOverloads` when nums is 0

### LINKED ISSUE #2370: `defineEmits<{}>()` cannot pass type-checking
# Problem
Declaring empty emits like `defineEmits<{}>()` would fail the type-checking process.

# Reproduction
https://stackblitz.com/edit/vitejs-vite-ibxwhy?file=src/components/HelloWorld.vue

**Steps**
Run `yarn type-check`.

# Environment
vue@2.7.14, vue-tsc@1.0.24
(Also tested on vue@3, no errors but still get wrong types in template, where the `msg` is inferred as of type `never`.)

lsdsjy:
Fixes #2370.

As for now Volar generates code like this when `defineEmits` is used with empty interface:
```ts
type __VLS_ConstructorOverloads<T> =
// 0
T extends {
} ? (
) :
// 0
{};
```
which is not valid TypeScript code. This PR skips the `nums == 0` case.

### DIFF
--- a/packages/vue-language-core/src/generators/script.ts
+++ b/packages/vue-language-core/src/generators/script.ts
@@ -791,7 +791,7 @@ function genConstructorOverloads(name = 'ConstructorOverloads', nums?: number) {
 			gen(i);
 		}
 	}
-	else {
+	else if (nums > 0) {
 		gen(nums);
 	}
 	code += `// 0\n`;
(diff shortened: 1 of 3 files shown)


johnsoncodehk:
Thanks for the PR! Could you also add a test in https://github.com/vuejs/language-tools/tree/master/packages/vue-test-workspace/vue-tsc?

lsdsjy:
Done. Thanks for the suggestion. Otherwise I wouldn't notice that I've missed another `genConstructorOverloads` declaration.😅

johnsoncodehk:
Thanks!

PR #2400, opened 2023-02-17
Update README.md

Nothing-Works:
The original description is not accurate. The original one feels like it's not recommended to use this plugin, but what we want to say is that use this plugin but with take-over mode, also changed the link to the VUE docs section


johnsoncodehk:
Thanks!

Issue #2458, opened 2023-02-26
Volar auto import not working with kebab-case


Wissiak:
Volar seems to only detect auto import components when using PascalCase:
![image](https://user-images.githubusercontent.com/31206249/221407304-f5543b2c-db64-470f-b1f1-32cba6ca4487.png)

As per the docs I recently disabled Vetur and since then the auto import components are not correctly shown if I use kebab-case. Here, no suggestion comes up:
![image](https://user-images.githubusercontent.com/31206249/221407384-f7c4deae-56ea-4d3a-a78e-b3b7fc4624f1.png)

I've already tried the following settings but they did not fix the bahavior:
![image](https://user-images.githubusercontent.com/31206249/221407438-c009affe-2cbe-4111-bf7e-3dad0aee579d.png)

Did I miss some setting or is there a bug?

Chriztiaan:
I am also having trouble with this.

WangJincheng4869:
这个bug何时修复发版？目前这个bug严重影响开发效率

KiddoV:
I am still having the problem. I am on Volar v1.8.25

TurningTide:
Same with v2.1.10

BrevAlessioMk:
Same with 2.2.4

crschmidt2:
Using neovim, I was able to get kebab-casing working on completions by using the workspace/didChangeConfiguration option: 
```lua
vue.complete.casing.tags = "kebab"
```

For neovim users, configuring this with lspconfig looks like the following:
```lua
require('lspconfig').volar.setup {
    settings = {
        vue = {
            complete = {
                casing = {
                    tags = "autoKebab"
                }
            }
        }
    }
}
```

PR #2460, opened 2023-02-26
fix: Tag and attribute name case do not take effect in real time when they are configured as `kebab`

Sight-wcg:
fix: #2458

在 vscode 设置面板，将 `preferredTagNameCase` 或 `PreferredAttrNameCase` 从其他选项更改为 `kebab` 时，不会更新文件的  nameCase，此时需要关闭并重新打开选项卡或重启 volar 才能生效。


Chriztiaan:
[johnsoncodehk](https://github.com/johnsoncodehk) can we please get this in?

johnsoncodehk:
Thanks for the PR, but this will cause the name casing status selected by the user to not be saved when switching documents. Alternative by https://github.com/vuejs/language-tools/commit/b37a2621f7194dafa574185167a94a3e86b058c5.

PR #2471, opened 2023-02-28
workaround: require.resolve bug in pnpm

dschmidt:
This makes it possible to use a tsconfig package from the local pnpm workspace (while keeping it possible to use any other package installed from a registry).

We implemented it as a local patch a few weeks ago but the file has been changed around quite a bit recently and I've now come to the conclusion it might be worth to upstream the patch.

Let me know what you think, thanks in advance!


dschmidt:
Yes, not today, but will do so tomorrow. Should be fairly trivial.

dschmidt:
Here we go: https://github.com/dschmidt/vue-tsc-pnpm

While preparing and double checking that I realized I had messed up this patch a bit but it definitely works now for the package-name case.
On discord you told me this would break for file names, which was a very good catch. I can fix it, if you let me know which way you prefer.
1) trying to be smart by analyzing the `extendsPath`
2) just use this as a fallback if the first `require.resolve` fails
3) Something else

rchl:
I now see `Failed to resolve tsconfig path: /.../tsconfig.json` in a standard project using `npm`.

johnsoncodehk:
@rchl After this PR we gave up on resolving extends path by ourselves, but steal the extends resolved result from typescript. Have you encountered a situation where extends is valid for compilerOptions, but not for vueCompilerOptions?

rchl:
My project doesn't use any extends. It's a simplest one you can get with plain tsconfig

johnsoncodehk:
Could you log `err` at the following code in `node_modules/@volar/vue-language-core/out/utils/ts.js`, and see what is the error message?

https://github.com/vuejs/language-tools/blob/faa9f52fb4b0155fa238c9f08ae7bf5c494e4849/packages/vue-language-core/src/utils/ts.ts#L66

rchl:
```
LSP-volar: TypeError: host.readDirectory is not a function
LSP-volar:     at getFileNamesFromConfigSpecs (/.../node_modules/typescript/lib/typescript.js:42711:40)
```

johnsoncodehk:
Could be fixed by https://github.com/vuejs/language-tools/commit/6187ab13ea82b4edb99c6dbe722c65824120e92a, thanks to let me know.

rchl:
Seems to work now.

PR #2483, opened 2023-03-03
refactor: simplify the code for loading html data.

Sight-wcg:


johnsoncodehk:
Thanks for the PR, I think but it can't be simplified, which is for ESBuild to be able to resolve the json paths.

Sight-wcg:
> 

Sorry, I didn't realize this, I will close it.

Issue #2495, opened 2023-03-08
Idea: Extract to new SFC component refactoring


zardoy:
Code action which would allow us to quickly select `<template>` tag to extract it into a new file.
There is an obvious motivation for this: split one bloated component into a smaller ones. And automatic extract action can save a lot of time when doing this repetive task.
TypeScript already has builtin code action to extract selection into function or a new file, so I don't see a reason to not also provide code action of such kind out of the box here.

Note: there are already community solutions like [Vue Extract Component](https://marketplace.visualstudio.com/items?itemName=dsalex1.vue-extract-component). However the feature must be integrated with access to TS language service so its performant and can update imports, no doubt it should be implementated as plugin (I see Volar already has plugins).

I also plan to move content not only from template, but also from script e.g. for this:

```vue
<script setup lang="ts">
import { ref } from 'vue'
const text = ref('')

const handleClick = () => {
    text.value = 'replaced from click'
}
</script>
<template>
    <input :value.sync="text" />
    <button @click="handleClick">{{ text }}</button>
</template>
```

Extracting with `<button>` selection would move not only button from template, but also handleClick into a new file component (with possibly updating imports like TypeScript's [moveToNewFile](https://github.com/microsoft/TypeScript/blob/e1ae30e1abd29ba2a83cd497745b29169b202c11/src/services/refactors/moveToNewFile.ts#L290) do).

PR #2496, opened 2023-03-08
Extract into simple component refactoring

zardoy:
ref #2495 

I'm attaching gif so you can easier to understand what is happening there:

<details>
<summary>GIF</summary>

[code](https://gist.github.com/zardoy/938dde340e547610221e427ab0dc4f2e)
</details>

Please let me know wether you are interested in the idea itself (and probably having some help here ;)

Updated questions:

- I didn't find a way to get needed `virtualFile` mapping for specific interpolation range (eg v-bind directive). When I figure it out I will remove `templateFormatScript` usage in favor of `templateAst` so there is a better detection wether identifier should be written to `emits`. Sorry, for me its really hard to understand what is happening in template's codegen 😆
In other words I currently map over generatedRanges and filter only identifiers and while it works in most cases I think it can give unpredictable results. 
- It seems that `edit.changes` cant be used along with `documentChanges`, is this intentional? So how to update current file and create a new one with code action then? On the other hand I think we should ask a name for component before, so it means moving to command with arguments like `edits` and `addImportPosition`. Or we can ask for component name after its creation and do file rename then for now.
- I guess this is fine if I use ChangesTracker from TS API? For context: it is not exposed with typings, but TS code actions using it. It is needed for better formatting with factory API and handling import fixes.

todo:
- handle import adding for types, like TS infer from usages [do](https://github.com/microsoft/TypeScript/blob/ca1cf3dd8595b3e9a88e35cf1f8a203272e8ae57/src/services/codefixes/inferFromUsage.ts#L315)
- support options API

Also if you're happy with idea, settings should be added to allow configure creating new components (e.g. force composition or options API to use)


johnsoncodehk:
Hi, can you update the questions it their status has changed? I will answer later.

zardoy:
> Hi, can you update the questions if their status has changed? I will answer later.

Done.

It also would be nice to normalize generated `\t`, but I will do it like https://github.com/volarjs/volar.js/pull/20 is done, once issue with creating files from code action is resolved

johnsoncodehk:
1. It seems a same question to https://github.com/vuejs/language-tools/pull/2524#issuecomment-1482122836?
2. We have no restriction to use `edit.changes` and `documentChanges` at the same time, probably a limitation of the IDE itself. I prefer to add a rename file command to the code action.
3. Can you share some references about `ChangesTracker`? I don't find it in TypeScript source code.

The current code is a bit difficult to review. It would be great if it can be abstracted into multiple functions according to the intention. I will be happy to help realize it.

hkochniss:
@zardoy 
hope this doesn't go to "sleep". This is literally the one refactor that's missing from Volar (in comparison to let's say WebStorm)

zardoy:
@hkochniss For vscode specifically you can check [Vue Extract Component](https://marketplace.visualstudio.com/items?itemName=dsalex1.vue-extract-component)

As you noticed I currently don't work on this, this pr requires some debugging & issues resolving. Probably going to spend some time later, but for now I don't see a priority for this

johnsoncodehk:
I will look into this and #2524 in v1.5.x.

johnsoncodehk:
I think the functionality is working fine now, let's merge it.

johnsoncodehk:
> * handle import adding for types, like TS infer from usages [do](https://github.com/microsoft/TypeScript/blob/ca1cf3dd8595b3e9a88e35cf1f8a203272e8ae57/src/services/codefixes/inferFromUsage.ts#L315)

We can try setting addMissingImports command for `codeAction.command` to do it.

> * support options API

Let's ditch the options api for this functionality, the current workload is already a bit unmaintainable.

johnsoncodehk:
Options API is also supported by https://github.com/vuejs/language-tools/commit/f012a5715baf1adaeccb5193a4c467a5ff749213, I found that we already have the code to modify the `components` option.

zardoy:
> Options API is also supported by f012a57, I found that we already have the code to modify the components option.

omg thank you so much, now this code action would save a lot of time in big legacy code bases that use options api!

Issue #2425, opened 2023-02-20
refactor extract to interface is a problem with the functionality of


rxliuli:
demo

![Kapture 2023-02-20 at 12 42 23](https://user-images.githubusercontent.com/24560368/220011255-3e6fe0fd-72f8-4126-ab20-fcd66d8117dd.gif)

PR #2498, opened 2023-03-08
Allow code actions to run commands after applying

zardoy:
ref vuejs/language-tools#2425 (no rename action after applying)

Linked pr: https://github.com/volarjs/plugins/pull/29

Of course current implementation isnt ideal and should be extended, but I hope you get the idea.

`resolveCodeAction` can only change `edit` of code action, so its useless to change `command` property of code action after `getEditsForRefactor` (in linked pr).

Obviously, we could just run commands right in `resolveCodeAction`, but it can be called from API and it should not have side effects: `vscode.commands.executeCommand('vscode.executeCodeActionProvider', uri, rangeOrSelection, kind, itemResolveCount)`.

Just for reference: typescript builtin extension doesn't use edits at all, it provides only `command` property without resolve stage at all.

Also I noticed another general problem with passing vscode.Position and vscode.Uri from server as vscode API uses instanceof checks and these arguments don't work.

Feel free to close it favor of your implementation, as I currently don't have ideas how it can be improved 🤔

P.S. As I understand it is generally bad idea to use middlewares here right?


zardoy:
I first thought linked issue is about getting rename box after code actions, but its about `formatOptions: vscode.FormattingOptions` not passing to `formatOptions`

johnsoncodehk:
Look to [LSP Code Action Resolve Request Specification](https://microsoft.github.io/language-server-protocol/specifications/lsp/3.17/specification/#codeAction_resolve), we need to check if `textDocument.codeAction.resolveSupport.properties` includes `'command'`. If not, `@volar-plugins/typescript` should execute `getEditsForRefactor()` in advance to get `edit` + `command`. I'm not sure how much this affects performance, and how many IDEs implement resolve support for `'command'`, I think it would be nice to ask VSCode (and other IDEs) to implement it.

I will implement the judging `textDocument.codeAction.resolveSupport.properties` way first.

johnsoncodehk:
I think it's mostly working, but there are still two hard problems.

1. `renameLocation` is based on the location after editing, but source map mapping is based on the text before editing, so `renameLocation` mapping may be wrong. To solve it, it seems that you need to perform mapping in the language server apply edit in advance. At present, I just ignore this problem for the time being.

2. Since Formatting Request and Code Actions Request are executed by different language servers, FormattingOptions cannot be shared.

zardoy:
> If not, @volar-plugins/typescript should execute getEditsForRefactor() in advance to get edit + command.

Hey, @johnsoncodehk as I understand now https://github.com/volarjs/volar.js/blob/033129d71b355a0f5bbf514c6887ed3a6e083c21/packages/vscode/src/index.ts#L36-L42 now never runs, and I believe asking for applicable code actions and then for edits for each code action is just huge performance regression, why we didn't go with some kind of solution that still uses code action resolve? `getEditsForRefactor` does all the work that `getApplicableRefactors` does + some little extra work, so now we can gurantee that for top level nodes now showing code actions at least 2x slower.

> Since Formatting Request and Code Actions Request are executed by different language servers, FormattingOptions cannot be shared.

Hm, why don't use default format settings at least?

zardoy:
I also see [`resolveCodeAction`](https://github.com/johnsoncodehk/volar/blob/814a03fb52f3f1f4bfb8ac17c6953d4db652c32d/packages/vue-language-service/src/languageService.ts#L226) in vue-language-service now runs only for import all actions, which means that code actions such as import quickfix or extract to const now don't insert edits on second line after `<script>`.

johnsoncodehk:
> why we didn't go with some kind of solution that still uses code action resolve?

Because the best solution is to default to working with all IDEs rather than being specific to VSCode. In my testing, it didn't have too much impact on performance.

However, I am still looking for a way that doesn't require middleware.

PR #2506, opened 2023-03-11
perf(vue-component-meta): delay init for global props list

stafyniaksacha:
Setting `globalPropNames` avoid parsing empty component `<script setup lang="ts"></script>` while creating new meta parser. which can save seconds when starting!

In actual vue 3.2, the resolved value is `[ 'key', 'ref', 'ref_for', 'ref_key', 'class', 'style' ]`

### DIFF
--- a/packages/vue-component-meta/src/index.ts
+++ b/packages/vue-component-meta/src/index.ts
@@ -183,7 +183,7 @@ export function baseCreate(
 	});
 	const tsLs = ts.createLanguageService(proxyHost);
 	let globalPropNames: string[] = [];
-	globalPropNames = getComponentMeta(globalComponentName).props.map(prop => prop.name);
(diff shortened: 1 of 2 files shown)


johnsoncodehk:
Hi, thanks for pointing this out, I tried moving `getComponentMeta(globalComponentName)` to `getProps` to avoid initialization performance issues, this should bypass the extra option.

stafyniaksacha:
Delay initialization is a good thing, but I also think we should let the option: resolving this array can take up to 5 seconds. That would be great to be able to set it

johnsoncodehk:
I think the 5 seconds saved is actually subsumed into the first call of check component meta, so removing this step almost no performance improvement in terms of results.

This is just guesswork from what I know, if you find this does improve performance please let me know.

johnsoncodehk:
I just merged it first to release and be able to test, if you find that skipping checking global props improves performance, will be great to have a new PR.

Issue #2087, opened 2022-11-02
✨ Ability to toggle v-bind on attribute/prop via keyboard shortcut 


jd-solanki:
Hi 👋🏻

As volar is now everyday use, I found something which we can improve.

Whenever I hit the attribute autocompletion I would like to attach `v-bind` via shortcut `:` to the attribute for additional suggestions for value.

It will be great if we can provide keyboard shortcut to toggle `v-bind`/`:`

https://user-images.githubusercontent.com/47495003/199463267-fd6a9fa4-6412-4413-ac44-96c1427ba729.mov

Thanks for Volar and additional packages 😇

PR #2524, opened 2023-03-19
Add `v-bind` code action

zardoy:
Fixes https://github.com/vuejs/language-tools/issues/2087

Quotes are not added by design, as the code action intended for cases where user meant an identifier

So @jd-solanki could add the following keybinding:

```json
{
    "key": "...",
    "command": "editor.action.codeAction",
    "args": {
        "kind": "refactor.rewrite.addVBind",
        "apply": "ifSingle"
    }
}
```

IMO its worth also adding `:` when `:` typed after attribute name, it would be faster for me rather than adding additional keybinding.


johnsoncodehk:
Thanks for the PR! Are you planning to change the code action to two-way conversion?

> IMO its worth also adding `:` when `:` typed after attribute name, it would be faster for me rather than adding additional keybinding.

This is the implementation, but really not sure we should have it.

```ts
provideAutoInsertionEdit(document, position, insertCtx) {

	if (insertCtx.lastChange.text !== ':') return;

	const offset = document.offsetAt(position);
	const [virtualFile, source] = ctx!.documents.getVirtualFileByUri(document.uri);
	if (!virtualFile || !(source.root instanceof VueFile)) {
		return;
	}

	const { templateAst } = source.root.sfc;

	if (!templateAst) return;

	let result: vscode.TextEdit | undefined;

	walkElementNodes(templateAst, node => {
		for (const prop of node.props) {
			if (
				prop.type === NodeTypes.ATTRIBUTE
				&& offset === prop.loc.start.offset + prop.name.length
			) {
				result = {
					newText: ':' + prop.name.slice(0, -1) + '$0',
					range: {
						start: document.positionAt(prop.loc.start.offset),
						end: document.positionAt(prop.loc.start.offset + prop.name.length),
					},
				};
			}
		}
	});

	return result;
}
```

zardoy:
> Are you planning to change the code action to two-way conversion?

I was thinking of adding it, but only value is string: `:type="'password'"` -> `type="password"`, but it would require to walk every node used within v-bind attribute and I currently don't know how to do it (tried but failed in https://github.com/vuejs/language-tools/pull/2496)

I was also thinking of adding `="$0"` for current code action when attribute have no value.

> This is the implementation, but really not sure we should have it.

If you're against of having this feature, I don't think I need implementation, but thanks. Though I don't see any downsides of having it e.g. typing `:` after attribute name currently don't have correct usages, right? Anyway, its just a proposal: we don't think it will be useful - let's don't add it here.

packages/vue-language-service/src/plugins/vue-toggle-v-bind-codeaction.ts
```diff
@@ -0,0 +1,91 @@
...
+				}
+
+				const { templateAst, template } = vueFile.sfc;
+
+				if (!templateAst) return;
+
+				const templateStartOffset = template!.startTagEnd;
+				const result: vscode.CodeAction[] = [];
+
+				walkElementNodes(templateAst, node => {
```
zardoy, line 27: As I understand `walkElementNodes` would walk literally every single node in AST isn't it bad for performance?

zardoy: Oh, we don't really do any heavy computation here, so nevermind. Will remove findTemplateNode then

zardoy:
How its possible code action to use snippet edit? Want to move cursor inside added empty value.

rchl:
> How its possible code action to use snippet edit? Want to move cursor inside added empty value.

You can't. https://github.com/microsoft/language-server-protocol/issues/592

BTW. This probably would be more useful if it would be automatically provided in relevant contexts (whether cursor is withing the attribute's key or value. Then it wouldn't require custom key bindings and could be used by any LSP client by just triggering the default key binding for showing code actions.

zardoy:
> This probably would be more useful if it would be automatically provided in relevant contexts (whether cursor is within the attribute's key or value. Then it wouldn't require custom key bindings and could be used by any LSP client by just triggering the default key binding for showing code actions.

Sorry, didn't get what you mean. It is already using regular code action provider that doesn't require any additional keybindings in vscode and I guess any lsp client can use it?

Also
> whether cursor is within the attribute's key or value

Doesn't the current impl already do it?

rchl:
> Doesn't the current impl already do it?

Hmm, right. I guess it should. But I can't get this plugins' `provideCodeActions` implementation to trigger in my testing.

I'm testing on a Vue 2.7 project, if that matters.

johnsoncodehk:
> How its possible code action to use snippet edit? Want to move cursor inside added empty value.

I think you must need to use `CodeAction.command` to insert a `=`, and then language servers will trigger `doQuoteComplete()`, insert `""` and make the cursor in it.

packages/vue-language-service/src/languageService.ts
```diff
@@ -271,6 +272,7 @@ function resolvePlugins(
 	plugins['pug-beautify'] ??= createPugFormatPlugin();
 	plugins.json ??= createJsonPlugin(settings?.json);
 	plugins['typescript/twoslash-queries'] ??= createTsTqPlugin();
+	plugins['vue/toggleVBindCodeaction'] = createToggleVBindCodeaction();
```
rchl, line 275: All other plugins use `??=`. I guess to avoid creating plugin multiple times.

rchl:
The issue with not seeing this code actions is due to some bigger issue with Volar itself. Between 1.2.1 and 1.3.0 it just stopped functioning properly and now it stalls after sending diagnostics...

zardoy:
> > How its possible code action to use snippet edit? Want to move cursor inside added empty value.
> 
> I think you need to use `CodeAction.command` to insert `=`, and then language servers will trigger `doQuoteComplete()` to insert `""` and make the cursor in it.

Done, but it would really better to use SnippetEdit in future as now it requires to undo twice to get the same name

nekocom3745 (approved):

packages/vue-language-service/src/plugins/vue-toggle-v-bind-codeaction.ts
```diff
@@ -0,0 +1,78 @@
+import { LanguageServicePlugin } from '@volar/language-service';
+import { VueFile, walkElementNodes } from '@volar/vue-language-core';
+import { NodeTypes } from 'packages/vue-language-core/src/utils/vue2TemplateCompiler';
```
johnsoncodehk, line 3: `packages/vue-language-core/src/utils/vue2TemplateCompiler` is not a valid path in runtime.

zardoy: Indeed, have no idea why auto-import had to point me to src instead of out

jd-solanki:
Thanks @zardoy for the implementation

Does this mean, we can now add `v-bind` or `:` (shorthand) to existing attribute via hotkey?

That's super helpful 🚀

rchl:
Tried to test it a little:

https://user-images.githubusercontent.com/153197/226274381-0e410bca-88c9-4a22-9513-13ce724cb82a.mov


```
:: [08:30:11.320] <<< LSP-volar (13) (duration: 7ms): [{'edit': {'changes': {'file:///usr/local/workspace/temp/nuxt-2-async-data/pages/indexWithObjectProps.vue': [{'range': {'end': {'character': 20, 'line': 2}, 'start': {'character': 12, 'line': 2}}, 'newText': ''}]}}, 'title': 'Add v-bind to attribute', 'kind': 'refactor.rewrite.addVBind', 'data': {'original': {'edit': {'changes': {'file:///usr/local/workspace/temp/nuxt-2-async-data/pages/indexWithObjectProps.vue': [{'range': {'end': {'character': 20, 'line': 2}, 'start': {'character': 12, 'line': 2}}, 'newText': ''}]}}}, 'pluginId': 'vue/toggleVBindCodeaction', 'uri': 'file:///usr/local/workspace/temp/nuxt-2-async-data/pages/indexWithObjectProps.vue', 'type': 'plugin', 'version': 1}, 'command': {'title': '', 'arguments': [{'text': ':disabled='}], 'command': 'type'}}]
:: [08:30:13.423] --> LSP-volar workspace/executeCommand (14): {'arguments': [{'text': ':disabled='}], 'command': 'type'}
:: [08:30:13.424] <~~ LSP-volar (14) (duration: 1ms): {'message': 'Unhandled method workspace/executeCommand', 'code': -32601}
```

Is this change tying it's functionality to vscode extensions? Because it doesn't seem like the server supports `type` workspace command.

zardoy:
> But I guess you meant "editors" and not "language servers" and more like "VSCode" rather than "editors"

You're absolutely right, but I didn't find a way to figure out which client made request (e.g. vscode) from `ctx.capabilities`, so I can use workaround for vscode only.

@johnsoncodehk I think it'd be better to provide `ctx.commands.createSnippetEditCommand` for vscode >= 1.72.0, which would register custom command to apply `vscode.SnippetTextEdit[]` but it requires changes in other repos... Or I can just revert https://github.com/vuejs/language-tools/pull/2524/commits/ac6e3fa59bac52c60e0c30e47385092a87922c32

rchl:
You shouldn't be making workarounds for specific editors and I hope @johnsoncodehk thinks the same.

If you must make something vscode-specific then do it in the vscode extension. You can probably even intercept commands and add whatever you want there.

Think of creating a behavior that is best for all LSP-complient clients.

I wouldn't consider it a breaking issue if the user has to move the cursor manually to the attribute value after triggering this code action.

zardoy:
> If you must make something vscode-specific then do it in the vscode extension.

Isn't `ctx.commands` are already vscode-specific?

> I wouldn't consider it a breaking issue if the user has to move the cursor manually to the attribute value after triggering this code action.

Agree, it's not breaking, but could be a huge ux improvement, but I see logical limitation in code actions support, so I reverted that commit with workaround.

rchl:
> Isn't `ctx.commands` are already vscode-specific?

I wasn't aware of those but yes, those are vscode specific. Sublime Text implemented some like that (`editor.action.triggerSuggest`, `editor.action.triggerParameterHints`) because many MS-made servers use those. But it's still not cool.

johnsoncodehk:
> How its possible code action to use snippet edit? Want to move cursor inside added empty value.

I use `setSelection` VSCode command to support it. For other IDEs the cursor just has no move, I think it's fine.

We can merge the PR when you implement "Toggle".

zardoy:
> We can merge the PR when you implement "Toggle".

Do you mean *Remove v-bind* refactoring for this case?

```vue
<Test :prop="'string'" /> -> <Text prop="string" />
```

will do!

zardoy:
Hey @johnsoncodehk, I was thinking of the following detection for this code action:
- find node at position of v-bind directive
- check this node is string and end == v-bind directive end

But I'm not sure how to obtain correct range of sourceFile. I tried to add this to `for (const prop of node.props) {`:

```ts
if (
	startOffset - templateStartOffset >= prop.loc.start.offset
	&& endOffset - templateStartOffset <= prop.loc.end.offset
	&& prop.type === NodeTypes.DIRECTIVE && prop.exp
) {
	for (const [_, map] of ctx!.documents.getMapsByVirtualFileUri(document.uri)) {
		const r = map.toGeneratedRange({
			start: document.positionAt(prop.exp.loc.start.offset),
			end: document.positionAt(prop.exp.loc.end.offset)
		});
		if (!r) continue;
		const ts = ctx!.typescript!.module;
		const sourceFile = ctx!.typescript!.languageService.getProgram()!.getSourceFile(_.mainScriptName)!;
		const startOffset = ts.getPositionOfLineAndCharacter(sourceFile, r.start.line, r.start.character);
		const endOffset = ts.getPositionOfLineAndCharacter(sourceFile, r.end.line, r.end.character);
		const generatedText = sourceFile.getFullText().slice(startOffset, endOffset);
		// or the same, but much simpler
		const generatedText = map.virtualFileDocument.getText(r)
	}
}
```

But this doesn't seem to work. Is there an easy way to get that range in virtual script file?

johnsoncodehk:
@zardoy Sorry I probably didn't understand your request, is this what you're asking for?

```ts
if (
	startOffset - templateStartOffset >= prop.loc.start.offset
	&& endOffset - templateStartOffset <= prop.loc.end.offset
	&& prop.type === NodeTypes.DIRECTIVE && prop.exp
) {
	const nodeAtPosition = prop;
	const nodeIsString = (prop.exp.loc.source.startsWith('"') && prop.exp.loc.source.endsWith('"'))
		|| (prop.exp.loc.source.startsWith("'") && prop.exp.loc.source.endsWith("'"));
	const isEndOfDirective = endOffset - templateStartOffset === prop.arg?.loc.end.offset;
}
```

packages/vue-language-service/src/plugins/vue-toggle-v-bind-codeaction.ts
```diff
@@ -0,0 +1,77 @@
...
+									character: addValuePos.character + ':'.length + '="'.length,
+								};
+							}
+							result.push({
+								title: 'Add v-bind to attribute',
+								kind: 'refactor.rewrite.addVBind',
+								edit: {
+									changes: { [document.uri]: edits },
+								},
+								command: newPosition ? ctx?.commands.createSetSelectionCommand(newPosition) : undefined,
```
rchl, line 67: It's still using a custom command which will trigger a user facing error in form a dialog box because this is how ST handles invalid commands. I don't know about how other editors handle unsupported commands.

zardoy:
@johnsoncodehk This is too naive. I wanted to use TypeScript API to do this check: Let's say I have `:foo="'a' + b + 'c'"` your check will pass, but we could get virtual range instead (`'a' + __VLS_ctx.b + 'c'`) and with TS API do this check:

> check this node is string and end == v-bind directive end

packages/vue-language-service/src/plugins/vue-toggle-v-bind-codeaction.ts
johnsoncodehk: I will add an option to let the language client prompt for a list of supported commands.

johnsoncodehk:
@zardoy Since `:foo="'a' + b + 'c'"` will create multiple mappings, there is only one verbose way to do it.

image

```ts
if (
	startOffset - templateStartOffset >= prop.loc.start.offset
	&& endOffset - templateStartOffset <= prop.loc.end.offset
	&& prop.type === NodeTypes.DIRECTIVE && prop.exp
) {
	for (const [tsFile, map] of ctx!.documents.getMapsByVirtualFileName(vueFile.mainScriptName)) {
		const tsStart = map.map.toGeneratedOffset(prop.exp.loc.start.offset + templateStartOffset, false)?.[0];
		const tsEnd = map.map.toGeneratedOffset(prop.exp.loc.end.offset + templateStartOffset, true)?.[0];
		const ast = ctx?.typescript?.languageService.getProgram()?.getSourceFile(tsFile.fileName);
		if (tsStart !== undefined && tsEnd !== undefined && ast) {
			let node: ts.Node | undefined;
			ast.forEachChild(function visit(_node) {
				if (_node.getStart(ast) === tsStart && _node.getEnd() === tsEnd) {
					node = _node;
				}
				if (!node) {
					_node.forEachChild(visit);
				}
			});
			console.log(tsStart, tsEnd, node?.getText(ast));
		}
	}
}
```

You may also consider another simpler alternative.

```ts
const ast = ts.createSourceFile('/a.ts', prop.exp.loc.source, ts.ScriptTarget.Latest, true, ts.ScriptKind.TS);
```

> check this node is string and end == v-bind directive end

Did it mean `endOffset - templateStartOffset === prop.loc.end.offset`?

zardoy:
> ```ts
> const ast = ts.createSourceFile('/a.ts', prop.exp.loc.source, ts.ScriptTarget.Latest);
> ```

I believe creating new source file on every edit wihtin isn't cheap. However I see there is `prop.exp.loc.__volar_ast` that already contains this file, but I don't see typings for this. Can I reuse it? I would do this simple check then: `prop.exp.loc.__volar_ast.statements.length === 1 && ts.isStringLiteralLike(prop.exp.loc.__volar_ast.statements[0])`. Otherwise I'd try to use first solution (will try in other pr anyway).

johnsoncodehk:
`__volar_ast` is internal and it can be unpredictable, e.g. it sometimes wraps statements in `()`, but yes you can try it. Also I think the cost of doing createSourceFile for a single ExpressionNode is almost negligible.

johnsoncodehk:
Made some refinements, hope not messed up. Thanks!

johnsoncodehk:
This feature will be removed in #4601.

Issue #2472, opened 2023-02-28
JSX compatible slots


KaelWD:
I swear this used to work ages ago.

Vuetify defines slots as `$props.$children` and sets `JSX.ElementChildrenAttribute` to `$children`. Vue already comes with `JSX.ElementAttributesProperty` as `$props` so this works great in JSX:

```tsx
// import { VMenu } from 'vuetify/components'
import type { DefineComponent, VNodeChild } from 'vue'

declare const VMenu: DefineComponent<{
  $children: {
    activator: (props: { isActive: boolean }) => VNodeChild
  }
}>

const el = (
  <VMenu>
    {{ activator: props => String(props.isActive) }}
  </VMenu>
)
```

Volar is only looking at `$slots` or `children` though, so we'd have to define our slot types twice for them to work with it. 

https://github.com/vuejs/language-tools/blob/97b01e6c65d352b41c2bc1155c5ed022e2f30cf1/packages/vue-language-core/src/utils/localTypes.ts#L61-L65

This also means that slots from .vue files only work in other .vue files and can't be used in JSX:

```vue
<script setup lang="ts"></script>

<template>
  <slot name="activator" :isActive="false" />
</template>
```
```tsx
import Slots from './Slots.vue'

const el = (
  <Slots>
    {{ foo: props => props.isActive }} // expected boolean, got implicit any
  </Slots>
)
```

https://github.com/vuejs/core/pull/7083 adds ElementChildrenAttribute to vue core.

Volar could check `T[keyof JSX.ElementAttributesProperty][keyof JSX.ElementChildrenAttribute]` to support any combination of these, and generate its own slots in `$props` somewhere so they can be used in JSX.

KaelWD:
Adding `$props: { $slots }` to our types seems to work for most components, but because volar is using `T['$slots']` instead of `T['$props']['$slots']` it doesn't pick up any slots for components with generic props like VSelect.

Issue #1820, opened 2022-09-05
support slot required checking


xiaoxiangmoe:
Add `experimentalStrictSlotType` option to support  `$slots` required checking.

johnsoncodehk:
Closed due to the high implementation difficulty and lack of user demand.

PR #2547, opened 2023-03-26
feat: Improve Typed Slots Support

johnsoncodehk:
- [x] count anonymous default slot for references codeLens
- [x] #2472
- [x] #1820
- [ ] ~~Typed Slot Children Support~~ (delay)
- [ ] ~~defineSlots Support~~ (delay)

### DIFF
--- a/packages/vue-language-core/src/utils/localTypes.ts
+++ b/packages/vue-language-core/src/utils/localTypes.ts
@@ -66,7 +66,7 @@ export type ExtractComponentSlots<T> =
 	IsAny<T> extends true ? Record<string, any>
 	: T extends { ${slots}?: infer S } ? S
 	: T extends { children?: infer S } ? S
-	: T extends { [K in keyof JSX.ElementChildrenAttribute]?: infer S } ? S
+	: T extends { [K in keyof PickNotAny<JSX.ElementChildrenAttribute, {}>]?: infer S } ? S
 	: Record<string, any>;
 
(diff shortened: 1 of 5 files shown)


packages/vue-language-core/src/utils/localTypes.ts
```diff
@@ -64,8 +64,9 @@ export declare function makeOptional<T>(t: T): { [K in keyof T]?: T[K] };
 // TODO: make it stricter between class component type and functional component type
 export type ExtractComponentSlots<T> =
 	IsAny<T> extends true ? Record<string, any>
-	: T extends { ${slots}?: infer S } ? { [K in keyof S]-?: S[K] extends ((obj: infer O) => any) | undefined ? O : any }
-	: T extends { children?: infer S } ? { [K in keyof S]-?: S[K] extends ((obj: infer O) => any) | undefined ? O : any }
+	: T extends { ${slots}?: infer S } ? S
+	: T extends { children?: infer S } ? S
+	: T extends { [K in keyof PickNotAny<JSX.ElementChildrenAttribute, {}>]?: infer S } ? S
```
KaelWD, line 69: - Shouldn't this be first? DefineComponent always has $slots but the default value is essentially `any`, so ElementChildrenAttribute will never be checked.
- `ElementChildrenAttribute` is a property on the elements' props. In vue that's `$props`, so `JSX.ElementAttributesProperty` also needs to be used. The slots will be on `T[keyof JSX.ElementAttributesProperty][keyof JSX.ElementChildrenAttribute]`

decademoon:
> This doesn't seem to solve my other problem either, slots from .vue files are still not accessible from tsx:

Was this specific issue ever fixed? I'm still having the same problem.

PR #2555, opened 2023-03-29
fix(vue-tsc): fix typescript 5 support

blake-newman:
Typescript 5 will attempt to use the host `resolveModuleNameLiterals` if present, which prevents vue-tsc looking up Vue files.

Set this as undefined so it defaults back to existing logic.

---

I'm unsure on the api of `resolveModuleNameLiterals` so didn't try to implement a solution for this currently; thought it best to fallback to existing behavior and then discover if we get any benefits from using this API.


blake-newman:
This may give some insight on using the `resolveModuleNameLiterals` API.

https://github.com/sveltejs/language-tools/pull/1799

johnsoncodehk:
Thank you for investigating it!

PR #2558, opened 2023-03-29
fix: resolving of paths in vueCompilerOptions

rchl:
Have just tried https://github.com/vuejs/language-tools/pull/2267 and it totally didn't work because the path resolving always returned undefined.


johnsoncodehk:
Thanks!

PR #2578, opened 2023-04-03
fix(vue-language-core): type error with typeof ctx and props

blake-newman:
Latest local types causes error

Screenshot 2023-04-03 at 12 47 21

Using generics seems to fix the issue but unfamilar with the code and use case to know if using generic is viable.


johnsoncodehk:
It's because we removed `@ts-nocheck` [here](https://github.com/vuejs/language-tools/pull/2568/files#diff-1aabbc89bb6f6e01073f1cf4dc7c6fa718b2735ba164e2192816122baad680a5L13), we should add it back.

johnsoncodehk:
Closed because resolved, and vue 2 test is adding by #3351.

PR #2580, opened 2023-04-03
fix(tsc): patch build info roots

blake-newman:
Patch build info roots so that generated types are not considered when checking against project roots, also check for root existance for vue files by removing the virtual file extension.


johnsoncodehk:
Thanks for investigating and testing it!

PR #2597, opened 2023-04-08
chore: Lerna-Lite 2.0 publish & version are becoming optional commands

ghiscoding:
- in the next Lerna-Lite v2.0.0, I made `version` and `publish` as optional commands since it was requested by some users. So Lerna-Lite is now entirely modular and the user will need to manually install the command(s) he want to use (plus the CLI)
- I'm submitting this PR since your Lerna-Lite dependencies are set to `latest` and that will break your workflow, also note `version` is a strict dependency of `publish` which is why we can skip `version` install (we could also include it, it won't make a difference)
- see Lerna-Lite [v2.0.0-alpha.2](https://github.com/lerna-lite/lerna-lite/releases/tag/v2.0.0-alpha.2) release, I'm also expecting an official v2.0 release in a week or two

### DIFF
--- a/package.json
+++ b/package.json
@@ -29,6 +29,7 @@
 	},
 	"optionalDependencies": {
 		"@lerna-lite/cli": "latest",
+		"@lerna-lite/publish": "latest",
 		"opencc": "latest"
 	}
 }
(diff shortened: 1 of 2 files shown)


johnsoncodehk:
LGTM, thanks!

PR #2633, opened 2023-04-20
Update vue-language-configuration.json

webdevnerdstuff:
Added missing `lineComment`.


johnsoncodehk:
See https://github.com/aaron-bond/better-comments/issues/388#issuecomment-1515629054

PR #2665, opened 2023-04-21
fix(vue-typescript): add missing peer dependency `typescript`

merceyz:
`@volar/vue-typescript` depends on `@volar/typescript` which has a peer dependency on `typescript` that it isn't providing.
This PR fixes that by declaring `typescript` as a peer dependency of `@volar/vue-typescript`


johnsoncodehk:
Thanks!

PR #3099, opened 2023-04-30
feat: expose custom blocks attrs

HunYan-io:
Adds `attrs` property to `Sfc.customBlocks` to enable accessing custom block attrs through the Plugins API


johnsoncodehk:
Thanks!

roydukkey:
@HunYan-io Do you have an example of how this is used? Thank you!

Issue #2758, opened 2023-04-28
Using defineSlots with dynamic slot names causes error TS2349: This expression is not callable.


mtorromeo:
I have a generic component, where the slots are defined by the keys of the object bound to the generic parameter.

This is a simplified version of such a component:

MyComp.vue
```vue
<template>
  <div>
    <slot />
    <template v-for="(value, key) of itemData">
      <slot
        v-if="typeof key === 'string'"
        :name="`cell:${key}`"
        :value="value"
      />
    </template>
  </div>
</template>

<script lang="ts" setup generic="T extends DataType">
export type DataType = Record<string, string>;

defineSlots<
  {
    [K in keyof T as K extends string ? `cell:${K}` : never]: { value: T[K] };
  } & {
    default: Record<string, any>;
  }
>();

defineProps<{
  itemData: T;
}>();
</script>
```

And it can be used like this:

App.vue
```vue
<template>
  <MyComp :item-data="{ a: 'A', b: 'B' }">
    <template #cell:a="{ value }">
      <h1>{{ value }}</h1>
    </template>

    <template #cell:b="{ value }">
      <h2>{{ value }}</h2>
    </template>

    test
  </MyComp>
</template>

<script lang="ts">
import { defineComponent } from 'vue';
import MyComp from './components/MyComp.vue';

export default defineComponent({
  components: {
    MyComp: MyComp<{ a: string; b: string; }>,
  },
});
</script>
```

Up to version 1.2 the typechecking in App.vue was working fine, the slots names were recognized and so was the type of the scope variables.

Since the vue-tsc update that validates the slots in the component that defines them, I am now getting the following errors:

```
src/components/MyComp.vue:3:5 - error TS2349: This expression is not callable.
  Not all constituents of type 'Record<string, any> | ((props: ({ [K in keyof T as K extends string ? `cell:${K}` : never]: { value: T[K]; }; } & { default: Record<string, any>; })["default"]) => any)' are callable.
    Type 'Record<string, any>' has no call signatures.

3     <slot />
      ~~~~~~~~

src/components/MyComp.vue:5:7 - error TS2349: This expression is not callable.
  Type '{}' has no call signatures.

  5       <slot
          ~~~~~
  6         v-if="typeof key === 'string'"
    ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
... 
  8         :value="value"
    ~~~~~~~~~~~~~~~~~~~~~~
  9       />
[...]

PR #3116, opened 2023-05-03
Drop support for Vue 3.3 slots short definition

johnsoncodehk:
Issue: #2758

In order to support the slots short definition, we must have a `ToTemplateSlots` type conversion helper to convert the short definition into a legal slots type, but this makes it impossible to handle the generic type.

This is a Volar limitation, we temporarily removed `ToTemplateSlots` until we find a solution, you must fully define the slots type for now:

```diff
<script setup lang="ts">
defineSlots<{
-  foo: { id: string }
+  foo?: (props: { id: string }) => any
}>();
</script>
```

TS playground:

https://www.typescriptlang.org/play?#code/MYGwhgzhAEByD2AXA6vATgawDwBUB80A3gFDRnTDwB2EiaArsIugBQCURp53iAngA4BTaDng5BAW37hEggMogkEXAQC8RaAG0A0tACWVaBkG94AMxEBdAPwAuONVj0Q4AEYhBuHZYKCAHrJUACYwLAB0EWBoAOYQ9mBUvJqWHKoECbzQ1iLe0PYs-Gjw-HE52inQadAZ0AC+ANxc3GQeiNAQiogQAIT2ouJSMvKdyiTN41q6BkYm5iLVMLr+gSHtdAbRWdAABsCCLrYAJITatdt50FSCAG6CaJb2hNDXYCD0gn25DU3NtdAAZJwJs0goIzGBnIh7AAlQSUNBBLC0NAbAA01USeEawNqWJ+3EoNDaHSQlXaIzCoPBkMaAEhaXwhNAAJIQABqrz0QQUpPUjMEcxJbWWgmCoQiYSisXiiVS6USWzo7wu4JAEEE2ImAHotdwAHrWH61YjG4igSAwVCYFRA8iE5GMZhodi28atclKXqu4GTfSGYymCw4BbQJYBUWrZEbLa7fYgI4nM4XK63e75AD6j2er3en3KdTlGN4mvGf0BYx9ZCpEJAUOgLEz0Fh8MRUao0XRGTwhYyJd+fYJ1FoHra6iFEEpYJriDpDIEwlZHJAXJ5o+g-MFnWgIrF9YlUtKGR7CuySuE9lV6oH5B1+sN3GNxqAA


chinbor:
how about this ?

chinbor:
It's effective in solving “This expression is not callable. Type XXX has no call signatures”
image
but when I use the scopedSlot it reports another error "Property XXX does not exist on type YYY" 
image

jd-solanki:
@chinbor Try removing vue-macros plugins from `vueCompilerOptions` in `tsconfig.json` and reload the vscode, I hope this will resolve your issue. I was also facing the same. Let me know if this resolves your issue.

chinbor:
> @chinbor Try removing vue-macros plugins from `vueCompilerOptions` in `tsconfig.json` and reload the vscode, I hope this will resolve your issue. I was also facing the same. Let me know if this resolves your issue.

It doesn't work

PR #3145, opened 2023-05-08
fix(language-core): add missing peer dependency `typescript`

merceyz:
`@vue/language-core` tries to require `typescript` but doesn't declare it as a dependency.
This PR fixes that by declaring `typescript` as a peer dependency of `@vue/language-core`.

Ref https://github.com/yarnpkg/berry/actions/runs/4919080804/jobs/8786259067#step:5:63


johnsoncodehk:
Thanks!

PR #3156, opened 2023-05-11
fix(language-service): remove js extension from vue imports

### LINKED ISSUE #3150: Volar adds ".js" extension when autoimporting components
Version: v1.7.0

When `typescript.preferences.importModuleSpecifierEnding` and `javascript.preferences.importModuleSpecifierEnding` is enabled (`js`), VSCode auto imports Vue components with `.vue.js` extension instead of just `.vue` (when typing component name in script or template blocks).

Reproduction: https://github.com/sheremet-va/volar-bug-extension

kalvenschraut:
My best attempt to fix #3150.

Mainly it seems to be fixing it after the fact, possible something is adding .vue.js extension in the first place.
Not sure where to look to find that though.

Also I was looking at trying to add tests, but doesn't seem to be a straight forward way to add another completion test, but with different vscode settings?

### DIFF
--- a/packages/vue-language-service/src/languageService.ts
+++ b/packages/vue-language-service/src/languageService.ts
@@ -82,7 +82,6 @@ function resolvePlugins(
 							&& (!item.labelDetails?.description || item.labelDetails.description.indexOf('__VLS_') === -1)
 						) {
 							if (item.labelDetails?.description?.endsWith('.vue.js')) {
-								console.log(item);
 								// remove .js from .vue files
 								item.labelDetails.description = item.labelDetails.description.slice(0, -3);
 							}
--- a/packages/vue-language-service/src/languageService.ts
+++ b/packages/vue-language-service/src/languageService.ts
@@ -75,11 +75,21 @@ function resolvePlugins(
 				const result = await base.provideCompletionItems?.(document, position, context, item);
 				if (result) {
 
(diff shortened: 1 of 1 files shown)


johnsoncodehk:
Hi @kalvenschraut, I see you have multiple PRs, can you rebase these to the v1 branch? We will have a 1.6.5 patch.

kalvenschraut:
All PRs should be rebased to the v1 branch now

johnsoncodehk:
Thank you for your PR, this problem is caused by the existence of `import('__VLS_types.js')` in the virtual code, I avoid it in https://github.com/vuejs/language-tools/commit/8aa57a3d5f66aee3a4e2b4de897aa1f1ba9501fe this situation.

Since it is a breakthrough change, it will be between 1.7 and 2.0. If you try it, please let me know if the problem is solved.

kalvenschraut:
Finally got around to testing this. Now vue files are imported with `vue.ts` instead of `vue.js`. Typescript files correctly are being imported with the `.ts` extension now though.

johnsoncodehk:
Thanks for confirming it, .vue.ts is technically expected, I'm not sure if it needs to be handled because there are no user requests about this.

kalvenschraut:
If possible I would prefer for it to just import as .vue. I can open up a separate issue if needed

PR #3158, opened 2023-05-11
perf(language-service): avoid looping through all items twice

kalvenschraut:
Noticed this while working on my other PRs, I saw two forEachs over the item to avoid this.


johnsoncodehk:
Thanks, I messed up this code so I'm still leaning towards not merging into a loop just to make the logic a bit more readable, but I'm keeping your cached `getNameCasing()` improvement.

Issue #2662, opened 2023-04-21
typescript 5.0 and nodenext module resolutions results in vue files not found


kalvenschraut:
Original discussion about initial support [here](https://github.com/vuejs/language-tools/issues/1394)
I confirmed was still an issue as of vue-tsc 1.4. 

Reproduction [here](https://github.com/RtVision/vue-tsc-nodenext-repro)

Also wanted to note that this error does not appear in my editor, only when running vue-tsc.

johnsoncodehk:
I discussed it with team members and the conclusion is that you should use `"moduleResolution": "bundler"`. Is there any reason you can't use it?

kalvenschraut:
Mainly for consistency across my monorepo codebase.

It  was working previously and no error appears in my editor. Feel like there is just something missing in vue-tsc that was not ported over from the language-server.

Mainly using nodenext as I was trying to get all my ts code compiled to ESM (have code not using vite atm, primarly API code). 

Looks like bundle resolution doesn't enforce extensions  to be used everywhere, which I preferred with the nodenext resolution.  Guess this probably is a linting rule somewhere if I look for one though.

It appears module resolution bundle was first introduced in 5.0.

kalvenschraut:
Tried setting up the eslint rule to cover this case and had mixed results. 

Is there a specific reason that Node16/Next can't be supported also?

If it is just a time factor, I can attempt to look into this if there is any specific area you would recommend looking at.

kalvenschraut:
I think i more understand the problem now.

Typescript changed to ESM which makes the patching of (this)[https://github.com/vuejs/language-tools/blob/master/packages/vue-language-core/src/languageModule.ts#L97]. Not work

Typescript seems to recommend to do module resolution bundler here, or I just live with .vue.js files.

Actually if https://github.com/vuejs/language-tools/issues/3150 is fixed then this would resolve some of the issues I ran into in my previous comment.

kalvenschraut:
@johnsoncodehk  I have a somewhat working version for this I can submit a PR for, but more so wondering if volar wants to take the stance of supporting NodeNext or push everyone to bundler resolution.

Lot of inspiration came from svelte language tools, https://github.com/sveltejs/language-tools/blob/9dd59e03d4e5e0e83b5bbfd2aa8331e6bf5cebaa/packages/typescript-plugin/src/module-loader.ts#L199

so1ve:
@kalvenschraut Can we close this issue? :)

[...]

PR #3159, opened 2023-05-11
fix(language-core): try handle node next module resolution

kalvenschraut:
These are the changes I mentioned in #2662 in my last comment.

If direction is to not support node next then this PR can be closed and disregarded.


packages/vue-language-core/src/languageModule.ts
```diff
@@ -2,20 +2,49 @@ import type { Language } from '@volar/language-core';
...
+	return useCaseSensitiveFileNames ? identity : toFileNameLowerCase;
+}
+let moduleCache: ts.ModuleResolutionCache | null = null;
+function getModuleResolutionCache(
+	ts: typeof import("typescript/lib/tsserverlibrary"),
+	options: ts.CompilerOptions
+) {
+	if (moduleCache === null) {
+		moduleCache = ts.createModuleResolutionCache(
+			process.cwd(),
```
yoyo930021, line 34: The path isn't correct. 
It need to from `host.getCurrentDirectory()`

packages/vue-language-core/src/languageModule.ts
```diff
@@ -69,6 +98,29 @@ export function createLanguage(
...
 				},
+				resolveModuleNameLiterals(
+					moduleLiterals,
+					containingFile,
+					redirectedReference,
+					options,
+					sourceFile
+				) {
+					return moduleLiterals.map((_, index) => {
+						let moduleName = moduleLiterals[index].text;
```
yoyo930021, line 109: Why not use first argument?

yoyo930021 (changes requested):
I will test in my project for this PR.

packages/vue-language-core/src/languageModule.ts
kalvenschraut: No reason that it can't be now. These changes started out from me looking at https://github.com/sveltejs/language-tools/blob/3f247aed28488f479399da2d852b3d258875bfbc/packages/typescript-plugin/src/module-loader.ts#L231.

packages/vue-language-core/src/languageModule.ts
kalvenschraut: Yea I wasn't sure what path it was looking for exactly. compiler-sfc uses process.cwd so went with that.

packages/vue-language-core/src/languageModule.ts
```diff
@@ -1,21 +1,51 @@
...
+}
+
 export function createLanguage(
 	compilerOptions: ts.CompilerOptions = {},
 	vueCompilerOptions = resolveVueCompilerOptions({}),
 	ts: typeof import('typescript/lib/tsserverlibrary') = require('typescript'),
 	codegenStack: boolean = false,
 ): Language {
-
-	patchResolveModuleNames(ts, vueCompilerOptions);
```
johnsoncodehk, line 17: Can we keep it to maintain support for TS 4?

packages/vue-language-core/src/languageModule.ts
```diff
@@ -69,6 +99,29 @@ export function createLanguage(
...
+				resolveModuleNameLiterals(
+					moduleLiterals,
+					containingFile,
+					redirectedReference,
+					options,
+					sourceFile
+				) {
+					return moduleLiterals.map((moduleLiteral) => {
+						let moduleName = moduleLiteral.text;
+						if (sourceFile.impliedNodeFormat === ts.ModuleKind.ESNext && moduleName.endsWith('.vue')) {
```
johnsoncodehk, line 111:
```suggestion
						if (sourceFile.impliedNodeFormat === ts.ModuleKind.ESNext && vueCompilerOptions.extensions.some(ext => moduleName.endsWith(ext))) {
```

johnsoncodehk:
@yoyo930021 changes look neat, have you tested it in a real project?

johnsoncodehk:
You can revert master merge and I'll handle it, it can't just resolve in this PR.

kalvenschraut:
okay, undid my master merge and yea quickly realize there was some typing problems. Looks like the problem still exists on the action checks? Seems fine locally so guessing the volar package change and is probably what you were referring to what you needed to resolve?

Otherwise should be all cleaned up now with request changes.

johnsoncodehk:
Hey @kalvenschraut, I can't push commit to RTVision:node-next-resolution branch, can you check permissions?

kalvenschraut:
I know on newer pull requests I make there is a checkbox to allow the maintainers to do this, not sure how I can retroactively give permission. Best thing I could find is inviting you as a collaborator on the fork.

johnsoncodehk:
Thanks!

PR #3215, opened 2023-05-22
feat: Support for `@vue-ignore`, `@vue-skip`, `@vue-expected-error` directive comments

### LINKED ISSUE #2366: Allow using `@ts-expect-error` in Vue templates
Hi all, and thanks again @johnsoncodehk for the amazing Volar.

I have a case on my Nuxt module (https://github.com/victorgarciaesgi/nuxt-typed-router), where I need to test types output of components.
I use `vue-tsc` to scan the fixtures to make sure things break where it should, and things are valid where it should.
It works well for ts files because I can use `@ts-expect-error`.
But on my case I want to test that the overrided types of `<NuxtLink/>` are working well.

The workaround I found for now is to use a .tsx and to test on a `render` function, but that miss the point the types are correctly injected into the global Vue scope.

What I can do in tsx:

```tsx
import {NuxtLink} from '@typed-router';

function render() {
  return (
        {/* @ts-expect-error */}
        <NuxtLink to={{ name: 'user-id', params: { foo: 'bar' } }}></NuxtLink>;
   )
}
```

What I want to do in vue templates:

image

johnsoncodehk:
Close #2366

Support 3 directive comment syntax for template.

`@vue-ignore`: Similar to `@ts-ignore`, ignore errors of the next template node. (Not just the next line)
`@vue-expected-error`: Similar to `@ts-expected-error`, report if the next template node has no errors.
`@vue-skip`: The next template node and its child nodes will skip generating virtual code.

## Usage

```html
<template>
  <!-- @vue-ignore -->
  <HelloWorld :msg="wrongType" />

  <!-- @vue-expected-error -->
  <HelloWorld :msg="wrongType" />

  <!-- @vue-skip -->
  <SomethingYouDontWantVolarToRecognizeIt>
    ...
  </SomethingYouDontWantVolarToRecognizeIt>
</template>
```

### DIFF
--- a/package.json
+++ b/package.json
@@ -22,7 +22,7 @@
 	},
 	"devDependencies": {
 		"@types/node": "latest",
-		"@volar/language-service": "1.6.8",
+		"@volar/language-service": "1.6.9",
 		"typescript": "latest",
 		"vite": "latest",
 		"vitest": "latest"
--- a/packages/vscode-vue/package.json
+++ b/packages/vscode-vue/package.json
@@ -732,7 +732,7 @@
 	"devDependencies": {
 		"@types/semver": "^7.3.13",
 		"@types/vscode": "1.67.0",
-		"@volar/vscode": "1.6.8",
+		"@volar/vscode": "1.6.9",
 		"@vue/language-server": "1.7.6",
 		"esbuild": "0.15.18",
 		"esbuild-plugin-copy": "latest",
--- a/packages/vue-component-meta/package.json
+++ b/packages/vue-component-meta/package.json
@@ -13,7 +13,7 @@
 		"directory": "packages/vue-component-meta"
 	},
 	"dependencies": {
-		"@volar/language-core": "1.6.8",
+		"@volar/language-core": "1.6.9",
 		"@vue/language-core": "1.7.6",
 		"typesafe-path": "^0.2.2",
 		"vue-component-type-helpers": "1.7.6"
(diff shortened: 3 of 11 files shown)


Anoesj:
TypeScript uses `@ts-expect-error`, not `@ts-expected-error`. Other than that, great stuff! 😄

nekocom3745 (commented):
packages/vue-language-core/src/generators/template.ts

nekocom3745 (commented):
packages/vue-language-core/src/generators/template.ts

johnsoncodehk:
@Anoesj Unfortunately bug 🤦, will fix it in the next release, thanks for the report.

johnsoncodehk:
> TypeScript uses `@ts-expect-error`, not `@ts-expected-error`.

Fixed by https://github.com/vuejs/language-tools/commit/5696eb2a8aea8ccf2af9eed9e740d0c1a3361458

saibotk:
What about allowing comments to follow the tags?

eg:
```vue
<!-- @vue-expect-error We could type this with some magic, but it would probably be complex to do -->

equivalent in ts:
// @ts-expect-error We could type this with some magic, but it would probably be complex to do
```

wenfangdu:
> What about allowing comments to follow the tags?
> 
> eg:
> 
> ```
> 
> 
> equivalent in ts:
> // @ts-expect-error We could type this with some magic, but it would probably be complex to do
> ```

That's already supported.

wenfangdu:
> What about allowing comments to follow the tags?
> 
> eg:
> 
> ```
> 
> 
> equivalent in ts:
> // @ts-expect-error We could type this with some magic, but it would probably be complex to do
> ```

@saibotk That's already supported in 1.8.0.

Issue #3224, opened 2023-05-25
failed to load tsconfig json schema


tjx666:
<img width="1440" alt="image" src="https://github.com/vuejs/language-tools/assets/41773861/4c75652a-650e-4bb3-8df8-7e728e8f0167">

```txt
Problems loading reference 'file:///Users/yutengjing/.vscode/extensions/vue.volar-1.7.8/dist/schemas/vue-tsconfig.deprecated.schema.json': Unable to load schema from '/Users/yutengjing/.vscode/extensions/vue.volar-1.7.8/dist/schemas/vue-tsconfig.deprecated.schema.json': ENOENT: no such file or directory, open '/Users/yutengjing/.vscode/extensions/vue.volar-1.7.8/dist/schemas/vue-tsconfig.deprecated.schema.json'.(768)
```

<img width="920" alt="image" src="https://github.com/vuejs/language-tools/assets/41773861/c2892727-61a3-472c-a0cf-3d891d9e88ad">

```
Version: 1.78.2
Commit: b3e4e68a0bc097f0ae7907b217c1119af9e03435
Date: 2023-05-10T15:00:40.428Z
Electron: 22.5.2
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Darwin arm64 22.4.0
Sandboxed: No
Volar: v1.7.8 Pre-Release
```

PR #3228, opened 2023-05-26
fix: failed to load tsconfig json schema #3224

tjx666:


tjx666:
because lack of dep @volar/language-core:

https://github.com/vuejs/language-tools/blob/master/packages/vscode-vue/scripts/build.js#L37

PR #3259, opened 2023-06-08
perf: Avoid doing type gymnastics with `JSX.IntrinsicElements`

### LINKED ISSUE #2533: vue-tsc very slow since typescript version 4.9
Good day everyone,

ever since I updated my typescript version from 4.8 to 4.9, my linting times of `vue-tsc` take significantly longer (about 5 times longer). Updating to typescript 5.0 didn't help either. Before it took a little less than 3 minutes but now they take 15 minutes.

I wanted to know if this is a known problem or if anyone else experiences the same issue?

Sadly I cannot share my company's codebase to give a reproducible example, but if there are other ways to track down the underlying issue I'll gladly offer my help.

### LINKED ISSUE #3130: 1.6.1+ is super slow on saving
I disabled all extensions except Volar. In 1.6.0 it's fine but could take ~20 seconds to save files in 1.6.1+ versions.

### LINKED ISSUE #3148: running vue-tsc with 1.7 takes 5min instead of 1.5min
I recently updated from vue-tsc v1.2 to the prereleased 1.7 and my ci went up from only 1.5min to a whopping 5min.

Since since is surely not by design, I wonder what changed that makes it so slow and if that is expected.

I will update this ticket with a more narrow version constraint to see with which version it got so slow

// EDIT:
The time increase happens from version 1.5 to version 1.5.1.

These are the changes made between the versions:
https://github.com/vuejs/language-tools/compare/v1.5.0...v1.5.1

Exact commit that caused it:
https://github.com/vuejs/language-tools/commit/6d8829980981b04f6ce69462e47c6be5825088b3

### LINKED ISSUE #3226: Volar 1.6 and 1.7 very slow for large project
I have a large Vue 3 + Typescript project. Ever since updating Volar to 1.6.x, the performance of Volar took a nosedive. If I make any edit  in `.vue` files, it takes up to 20 seconds for linting and syntax highlighting to be updated. There is no performance issue when editing `.ts` files in the same project, so I don't think the typescript server is the limiting factor. If I downgrade Volar to 1.2.x or 1.4.x, the performance is fast again.

It would be difficult for me to make this into a reproduction repo, are there other steps I can take to help debug this issue? Which features or changes in 1.6 could be responsible for such a decrease in performance?

johnsoncodehk:
close #3226, close #3148, close #3130, close #2533

First I wasn't able to actually reproduce the problem, so please let me know if it doesn't solve the above closed issue. Thanks to @Fuzzyma for helping to investigate, I think the problem is most likely happening in #2685.

In this PR we no longer do type gymnastics with `JSX.IntrinsicElements`, so TS type evaluation may be faster.

But it also makes the template unable to separate `JSX.IntrinsicElements` and components types, which affects DX in some cases, but I think it is a reasonable trade-off, and it is too expensive to trade performance for it.

### DIFF
--- a/packages/vue-language-core/src/generators/template.ts
+++ b/packages/vue-language-core/src/generators/template.ts
@@ -179,7 +179,7 @@ export function generate(
 
 		const data: Record<string, string> = {};
 
-		codes.push(`let __VLS_templateComponents!: {}\n`);
+		codes.push(`let __VLS_templateComponents!: __VLS_IntrinsicElements\n`);
 
 		for (const tagName in tagNames) {
 
--- a/packages/vue-language-core/src/utils/directorySharedTypes.ts
+++ b/packages/vue-language-core/src/utils/directorySharedTypes.ts
@@ -62,12 +62,10 @@ declare function __VLS_makeOptional<T>(t: T): { [K in keyof T]?: T[K] };
 
 type __VLS_SelfComponent<N, C> = string extends N ? {} : N extends string ? { [P in N]: C } : {};
 type __VLS_WithComponent<N0 extends string, Components, N1 extends string, N2 extends string, N3 extends string> =
-	__VLS_IsAny<__VLS_IntrinsicElements[N0]> extends true ? (
-		N1 extends keyof Components ? N1 extends N0 ? Pick<Components, N0> : { [K in N0]: Components[N1] } :
-		N2 extends keyof Components ? N2 extends N0 ? Pick<Components, N0> : { [K in N0]: Components[N2] } :
-		N3 extends keyof Components ? N3 extends N0 ? Pick<Components, N0> : { [K in N0]: Components[N3] } :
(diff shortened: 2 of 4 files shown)


rchl:
> But it also makes the template unable to separate `JSX.IntrinsicElements` and components types

Can you expand a bit on that?
Are `JSX.IntrinsicElements` all the native DOM elements?
Does this PR affect the behavior of handling unknown attributes/props on components? For example I think it would be fine to allow any custom attribute on a native DOM element but I would expect an error to be raised when setting unknown prop/attribute on a component (with some exceptions like `class` and other).

johnsoncodehk:
For example, for the following code:

```html
<template>
  <img alt="Vue logo" src="./assets/logo.png" />
</template>

<script lang="ts" setup>
import img from './components/HelloWorld.vue';
</script>
```

`` type will be `JSX.IntrinsicElements['img'] & typeof import('./components/HelloWorld.vue')['default']`.

So you just need to avoid component name same with `JSX.IntrinsicElements` properties.

PR #3274, opened 2023-06-13
fix: component type missing for TS 5.1.3

### LINKED ISSUE #3252: Emit variable is set to `any`, breaks build but works fine in dev mode
### Vue version

3.3.4

https://github.com/rodrigocfd/vue-bug-emits

### Steps to reproduce

Run `npm run dev` and everything works fine.

Run `npm run build` and it won't compile.

### What is expected?

A successful build with `npm run build`.

With `npm run build`, the following error shows up:

```
src/App.vue:10:22 - error TS7006: Parameter 'n' implicitly has an 'any' type.

10  <Child @changeName="n => name = n" />
                        ~


Found 1 error in src/App.vue:10
```

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i7-8700 CPU @ 3.20GHz
    Memory: 25.86 GB / 31.95 GB
  Binaries:
    Node: 18.14.1 - C:\Program Files\Nodejs\node.EXE
    npm: 9.6.7 - C:\Program Files\Nodejs\npm.CMD
  Browsers:
    Chrome: 114.0.5735.91
    Edge: Spartan (44.19041.1266.0)
    Internet Explorer: 11.0.19041.1566
  npmPackages:
    vue: ^3.3.4 => 3.3.4
```

### Any additional comments?

I tested the original dependencies installed with `npm create vite@latest`, and then updating all dependencies. The behavior is the same.

Curiously, despite running flawlessly in dev mode, when hovering the variable in VSCode, Volar reports the emit parameter as `any`, which is exactly the build error:

volar-any

### LINKED ISSUE #3265: volar 功能失效
vscode版本：
版本: 1.79.0
提交: b380da4ef1ee00e224a15c1d4d9793e27c2b6302
日期: 2023-06-07T14:29:00.206Z
Electron: 22.5.5
Chromium: 108.0.5359.215
Node.js: 16.17.1
V8: 10.8.168.25-electron.0
OS: Darwin x64 20.6.0

volar 发布版无效，切换至预览版也是无效

都是any类型，无提示。检测不到任何错误，只有语法高亮的功能。

image
typescript版本为：5.1.3

### LINKED ISSUE #3268: Component type is missing
I have just created a project with vite `npm create vite@latest`:

The type of the component is always `any`:

image

so1ve:
close #3268, close #3265, close #3252
Ref: microsoft/TypeScript#54630

### DIFF
--- a/packages/vue-language-core/src/utils/directorySharedTypes.ts
+++ b/packages/vue-language-core/src/utils/directorySharedTypes.ts
@@ -24,6 +24,8 @@ type __VLS_VueRuntimeCoreGlobalComponents = import('@vue/runtime-core').GlobalCo
 // @ts-ignore
 type __VLS_VueRuntimeDomGlobalComponents = import('@vue/runtime-dom').GlobalComponents;
 
+// Hack: `Type & Type` is a hack to make unresolved imports' type work.
+// Ref: #3274
 type __VLS_GlobalComponents =
 	& __VLS_PickNotAny<__VLS_VueGlobalComponents & __VLS_VueGlobalComponents, {}>
 	& __VLS_PickNotAny<__VLS_VueRuntimeCoreGlobalComponents & __VLS_VueRuntimeCoreGlobalComponents, {}>
--- a/packages/vue-language-core/src/utils/directorySharedTypes.ts
+++ b/packages/vue-language-core/src/utils/directorySharedTypes.ts
@@ -17,6 +17,7 @@ type __VLS_Prettify<T> = {
 	[K in keyof T]: T[K];
 } & {};
 
+// TODO: revert this once microsoft/TypeScript#54630 is fixed
 // @ts-ignore
 type __VLS_VueGlobalComponents = import('vue').GlobalComponents;
 // @ts-ignore
--- a/packages/vue-language-core/src/utils/directorySharedTypes.ts
+++ b/packages/vue-language-core/src/utils/directorySharedTypes.ts
@@ -18,14 +18,10 @@ type __VLS_Prettify<T> = {
 } & {};
 
 type __VLS_GlobalComponents =
-	// @ts-ignore
 	__VLS_PickNotAny<import('vue').GlobalComponents & {}, {}>
-	// @ts-ignore
 	& __VLS_PickNotAny<import('@vue/runtime-core').GlobalComponents & {}, {}>
-	// @ts-ignore
 	& __VLS_PickNotAny<import('@vue/runtime-dom').GlobalComponents & {}, {}>
 	& Pick<typeof import('${vueCompilerOptions.lib}'),
-		// @ts-ignore
 		'Transition'
 		| 'TransitionGroup'
 		| 'KeepAlive'
--- a/packages/vue-language-core/src/utils/directorySharedTypes.ts
+++ b/packages/vue-language-core/src/utils/directorySharedTypes.ts
@@ -25,6 +25,18 @@ type __VLS_VueRuntimeCoreGlobalComponents = import('@vue/runtime-core').GlobalCo
 type __VLS_VueRuntimeDomGlobalComponents = import('@vue/runtime-dom').GlobalComponents;
 
 // Hack: \`Type & Type\` is a hack to make unresolved imports' type to be "real any" in TS 5.1.3.
+// Since typescript 5.1.3, the behavior of unresolved imports has changed.
+// For example, the following code:
+// ```ts
+// type A = import('foo').NotExist extends any ? true : false;
+// type B = A extends any ? true : false;
+// ```
+// In typescript 5.0.4, A and B are both `true`.
+// However, in 5.1.3, A becomes `any` and B becomes `any` too!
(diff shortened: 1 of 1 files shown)


so1ve:
Not the right way to fix...?

so1ve:
Hi @rchl! I have pushed a new commit. Could you please check it out if it works?

rchl:
I don't know if it's the right approach. As I said, I don't think it's OK to import packages that might not exist but maybe TS is fine with it...

so1ve:
> I don't know if it's the right approach. As I said, I don't think it's OK to import packages that might not exist but maybe TS is fine with it...

Yes, the original logic is to ignore errors through `@ts-ignore`. In this case the imported type will be `any`, but I don't know why in 5.1.3 the type is still `any` but the behavior changed.

rchl:
Oh, I see now that the original code also imported those so I guess this is not any worse.

packages/vue-language-core/src/utils/directorySharedTypes.ts
```diff
@@ -17,13 +17,17 @@ type __VLS_Prettify<T> = {
...
+type __VLS_VueRuntimeDomGlobalComponents = import('@vue/runtime-dom').GlobalComponents;
+
 type __VLS_GlobalComponents =
-	// @ts-ignore
-	__VLS_PickNotAny<import('vue').GlobalComponents, {}>
-	// @ts-ignore
-	& __VLS_PickNotAny<import('@vue/runtime-core').GlobalComponents, {}>
-	// @ts-ignore
-	& __VLS_PickNotAny<import('@vue/runtime-dom').GlobalComponents, {}>
+	& __VLS_PickNotAny<__VLS_VueGlobalComponents & __VLS_VueGlobalComponents, {}>
```
rchl, line 28: Maybe include a comment explaining this "hack"?
Personally I don't understand what's the purpose of doing `&` against same type.

so1ve: Done!

rchl: You wrote that it's a hack but I still don't understand what it actually does. What difference does it make to types?

so1ve: Well I added a link in the pr description

packages/vue-language-core/src/utils/directorySharedTypes.ts
```diff
@@ -17,13 +17,31 @@ type __VLS_Prettify<T> = {
...
+// @ts-ignore
+type __VLS_VueRuntimeDomGlobalComponents = import('@vue/runtime-dom').GlobalComponents;
+
+// Hack: \`Type & Type\` is a hack to make unresolved imports' type to be "real any" in TS 5.1.3.
+// Since typescript 5.1.3, the behavior of unresolved imports has changed.
+// For example, the following code:
+// \`\`\`ts
+// type A = import('foo').NotExist extends any ? true : false;
+// type B = A extends any ? true : false;
+// \`\`\`
```
rchl, line 33: Why are you escaping characters within the comment?

packages/vue-language-core/src/utils/directorySharedTypes.ts
```diff
@@ -17,13 +17,31 @@ type __VLS_Prettify<T> = {
...
+// In typescript 5.0.4, A and B are both \`true\`.
+// However, in 5.1.3, A becomes \`any\` and B becomes \`any\` too!
+// Actually, any type which includes unresolved imports will be turned into \`any\`.
+// Even only a reference, like B.
+// So we need to turn it into a \`real any\` and avoid pollution.
+// This hack does it by intersect itself with itself 
+// Ref: #3274
```
rchl, line 40: Why are you escaping characters within the comment?

rchl, line 40: Can you link to https://github.com/microsoft/TypeScript/issues/54630 since it might be just a bug in TS and won't need the workaround once its fixed.

so1ve: Because I use backsticks to wrap code snippets, but the template literal also uses backsticks

rchl: This is just a plain comment, not parsed as markdown by anything. So I don't see the point of escaping. Just use a plain text.

so1ve: Added a TODO

packages/vue-language-core/src/utils/directorySharedTypes.ts
```diff
@@ -17,6 +17,7 @@ type __VLS_Prettify<T> = {
 	[K in keyof T]: T[K];
 } & {};
 
+// TODO: revert this once microsoft/TypeScript#54630 is fixed
```
rchl: Can't you insert a normal link? :)
It's easier to just click or copy&paste then having to look it up.

eimigueloliveir (approved):

PR #3279, opened 2023-06-14
feat: redo `nativeTags`

### LINKED ISSUE #3148: running vue-tsc with 1.7 takes 5min instead of 1.5min

johnsoncodehk:
close #3148

This PR is a rollback of #2685.

After confirmation, #3259 does not solve #3148. It seems that it is necessary to generate more efficient virtual code through the nativeTags of hard code, instead of entrusting the work of judging IntrinsicElements to TS.

This also fixed the regression introduced by #3259.

cc @ModyQyW

### DIFF
--- a/packages/vue-component-meta/src/index.ts
+++ b/packages/vue-component-meta/src/index.ts
@@ -85,7 +85,7 @@ function createComponentMetaCheckerWorker(
 	};
 
 	return {
-		...baseCreate(_host, parsedCommandLine.vueOptions, checkerOptions, globalComponentName, ts),
+		...baseCreate(_host, vue.resolveVueCompilerOptions(parsedCommandLine.vueOptions), checkerOptions, globalComponentName, ts),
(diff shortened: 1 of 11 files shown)


ModyQyW:
Thanks! I will check it later. Currently, I am stuck in company issues and having bad days 😅

PR #3280, opened 2023-06-14
fix: inline variables are wrongly transformed as ctxVars

### LINKED ISSUE #3258: Can't define variables in inline event handler: "Property does not exist on type"
When defining a variable like this:
```vue
@click="() => {
  const x = 1
  if (x === 1) {
  }
}"
```
it yields two errors:
* `'x' is declared but its value is never read. ts(6133)`
* `Property 'x' does not exist on type 'MyComponent'. ts(2339)`

The code using such variable **works correctly**, and I've been using this syntax since Vue 2 with **Vetur**. After migrating to Vue 3 with **Volar**, this type safety and inference in VS Code fails.

image

so1ve:
closes #3258

### DIFF
--- a/packages/vue-language-core/src/utils/transform.ts
+++ b/packages/vue-language-core/src/utils/transform.ts
@@ -133,6 +133,7 @@ function walkIdentifiers(
 		walkIdentifiers(ts, node.expression, cb, localVars);
 	}
 	else if (ts.isVariableDeclaration(node)) {
+
 		colletVars(ts, node.name, blockVars);
 
 		for (const varName of blockVars)
--- a/packages/vue-language-core/src/utils/transform.ts
+++ b/packages/vue-language-core/src/utils/transform.ts
@@ -20,7 +20,7 @@ export function walkInterpolationFragment(
 
 	const varCb = (id: ts.Identifier, isShorthand: boolean) => {
 		if (
-			localVars[id.text] !== undefined ||
+			!!localVars[id.text] ||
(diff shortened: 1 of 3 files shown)


packages/vue-language-core/src/utils/transform.ts
```diff
@@ -39,7 +39,10 @@ export function walkInterpolationFragment(
 	};
 	ast.forEachChild(node => walkIdentifiers(ts, node, varCb, localVars));
 
-	ctxVars = ctxVars.sort((a, b) => a.offset - b.offset);
+	// Avoid duplicate ctx vars
+	// Fixes https://github.com/vuejs/language-tools/issues/3258
+	const localVarsKeys = Object.keys(localVars);
+	ctxVars = ctxVars.filter(v => !localVarsKeys.includes(v.text)).sort((a, b) => a.offset - b.offset);
```
rchl, line 45: Wouldn't it make sense to avoid pushing to `ctxVars` in the first place above?

so1ve: Makes sense. Update later

so1ve: Done

so1ve:
Oops. Why?

johnsoncodehk:
If `localVars[id.text]` is 0 or undefined, it means that the current code scope cannot access the local variable, so this is a problem that `walkIdentifiers` does not collect variables correctly.

so1ve:
Thanks for your explanation! Now I know how auto unwrapping works :)

PR #3286, opened 2023-06-14
feat(component-meta): expose component type

### LINKED ISSUE #3151: [Feature request]  api that accepts a file ( .vue, .ts, .js, .tsx...) and return if it is a component or no.
**vue-component-meta** is a tool that enables the extraction of metadata, such as props, slots, events, etc., from a file, regardless of whether the file is a component or not. While .vue files are compiled as components, .js and .ts files may not export any component.

In a Storybook application, **vue-component-meta** can be useful for extracting metadata from component files. However, the current implementation only allows filtering of .vue files, assuming they are components. To extract metadata from .ts, .js, .tsx, and .jsx files, a new API is needed to filter only the files that are components (SFC, options API components, functional components, etc.).

Similarly, in other contexts such as a UI library, **vue-component-meta** can be useful for code analytics, such as counting the number of components and their types. The new API that filters only component files would be beneficial in this context as well. Overall, **vue-component-meta** has several potential use cases where the ability to extract metadata from a variety of file types is necessary.

johnsoncodehk:
close #3151

Expose `type` property to meta:

- `0`: Unknown
- `1`: Component is a class type
- `2`: Component is a function type

cc @chakAs3

### DIFF
--- a/packages/vue-component-meta/src/index.ts
+++ b/packages/vue-component-meta/src/index.ts
@@ -194,6 +194,7 @@ import * as Components from '${fileName.substring(0, fileName.length - '.meta.ts
 export default {} as { [K in keyof typeof Components]: ComponentMeta<typeof Components[K]>; };
 
 interface ComponentMeta<T> {
+	type: ComponentType<T>;
 	props: ComponentProps<T>;
 	emit: ComponentEmit<T>;
 	slots: ${vueCompilerOptions.target < 3 ? 'Vue2ComponentSlots' : 'ComponentSlots'}<T>;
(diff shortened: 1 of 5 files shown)


packages/vue-component-meta/src/index.ts
```diff
@@ -245,6 +250,18 @@ ${typeHelpersCode}
...
+
+			const $type = symbolProperties.find(prop => prop.escapedName === 'type');
+
+			if ($type) {
+				const type = typeChecker.getTypeOfSymbolAtLocation($type, symbolNode!);
+				return Number(typeChecker.typeToString(type));
+			}
+
+			return 0;
+		}
```
chakAs3, line 263: great it was easier than I thought 😄  , so if the return is 0  => it is not a valid Vue component?

johnsoncodehk: Yes it's designed that way, I'm just not sure how applicable it is for your use case.

chakAs3: very useful, as I'm  loading folders with components and *.stories.ts  so many are not exposing any vue component

Issue #2726, opened 2023-04-27
Property 'slot' does not exist on type 'Pick<IntrinsicElements, "div">'


Jungzl:
### Description
Assume you have two components, `CompA` and `CompB`

```vue
<!-- CompA -->
<script lang="ts" setup>
  import { ref } from 'vue'
  const foo = ref('bar')
  defineSlots<{
    default: {
      foo: string,
      buz: number
    }
  }>()
</script>

<template>
  <div>
    <slot v-bind="{ foo }" />
    <!-- ^^^ Property 'slot' does not exist on type 'Pick<IntrinsicElements, "div">'. -->
  </div>
</template>
```

```vue
<!-- CompB -->
<script lang="ts" setup>
  import CompA from './CompA.vue'; // using `unplugin-vue-components` and comment this line will throw error below
</script>

<template>
  <CompA>
    <!-- ^^^ Property 'CompA' does not exist on type 'Pick<IntrinsicElements, "template"> & Pick<IntrinsicElements, "div">'. -->
    <template #default="{ foo, buz }">
      <div>
        {{ foo }}{{ buz }}
      </div>
    </template>
  </CompA>
</template>
```

### Reproduction
[Github](https://github.com/Jungzl/vite-vue-volar-test)
[Stackblitz](https://stackblitz.com/edit/vitejs-vite-2gdyij?file=src%2Fcomponents%2FCompA.vue,src%2Fcomponents%2FCompB.vue&terminal=dev)

1. run `pnpm run check:type` throw 1 error.
2. comment import line in CompB & rerun `pnpm run check:type` throw 3 errors.

```node
> vue-tsc

src/components/CompB.vue:10:4 - error TS2339: Property 'CompA' does not exist on type 'Pick<IntrinsicElements, "template"> & Pick<IntrinsicElements, "div">'.

10   <CompA>
      ~~~~~

src/components/CompA.vue:21:6 - error TS2339: Property 'slot' does not exist on type 'Pick<IntrinsicElements, "div">'.

21     <slot v-bind="{ foo }" />
        ~~~~

src/components/CompB.vue:17:5 - error TS2339: Property 'CompA' does not exist on type 'Pick<IntrinsicElements, "template"> & Pick<IntrinsicElements, "div">'.

17   </CompA>
       ~~~~~


Found 3 errors in 2 files.

Errors  Files
     2  src/components/CompB.vue:10
     1  src/components/CompA.vue:21
```

### Environment Info
OS: Windows 10
[...]

PR #3287, opened 2023-06-14
test: move the test for #2726 in strict-template directory

sapphi-red:
Moved the test for #2726 to `strict-template` directory as there was a TODO comment.


johnsoncodehk:
Thanks!

PR #3300, opened 2023-06-18
fix: symbol types are lost

### LINKED ISSUE #3295: Element implicitly has an 'any' type because expression of type 'symbol' can't be used to index type
I'm trying to use symbols to access an object's property, which is valid TS (as you can see [here](https://www.typescriptlang.org/play?#code/MYewdgzgLgBBCeBbGBeGBlJAjEAbAFAOQKKECUA3AFBXC4CGEEMAKgKbQwDeVMMA2iQC6qGADN6uCG2oBfGqEiwATm1DKAJqLBsA7qw5R8lWuAh42AOlwgA5vlXqNgpELJA)).  The code that I'm using is [this](https://play.vuejs.org/#eNpNkL1uhDAQhF9lRQMnJdAnHFK6K5IqdCGFDxaw5D+tbU4I8e5ZcHHpPN6Zb8fesg/nyiVi9pbVAbVTImDTGYBaUJC9SoLlIBdYXuV47TLC3tLw41f922XNTdYVD1Ooeqbq6h+Ppe9JugAeQ3SghJmYFDwDOiO1sxRgA8IRdhjJasi5VP5+JHtrPOdWDVf4XvXdqiJnlV/SVAnvoUW2bEeFsxY7A0Vkw/4kpNo84i2FwccZKi4HpqqgnaWHwaI3eYAwk32AMIBEll7gHo87hFv79Qnsm8UizcQnH9EnvlVYKjsVaUu5CBXx7HI5viI9vsn2PwrqgRI=), which throws the error `Element implicitly has an 'any' type because expression of type 'symbol' can't be used to index type` in VSCode using v1.8.0 of Language Tools (both TypeScript Vue Plugin and Vue Language Features).  I have also done [Takeover Mode](https://vuejs.org/guide/typescript/overview.html#volar-takeover-mode).  Am I missing anything else, or is this a bug?

so1ve:
closes #3295


johnsoncodehk:
In my memory this will cause other problems, will confirm it later.

hexiaokang (approved):

PR #3305, opened 2023-06-19
refactor: simplify __VLS_IsAny

so1ve:
Thus we don't have to use a hack to solve https://github.com/microsoft/TypeScript/issues/54630


johnsoncodehk:
LGTM, thanks!

PR #3324, opened 2023-06-25
fix: allow passing inline components or without using defineComponent

### LINKED ISSUE #3171: "TS1005: ',' expected" in component :is object literal
```html
<template>
  <div>
    <component :is="{ render: eventSummary }" />
  </div>
</template>

<script setup lang="ts">
  import { h } from 'vue'

  const eventSummary = () => h('div')
</script>
```
worked in 1.4.4, fixed in 1.6.0, broken again in 1.6.4

generated code is
```js
const __VLS_4 = { __VLS_ctx.render: __VLS_ctx.eventSummary };
```

should be
```js
const __VLS_4 = { render: __VLS_ctx.eventSummary };
```

so1ve:
fix #3171

### DIFF
--- a/packages/vue-language-core/src/generators/template.ts
+++ b/packages/vue-language-core/src/generators/template.ts
@@ -665,7 +665,7 @@ export function generate(
 		else if (dynamicTagExp) {
 			codes.push(
 				`const ${var_originalComponent} = `,
-				...createInterpolationCode(dynamicTagExp.loc.source, dynamicTagExp.loc, dynamicTagExp.loc.start.offset, capabilitiesPresets.all, '', ''),
+				...createInterpolationCode(dynamicTagExp.loc.source, dynamicTagExp.loc, dynamicTagExp.loc.start.offset, capabilitiesPresets.all, '(', ')'),
 				';\n',
 			);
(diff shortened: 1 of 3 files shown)


so1ve:
@johnsoncodehk I think we should allow passing inline components since they just work well :(

johnsoncodehk:
We need to avoid introducing additional conventions (`__isFragment`, `__isTeleport`, `__isSuspense` etc).
Perhaps we could append `as any` to ignore the type when the `:is` expression is an object literal.

so1ve:
However, if we do so, type-checking will be disabled. I think my approach is fine, just need to think a new way to detect whether the argument is defined by using `defineComponent`.

so1ve:
How about wrap it with `defineComponent` when the expression is an object literal?

johnsoncodehk:
It cannot be perfectly supported, and if we decide to support object literals, we should also handle edge cases like `:is="condition ? { render } : { props }"`.

Is your use case specific to the render option? If so, you can replace `:is="{ render: eventSummary }"` with `:is="eventSummary"`, since functional component is a valid type.

PR #3333, opened 2023-06-27
fix: properly get components

### LINKED ISSUE #1886: A local component cannot override a global component
_Initial issue: nuxt/framework#7401_

It seems like Volar can't correctly find a type of Functional Component when it is used inside the template.

image

**Environment:**
VSCode + Volar + TakeOver mode 4.8.2

so1ve:
close #1886


packages/vue-test-workspace/vue-tsc/non-strict-template/#2157/main.vue
```diff
@@ -0,0 +1,11 @@
+<script setup lang="ts">
+declare module 'vue' {
+  export interface GlobalComponents {
+    Transition: typeof import('vue')['Transition'];
+  }
+};
```
rchl, line 6: Even though it's within a specific file, it will affect all tests, I believe.
If so then maybe use some more unique name so that there is no chance of affecting other tests with that?

so1ve:
Just updated

PR #3338, opened 2023-06-27
fix: add missing nativeTags

### LINKED ISSUE #3337: `Property button does not exist on type IntrinsicElements`
Since Volar 1.8.0, I have been getting typescript errors in my code for built-in html tags like `button` and `label`: 
image

Reverting to 1.6.5 appears to fix the issue.

A few project-specific details that may or may not be relevant:
- Project is Vue 2 (2.6.12)
- Issue occurs in components that are class style (with `vue-class-component`) or with composition api (`export default defineComponent(...)`)
- I am utilizing "takeover mode," and have disabled the builtin typescript support for this project
- using typescript `4.5.5`

so1ve:
close #3337


so1ve:
Oops, they are already in the list :(

PR #3350, opened 2023-07-02
fix: support type narrowing for components

### LINKED ISSUE #3138: TS2345: Argument of type '{}' is not assignable to parameter of type 'never'
After upgrading from vue-tsc@1.6.0 to 1.6.4, I get this error when running `npm run type-check`:
```
TS2345: Argument of type '{}' is not assignable to parameter of type 'never'
```

The error is caused by `<reload-prompt />`

This is the component for which the error occurs:

```vue
// ReloadPromptSsr.vue

<script setup lang="ts">
import { defineAsyncComponent } from 'vue'

const ReloadPrompt =
  typeof window !== 'undefined'
    ? defineAsyncComponent(() => import('@/components/ReloadPrompt.vue'))
    : undefined
</script>

<template>
  <template v-if="ReloadPrompt">
    <reload-prompt />
  </template>
</template>
```

```vue
// ReloadPrompt.vue
<script setup lang="ts">
import { ref, watch } from 'vue'
import { useRegisterSW } from '@/plugins'

const { offlineReady, needRefresh, updateServiceWorker } = useRegisterSW({
  onRegisteredSW(swScriptUrl: string, sw?: ServiceWorkerRegistration): void {
    if (!sw) return

    setInterval(() => sw.update().then(), 60000 * 5)
  },
  onRegisterError(error): void {
    console.error('Failed service worker registration:', error)
  },
  onNeedRefresh(): void {
    console.log('The app has been updated.')
  },
  onOfflineReady(): void {
    console.log('The service worker is ready to work offline.')
  }
})

const snackbar = ref<boolean>(false)

function update(value: boolean): void {
  snackbar.value = value
}

function updateSW(): void {
  updateServiceWorker().then(() => window.location.reload())
}

watch(() => offlineReady.value || needRefresh.value, update)
</script>

<template>
  <v-snackbar
    v-if="needRefresh"
    v-model="snackbar"
    location="bottom right"
    rounded="xl"
    timeout="-1"
    vertical
  >
    <div class="text-subtitle-1 pb-2">App update</div>
    <p>A new update is available.</p>

    <template v-slot:actions>
      <v-btn variant="text" @click="updateSW">Update</v-btn>
      <v-btn variant="text" @click="update(false)">Close</v-btn>
    </template>
  </v-snackbar>
</template>

<style scoped lang="scss"></style>
```

Used packages:

```
typescript: 4.9.5
vite: 4.3.5
vue: 3.2.47
vue-tsc: 1.6.4
```

so1ve:
closes #3138


so1ve:
I have to discard the changes in workflow file :(

rchl:
I think you just need to run `pnpm i` to update the lock file to fix the CI.

so1ve:
> I think you just need to run `pnpm i` to update the lock file to fix the CI.

Yes, you are 100% right :)

However Johnson didn't update the lockfile in `master` branch, so we may push to `master` directly.

rchl:
I see that your workflow changes change line endings from Windows to Unix which seems like a good change overall but probably should be done separately.
(test.yml workflow is also using Windows line-endings)

rchl:
BTW. I don't have access to be able to push to master but if you do then I suppose it might be a good idea.
Or otherwise we could create a separate PR that does just that.

rchl:
https://github.com/vuejs/language-tools/pull/3352

so1ve:
> BTW. I don't have access to be able to push to master but if you do then I suppose it might be a good idea. Or otherwise we could create a separate PR that does just that.

I have the access to, but I think pushing directly without Johnson's permission is not a good idea.

rchl:
I don't have write access to the repo so I can't help directly but you should be able to just merge `master` into your branch and those changes should go away.

johnsoncodehk:
@rchl I added write permissions to you in case I get hit by a bus.

PR #3384, opened 2023-07-08
test: change defineComponent to support JSX

chakAs3:
Function Signature [​](https://vuejs.org/api/general.html#function-signature)

defineComponent() also has an alternative signature that is meant to be used with Composition API and [render functions or JSX](https://vuejs.org/guide/extras/render-function.html).

so I changed the test to match the new API and added test for JSX based component

```ts
// options syntax
function defineComponent(
  component: ComponentOptions
): ComponentConstructor

// function syntax (requires 3.3+)
function defineComponent(
  setup: ComponentOptions['setup'],
  extraOptions?: ComponentOptions
): () => any
```


johnsoncodehk:
Thanks!

PR #3393, opened 2023-07-11
Use faster method after config change

zardoy:
Though I wonder wether we cant just update initialization options and use regular server restart like:
```ts
context.subscriptions.push(vscode.commands.registerCommand('volar.action.restartServer', async () => {
			clients[0]!.clientOptions.initializationOptions = await getInitializationOptions(ServerMode.PartialSemantic, context)
			clients[1]!.clientOptions.initializationOptions = await getInitializationOptions(ServerMode.Syntactic, context)
            // ...
```

### DIFF
--- a/packages/vscode-vue/src/common.ts
+++ b/packages/vscode-vue/src/common.ts
@@ -143,7 +143,7 @@ async function doActivate(context: vscode.ExtensionContext, createLc: CreateLang
 			'Reload Window'
 		);
 		if (reload === undefined) return; // cancel
-		vscode.commands.executeCommand('workbench.action.reloadWindow');
+		vscode.commands.executeCommand('workbench.action.restartExtensionHost');
 	}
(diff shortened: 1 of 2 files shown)


packages/vscode-vue/src/common.ts
```diff
@@ -143,7 +143,7 @@ async function doActivate(context: vscode.ExtensionContext, createLc: CreateLang
 			'Reload Window'
 		);
 		if (reload === undefined) return; // cancel
-		vscode.commands.executeCommand('workbench.action.reloadWindow');
+		vscode.commands.executeCommand('workbench.action.restartExtensionHost');
```
zardoy, line 146: this is faster than reloading whole window as it reloads extension host only (instead of reloading other core vscode things as well)

for example in some cases it can save you from losing terminals state

johnsoncodehk:
Yes we can do this, thanks for suggesting!

PR #3401, opened 2023-07-14
Do not hide output channel on server restart :rocket:

zardoy:
Since output channels are created automatically right now, when you restart server, opened output channel disappears and you have to find it again... Right there is no api in vscode to focus specific output channel :(


johnsoncodehk:
Thanks!

PR #3418, opened 2023-07-18
fix: specify compatible engines

### LINKED ISSUE #3417: Specify engines in package.json
I note that `??=` is used in many places, so I think the node version should be not less than 16.

But now `engines` isn't specified in package.json. Maybe a user will run with Node 14 and get error from `??=`.

so1ve:
close #3417


johnsoncodehk:
I decided not to add it, I wanted to avoid package.json getting bloated. And I think in the real world, no users are really helped by the `engines` field. Thanks!

PR #3426, opened 2023-07-20
ci: use actions as the default committer

so1ve:
Linus is pushing these automated changes now 😂


johnsoncodehk:
Thanks! I will correct the authors of those commits after merging other PRs.

PR #3442, opened 2023-07-23
fix: generate valid code when using __VLS_PropsChildren

so1ve:


rchl:
Is this not testable?

so1ve:
@rchl We can't test it (and we don't have to) because it doesn't break the functionality without this fix.

rchl:
Why we can't test it? Is it because the type is `any` without this change and still doesn't fail?

so1ve:
@rchl Though it creates invalid code without this fix, it is functioning correctly - correct types and completions.

PR #3456, opened 2023-07-29
chore: support custom configPath for createParsedCommandLineByJson

qmhc:
I want to parse config content after overriding and I found two methods: `createParsedCommandLine` and `createParsedCommandLineByJson`.

`createParsedCommandLine` parse config via path and cannot specify the config content.

`createParsedCommandLineByJson` will always set config path to `rootDir + '/jsconfig.json'`.

So it would be great if `createParsedCommandLineByJson` can support custom config path.


johnsoncodehk:
LGTM, thanks!

PR #3465, opened 2023-08-02
test: validate types in JS script tags

rchl:
Enable type validation in plain JS script tags in tests.

Extracted from #3353


johnsoncodehk:
`non-strict-template/#3353` is deliberate to not report errors, to make sure our codegen will not make TS errors reporting for JS script blocks by default.

Can you create tsconfig in a new directory under `packages/vue-test-workspace/vue-tsc` to enable `checkJs` and place the new test?

rchl:
> `non-strict-template/#3353` is deliberate to not report errors, to make sure our codegen will not make TS errors reporting for JS script blocks by default.

Are you sure? Reading https://github.com/vuejs/language-tools/issues/3121 it appears to be about particular slot syntax not reporting errors rather than the whole template not reporting errors. With `checkJs` enabled and the fix for the original issue reverted it will still fail.

johnsoncodehk:
Maybe I didn't make it clear enough. We do want #3353/main.vue to not report an error because the same code in .js will also not report an error with checkJs disabled, and we want the script tag to behave consistently with .js.

PR #3487, opened 2023-08-17
fix: generic components should respect `strictTemplates`

### LINKED ISSUE #3201: Type-checking errors on fallthrough attributes and AllowedComponentProps
Consider a simple component using a `Field` from the [vee-validate](https://vee-validate.logaretm.com/v4) library:

```vue
<script setup lang="ts">
import { Field } from 'vee-validate';
</script>

<template>
  <Field id="name" name="name" value="Cedric" />
</template>
```

Note that `Field` has an `id`, which is not a prop but a fallthrough attribute, so the resulting HTML is:
```html
<input id="name" name="name">
```

The problem is `vue-tsc` (v1.6.5) does not compile, as `id` is not recognized:
```
npm run type-check
src/App.vue:6:10 - error TS2345: Argument of type '{ id: string; name: string; value: string; }' is not assignable to parameter of type '{ label?: string | undefined; as?: string | Record<string, any> | undefined; bails?: boolean | undefined; uncheckedValue?: any; validateOnInput?: boolean | undefined; validateOnChange?: boolean | undefined; ... 21 more ...; onVnodeUnmounted?: ((vnode: VNode<...>) => void) | ... 1 more ... | undefined; }'.
  Object literal may only specify known properties, and 'id' does not exist in type '{ label?: string | undefined; as?: string | Record<string, any> | undefined; bails?: boolean | undefined; uncheckedValue?: any; validateOnInput?: boolean | undefined; validateOnChange?: boolean | undefined; ... 21 more ...; onVnodeUnmounted?: ((vnode: VNode<...>) => void) | ... 1 more ... | undefined; }'.

6   <Field id="name" name="name" value="Cedric" />
```

This only happens with `strictTemplates: true` in `vueCompilerOptions`.

I used to get around this by using `AllowedComponentProps`:
```
declare module '@vue/runtime-core' {
  interface AllowedComponentProps {
    id?: string;
  }
}
```

but this does not do the trick here, as now `Field` is not found at all:
```
src/App.vue:6:4 - error TS2339: Property 'Field' does not exist on type '{}'.
```

You can find a small repro here: https://stackblitz.com/edit/github-ekza7z?file=src/App.vue

```
npm i && npm run type-check
```

So my questions are:
- is `AllowedComponentProps` the correct way to handle the case of attributes here?
- if so, as it does break the compilation by not finding `Field` anymore, is there an issue with vue-tsc?
- or is it something that the vee-validate author could do something about so it would work out of the box?

### LINKED ISSUE #3318: Generic & Falltrough Attributes: Object literal may only specify known properties...
### Vue version

3.3.4

### Description

If the component is generic vue-tsc complains when an html attribute is bound to it.

### What is expected?

I expect to bind any html attribute to a component without errors.

Volar and vue-tsc show an error on the first attribute, thinking that it's a prop and it needs to be defined on the child component.

### REPRO
1. Open: [stackblitz](https://stackblitz.com/edit/vitejs-vite-zvqqjp?file=tsconfig.json,src%2Fcomponents%2FGeneric.vue,src%2Fcomponents%2FSimple.vue,src%2FApp.vue&terminal=dev)

2. Run: ```npx vue-tsc```

3. Full Error: 
```bash
src/App.vue:11:25 - error TS2345: Argument of type '{ val: any; id: string; }' is not assignable to parameter of type '{ val: any; } & VNodeProps & AllowedComponentProps & ComponentCustomProps'.
  Object literal may only specify known properties, and 'id' does not exist in type '{ val: any; } & VNodeProps & AllowedComponentProps & ComponentCustomProps'.

11   <Generic :val="value" id="id" />
                           ~~


Found 1 error in src/App.vue:11
```

### Probably related issues:
#3201

so1ve:
close #3201, close #3318, close vuejs/core#8969


johnsoncodehk:
Thanks for the PR, but this will break vue-tsc emit types (we currently lack tests for vue-tsc emit dts). I will look for a workaround.

johnsoncodehk:
The original fix has been superseded by https://github.com/vuejs/language-tools/pull/3487/commits/8e6a8a97d9525b0edddaaf178f1c511e7c8831bb, the new changes will not affect emit types.

so1ve:
Wow that would be nice! This commit fixed this issue on consumer side right?

johnsoncodehk:
> This commit fixed this issue on consumer side right?

Right, we should avoid `strictTemplates` from affecting the definition side.

PR #3527, opened 2023-09-01
fix: should generate code for slots that uses `v-else-if` and `v-else` directives

### LINKED ISSUE #3466: False ts(6133) error message
### Issue:
VSCode with latest Volar (1.8.8) complains about values that are declared and never read if values are used inside `template` with `v-else` of `v-else-if`.

### Reproduction:
1. Clone this repo: [https://github.com/KamilBeda/nuxt-vs-code-typescript-bug](https://github.com/KamilBeda/nuxt-vs-code-typescript-bug)
2. Install dependencies
3. Open app.vue and look at the error at `doSomething` function

### Temporary fix:
Reverting Volar to 1.8.7

so1ve:
close #3466


so1ve:
Not the right fix

PR #3528, opened 2023-09-02
feat: support nested plugins

### LINKED ISSUE #3437: Support nested plugins
In Rollup and Vite, we can put nested plugins in `plugins` property, like `{ plugins: [a, [b, c]]}`

I hope that we can also support this feature in `vueCompilerOptions.plugins`. So that we can avoid requiring users to configure all of the plugins.
https://vue-macros.sxzz.moe/guide/bundler-integration.html#volar-support

so1ve:
close #3437

/cc @sxzz

### DIFF
--- a/packages/vue-language-core/src/utils/ts.ts
+++ b/packages/vue-language-core/src/utils/ts.ts
@@ -1,6 +1,6 @@
 import type * as ts from 'typescript/lib/tsserverlibrary';
 import * as path from 'path';
-import type { RawVueCompilerOptions, VueCompilerOptions, VueLanguagePlugin } from '../types';
+import type { RawVueCompilerOptions, VueCompilerOptions, VueLanguagePluginOption } from '../types';
 
 export type ParsedCommandLine = ts.ParsedCommandLine & {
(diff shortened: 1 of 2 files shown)


so1ve:
Ugh, seems plugin system will be rewritten in volar 2.0.

packages/vue-language-core/src/utils/ts.ts
```diff
@@ -158,7 +158,7 @@ function getPartialVueCompilerOptions(
 	}
 	if (rawOptions.plugins) {
 		const plugins = rawOptions.plugins
-			.map<VueLanguagePlugin | undefined>((pluginPath: string) => {
+			.map<VueLanguagePlugin[] | VueLanguagePlugin | undefined>((pluginPath: string) => {
```
sxzz, line 161: Supporting deep nested plugins will be even better. `[a, [b, [c, d], e]]`
https://github.com/sxzz/vite/blob/fba97c8119701de256efef581768f98c2348b322/packages/vite/src/node/config.ts#L128-L134

so1ve: Done

PR #3530, opened 2023-09-02
feat: support nested plugins

### LINKED ISSUE #3437: Support nested plugins

so1ve:
close #3437, supersedes #3528

### DIFF
--- a/packages/vue-language-core/src/types.ts
+++ b/packages/vue-language-core/src/types.ts
@@ -62,6 +62,11 @@ export type VueLanguagePlugin = (ctx: {
 	resolveEmbeddedFile?(fileName: string, sfc: Sfc, embeddedFile: VueEmbeddedFile): void;
 };
 
+export type VueLanguagePluginOption =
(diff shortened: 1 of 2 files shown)


so1ve:
Currently I simply flatten all nested plugins. @sxzz How should we handle nested plugins with circular references? How does vite handle that case?

sxzz:
Could you please make an example for me?

so1ve:
Ugh. For example:

```ts
const plugin = []

plugin.push(plugin)
```

sxzz:
@so1ve Vite and rollup won't handle this case. It's not a valid usage.

johnsoncodehk:
Thanks!

PR #3534, opened 2023-09-05
refactor: Removing `__VLS_types.d.ts`

johnsoncodehk:
As part of the preparatory work for migrating to ts plugin in 2.0, we cannot use global virtual files because tsserver will crash if the files do not actually exist, and there is no way to bypass this problem.

The solution here is to attach the global types code to one of the virtual vue ts file instead of creating a `__VLS_types.d.ts` file.

- [ ] When the attached file is deleted, it should be attached to another file instead.
- [ ] Simplify implementation
- [ ] Ensure global types are not emitted with `vue-tsc --emit`
- [ ] Considering that VueFile is actually reused across multiple tsconfig projects, it is necessary to ensure that in this case a project does not have multiple VueFiles with global types attached.

### DIFF
--- a/packages/vue-language-core/src/generators/script.ts
+++ b/packages/vue-language-core/src/generators/script.ts
@@ -14,6 +14,7 @@ import { getSlotsPropertyName } from '../utils/shared';
 import { walkInterpolationFragment } from '../utils/transform';
 import * as sharedTypes from '../utils/globalTypes';
 import * as muggle from 'muggle-string';
+import { geneateGlobalTypes } from '../utils/globalTypes';
 
 export function generate(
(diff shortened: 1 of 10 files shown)


rchl:
Does that mean that extending types through `experimentalAdditionalLanguageModules` like in https://github.com/rchl/volar-component-types/blob/5f10b2dd17f7c19490f528e966d8591f3d26dba6/src/volar/volar.ts#L31-L48 will also be rendered inoperable?

johnsoncodehk:
Yes, and we should have a new api via plugin for adding custom global types.

johnsoncodehk:
Closed as this implementation is obsolete and will open a new PR to do this.

PR #3536, opened 2023-09-06
fix(vue-component-type-helpers): correctly handle generic components when using `ComponentExposed`

### LINKED ISSUE #3206: Cannot use `InstanceType<typeof Comp<any>>>` on generic components
Since the push of adding generic components on Vue 3.3.0, we don't seem able to get the `InstanceType` aka `public instance` anymore.

```html
<script setup lang="ts" generic="T">
const props = withDefaults(defineProps<{
	value?: T | null;
	list: T[];
}>(), {
	value: null,
});
</script>

<template>
    <select>
       // code
    </select>
</template>
```

```html
<script lang="ts">
import MyComp from './MyComp.vue'

// This has error
const el = ref<null | InstanceType<typeof MyComp<any>>>( null )

</script>
<template>
   <my-comp ref="el" list="[]"/>
</template>
```

so1ve:
close #3206, close vuejs/core#8373

### DIFF
--- a/packages/vue-component-type-helpers/index.d.ts
+++ b/packages/vue-component-type-helpers/index.d.ts
@@ -24,7 +24,7 @@ export type ComponentEmit<T> =
 
 export type ComponentExposed<T> =
 	T extends new () => infer E ? E :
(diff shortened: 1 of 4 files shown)


johnsoncodehk:
LGTM, thanks!

Issue #3345, opened 2023-06-30
Build failed, error thrown


krzar:
When i'm trying to build my application with `vue-tsc --noEmit && vite build --emptyOutDir` I'm getting this error all the time:

```
/var/www/frontend/node_modules/vue-tsc/bin/vue-tsc.js:64
                        throw err;
                        ^

Error: Debug Failure. False expression: Paths must either both be absolute or both be relative
    at getRelativePathFromDirectory (/var/www/frontend/node_modules/typescript/lib/tsc.js:5637:9)
    at relativeToBuildInfo (/var/www/frontend/node_modules/typescript/lib/tsc.js:117963:38)
    at toFileId (/var/www/frontend/node_modules/typescript/lib/tsc.js:117968:22)
    at arrayFrom (/var/www/frontend/node_modules/typescript/lib/tsc.js:715:24)
    at toFileIdListId (/var/www/frontend/node_modules/typescript/lib/tsc.js:117974:21)
    at /var/www/frontend/node_modules/typescript/lib/tsc.js:117897:32
    at mapDefined (/var/www/frontend/node_modules/typescript/lib/tsc.js:300:22)
    at getBuildInfo2 (/var/www/frontend/node_modules/typescript/lib/tsc.js:117893:26)
    at newProgram.getBuildInfo (/var/www/frontend/node_modules/typescript/lib/tsc.js:118145:41)
    at Object.getBuildInfo (/var/www/frontend/node_modules/typescript/lib/typescript.js:117443:70)

Node.js v18.16.1
error Command failed with exit code 1.
```

I've this error before, but usually running `yarn upgrade` helped, now it doesn't even help (probably no package is getting updated). I was trying to remove node_modules and yarn.lock but it's still not working.

so1ve:
Could you please provide a minimal reproduction?

krzar:
Unfortunately, I don't know what is causing the problem, but I managed to solve it for now by reverting to an older yarn. lock in my project. I suspect that some package is conflicting, but I'm not sure which one exactly.

BenShelton:
## Reproduction

Seems related to https://github.com/vuejs/language-tools/issues/2622 which has been closed, but this is still happening on latest `vue-tsc` and `typescript` versions.

Here is a reproduction: https://codesandbox.io/p/sandbox/vue-tsc-debug-failure-n6l43n?welcome=true

[...]

Issue #2622, opened 2023-04-18
Throw an error when `composite` is enabled


sxzz:
Reproduction: https://github.com/sxzz/volar-starter/commit/e83eb070283dc0c06692b5d1bf7c4e486419e559

```
/project/node_modules/.pnpm/vue-tsc@1.3.16_typescript@5.0.4/node_modules/vue-tsc/bin/vue-tsc.js:53
                        throw err;
                        ^

Error: Debug Failure. False expression: Paths must either both be absolute or both be relative
    at getRelativePathFromDirectory (/project/node_modules/.pnpm/typescript@5.0.4/node_modules/typescript/lib/tsc.js:5717:9)
    at relativeToBuildInfo (/project/node_modules/.pnpm/typescript@5.0.4/node_modules/typescript/lib/tsc.js:117152:38)
    at convertToReusableDiagnosticRelatedInformation (/project/node_modules/.pnpm/typescript@5.0.4/node_modules/typescript/lib/tsc.js:117237:18)
    at /project/node_modules/.pnpm/typescript@5.0.4/node_modules/typescript/lib/tsc.js:117223:20
    at Array.map (<anonymous>)
    at convertToReusableDiagnostics (/project/node_modules/.pnpm/typescript@5.0.4/node_modules/typescript/lib/tsc.js:117222:22)
    at getBuildInfo2 (/project/node_modules/.pnpm/typescript@5.0.4/node_modules/typescript/lib/tsc.js:117099:11)
    at newProgram.getBuildInfo (/project/node_modules/.pnpm/typescript@5.0.4/node_modules/typescript/lib/tsc.js:117309:41)
    at Object.getBuildInfo (/project/node_modules/.pnpm/typescript@5.0.4/node_modules/typescript/lib/typescript.js:116651:70)
    at emitBuildInfo (/project/node_modules/.pnpm/typescript@5.0.4/node_modules/typescript/lib/typescript.js:109453:30)
```

Mostaard:
I'm encountering the same issue

stefanprobst:
i am getting the same error with `"incremental": true`

FabianMontoya:
I have the same error:

```
> vue-tsc --noEmit

/Users/fabian.montoya/git/vue-app/node_modules/vue-tsc/bin/vue-tsc.js:53
			throw err;
			^

Error: Debug Failure. False expression: Paths must either both be absolute or both be relative
[...]

PR #3556, opened 2023-09-12
fix(vue-tsc): prevent rebuild in incremental mode throwing error

blake-newman:
Fix rebuilding in incremental mode with tsbuildinfo file causing error.

Secondary rebuilds cause the following error:

```
Error: Debug Failure. False expression: Paths must either both be absolute or both be relative
```

This is caused because the mapping of the exportedModules looks for `__vls__` file id, but this has been removed in the other patches. This new patch filters this file which is virtual and removes it before mapping. This prevents the error occurring on rebuilds and composite mode. 

resolves: #3345
resolves: #2622


johnsoncodehk:
Thanks!

PR #3565, opened 2023-09-16
fix: avoid losing generic types

### LINKED ISSUE #3561: Generic SFC broken from 1.8.10, 1.8.11
I had a generic component that was working, but it isn't resolving correctly in version 1.8.10 and 1.8.11, but if I downgrade to 1.8.8, it works fine.

In this example in index.js, `item` is just resolving to `unknown`, which results in a lot of errors.

### index.vue
```JS
<CarouselSelect
  v-slot="{ item }"
  v-model="componentState.selected"
  :items="componentState.items"
  :get-key="(item) => item.id"
  class="mb-3"
>
  {{item.title}}
</CarouselSelect>
```

### CarouselSelect.vue
```JS
<script setup lang="ts" generic="T, TValue extends undefined|string|number|symbol">
const $props = defineProps<{
  items: Array<T>,
  modelValue: TValue,
  getKey: (item: T) => TValue,
}>();

defineSlots<{
  default: { item: T },
}>();
//...
</script>
```

so1ve:
close #3561

My implementation is terribly ugly so I need more time to refactor it.

### DIFF
--- a/packages/vue-language-core/src/utils/globalTypes.ts
+++ b/packages/vue-language-core/src/utils/globalTypes.ts
@@ -97,7 +97,7 @@ declare function __VLS_asFunctionalComponent<T, K = T extends new (...args: any)
 	}) => JSX.Element & { __ctx?: typeof ctx & { props?: typeof props; expose?(exposed: K): void; } }
 	: T extends () => any ? (props: {}, ctx?: any) => ReturnType<T>
 	: T extends (...args: any) => any ? T
-	: (_: T extends import('${vueCompilerOptions.lib}').VNode | import('${vueCompilerOptions.lib}').VNode[] | string ? {}: T${vueCompilerOptions.strictTemplates ? '' : ' & Record<string, unknown>'}, ctx?: any) => { __ctx?: { attrs?: any, expose?: any, slots?: any, emit?: any, props?: T${vueCompilerOptions.strictTemplates ? '' : ' & Record<string, unknown>'} } }; // IntrinsicElement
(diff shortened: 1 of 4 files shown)


packages/vue-language-core/src/utils/globalTypes.ts
```diff
@@ -118,6 +118,8 @@ declare function __VLS_componentProps<T, K>(comp: T, fnReturn: K):
 	__VLS_PickNotAny<K, {}> extends { __ctx: { props: infer P } } ? NonNullable<P>
 	: T extends (props: infer P, ...args: any) => any ? NonNullable<P> :
 	{};
+
+declare function __VLS_cast<T>(x: T): T;
```
rchl, line 124: Does that really work? I would imagine that it would cause a type error when passed x doesn't match T. To cast it would need to take unknown argument or something

so1ve: @rchl Ahhhh this is not `casting` some type to another type. This function is created to make the error more reasonable:

https://www.typescriptlang.org/play?#code/CYUwxgNghgTiAEAzArgOzAFwJYHtXzCgGcMAeAFQD4AKDALnnIEoHyBuAKA7DxKRxzwAvPADe8KAwCM8AL4SiYiQxIwsqAOZz4Aeh2MAFlkVEDOAO6KMBhCBgwcMeDkTwMATwAOCHqgBudkS4qNy8GPAARrDCBMRk4pLwqupasjQJ0nJMuvrkRiZmlm428HYOTi5uXggAtsY1UBhgBhxAA

PR #3573, opened 2023-09-20
fix: format slots with typeannotation correctly

### LINKED ISSUE #3572: Format code problem
When I use volar to auto fomat the sfc file, I found a strange problem. here is an example file: 
```vue
<script setup lang="ts">
interface Row {
  meow: any
}
</script>

<template>
  <div class="">
    <el-table-column prop="prop" label="label" width="80" align="center" v-for="meow in  13 ">
      <template #default="{ row }: { row: Row }">{{
        row.meow + meow
        }}</template>
    </el-table-column>
  </div>
</template>

<style scoped lang="scss"></style>
```
Every time to format the code with volar, the code `v-for="meow in 13"` will insert some space and it will be:
```
v-for="meow in                             13                            ">
```
and longer.
And if change `#default="{ row }: { row: Row }"` to `#default="{ row }"`, which means drop the type, the `v-for` will no longer be longer, and the problem disappear.
My volar version is 1.8.11.

so1ve:
close #3572


johnsoncodehk:
Thanks!

PR #3575, opened 2023-09-20
fix: generate valid syntax when noPropertyAccessFromIndexSignature is not enabled

### LINKED ISSUE #3574: Regression in vue-tsc 1.8.12: custom event types no longer inferred
With the update from vue-tsc 1.8.11 to 1.8.12 the types of custom event aren’t inferred anymore.

Comp.vue:
```vue
<script lang="ts" setup>
defineEmits<{
  (event: 'change', value: string): void
}>()
</script>
```

Parent.vue:
```vue
<template>
  <Comp @change="(value) => console.log(value)" />
  <!--            ^^^^^ is inferred as `any` now, should be `string` -->
</template>

<script lang="ts" setup>
import Comp from './Comp.vue'
</script>
```

so1ve:
close #3574


so1ve:
Forgot I could add parenthesis..

PR #3577, opened 2023-09-21
ci: upgrade node version

sxzz:


johnsoncodehk:
Unless we explicitly drop node 16 support, testing on node 16 is required to avoid accidentally introducing post-ES2022 (node ​​18) syntax and breaking downstream CI.

PR #3581, opened 2023-09-21
refactor: don't generate additional virtual code for `defineSlots`

johnsoncodehk:
For `defineSlots`, we can reuse its return type instead of generating duplicate virtual code for template.

For the following vue code.

```html
<script setup lang="ts">
defineSlots<{
  activator?(props: { isActive: boolean }): void
}>();
</script>

<template>
	<!-- ... -->
</template>
```

### Old virtual code

```ts
/* script setup part */
// ...
defineSlots<{
  activator?(props: { isActive: boolean }): void
}>();
// ...

/* template part */
var __VLS_slots!: {
  activator?(props: { isActive: boolean }): void
};
// some template virtual codes reference __VLS_slots
```

### New virtual code

```ts
/* script setup part */
// ...
const __VLS_slots = defineSlots<{
  activator?(props: { isActive: boolean }): void
}>();
// ...

/* template part */
// some template virtual codes reference __VLS_slots
```

### DIFF
--- a/packages/vue-language-core/src/generators/script.ts
+++ b/packages/vue-language-core/src/generators/script.ts
@@ -57,10 +57,11 @@ export function generate(
 			leadingCommentEndOffset: 0,
 			importSectionEndOffset: 0,
 			defineProps: undefined,
+			defineSlots: undefined,
+			slotsAssignName: undefined,
 			propsAssignName: undefined,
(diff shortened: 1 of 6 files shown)


johnsoncodehk:
Also looking to do that for `defineEmit`.

PR #3585, opened 2023-09-23
fix: use isomorphic `__VLS_ConstructorOverloads` when emits type is not inlined

so1ve:
close #3542

Most of the types come from @ShenQingchuan, thanks a lot!

PR #3588, opened 2023-09-24
test: add test for #1855

### LINKED ISSUE #1855: Event type bug in template
Reproduction:  https://github.com/kingyue737/volar-event-bug

1. Clone the repo and run `pnpm install`
2. Find the error in `App.vue`

image

Event `event-any` is defined with `arg: any` but volar thinks the type of handler should be `(arg: string) => void`
It seems this issue happens when one of the events has an argument typed as `any`

```ts
const emit = defineEmits<{
  (e: "event-any", arg: any): void;
  (e: "event-str", arg: string): void;
}>();
```

While volar extension shows ts error in template, `vue-tsc` doesn't complain about such issue.

### Environment
- Volar: 0.40.13
- Vue: 2.7.10
- VSCode: 1.71.0
- Typescript: 4.8.3 (in takeover mode)

so1ve:
close #1855

This issue should be fixed by https://github.com/vuejs/language-tools/commit/6df0db7472db0f8fcb8da22dd68ea584768118aa and https://github.com/vuejs/language-tools/pull/3585

### DIFF
--- a/packages/vue-test-workspace/vue-tsc/non-strict-template/#1855/Comp.vue
+++ b/packages/vue-test-workspace/vue-tsc/non-strict-template/#1855/Comp.vue
@@ -1,0 +1,6 @@
+<script setup lang="ts">
+const emit = defineEmits<{
+	(e: 'event-any', arg: any): void;
+	(e: 'event-str', arg: string): void;
+}>();
+</script>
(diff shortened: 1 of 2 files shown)


johnsoncodehk:
Thanks!

PR #3595, opened 2023-09-26
test: add tests for #3595

so1ve:
close #3595

I'm not sure if this is the right fix, because I found another issue in vue core: https://github.com/vuejs/core/issues/9296 which might be the root cause of this issue

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,10 @@
 # Changelog
 
+## 1.8.15 (2023/9/26)
+
+- fix: props type missing in JS component context ([#3592](https://github.com/vuejs/language-tools/issues/3592))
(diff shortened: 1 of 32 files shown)


johnsoncodehk:
Yes, it's a Vue types problem, we do a hack for this (`bypassDefineComponent` in code). The issue has been fixed by https://github.com/vuejs/language-tools/commit/db8857b40547949f7c45906272f3878ea546f0cc, could you update the PR to add a test?

so1ve:
I've updated this PR!

johnsoncodehk:
We also need to refactor the current vue-tsc test, otherwise this test will not run.

I think the current test directory is a bit confusing, and I want to sort it out first.

PR #3599, opened 2023-09-26
issue: failed test vue-tsc non-strict-template 

vsteenhoudt:
Tests fail on this code from v1.8.14.
No errors on v1.8.13


packages/vue-test-workspace/vue-tsc/non-strict-template/#3598/product-detail.vue
```diff
@@ -0,0 +1,42 @@
+<script setup lang="ts">
```
romainmenke, line 1: Thank you for submitting this patch!

I think we can make this repro smaller so that it's easier for maintainers to analyse the root cause.

I think a single `main.vue` file is sufficient :

```vue
<script setup lang="ts">
defineProps(['id']);
</script>

<script lang="ts">
import { watchEffect } from 'vue';

export default {
	data(): { foo?: {
	id: string;
} } {
		return {
			foo: undefined,
		};
	},

	methods: {
		async getData(id: string) {
			let fetchId = id;

			this.foo = {
				id: fetchId,
			};
		},
	},

	created() {
		watchEffect(async () => {
			await this.getData(this.id);
		})
	},
};
</script>

<template>
	<div class="foo-detail" v-if="!!foo">
		<h1>{{ foo.id }}</h1>
	</div>
</template>
```

-----

Maybe there is still redundant code here?

johnsoncodehk: Right, thanks for simplifying the test code! But the test has been added by #3600, let's close this. 🤞

PR #3600, opened 2023-09-27
fix: merge default export's properties properly

### LINKED ISSUE #3598: Test fails (property-does-not-exist) from v1.8.14
We noticed that are tests are failing on some of our code starting from v1.8.14.
The problem does not occur on version v1.8.13

We added our code to the vue-test-workspace/vue-tsc/non-strict-template and ran the tests under version v1.8.13 so we could confirm there was no problem in this version. 
From v1.8.14 and forward the tests fail on our code.

Edit: I'll fork the repo first. And update it here when ready. 

Kind regards,
Valentijn

### LINKED ISSUE #3620: Volar: custom components not properly recognised or highlighted in non-TS files
see #3299 

For Typescript-enabled projects when a component does not use `setup` the code coloring of custom components is not displayed (they show up as regular HTML tags). I created an [example repository](https://github.com/mgiraldo/vue-color-issue) where I just created a new project with `npm create vue@latest` and these options:

```✔ Add TypeScript? … Yes
✔ Add JSX Support? … No
✔ Add Vue Router for Single Page Application development? … No
✔ Add Pinia for state management? … No
✔ Add Vitest for Unit Testing? … No
✔ Add an End-to-End Testing Solution? › No
✔ Add ESLint for code quality? … Yes
✔ Add Prettier for code formatting? … No
```

image

Then I did a minor change to `.eslintrc.cjs` to use `kebab-case` in `vue/component-name-in-template-casing` and also in `src/components/App.vue` so that it does not use `setup`:

```html
<script>
import HelloWorld from './components/HelloWorld.vue'
import TheWelcome from './components/TheWelcome.vue'

export default {
  components: {HelloWorld, TheWelcome}
}
</script>
```

I would expect code coloring to be the same in both situations.

This is the result before the change:
image

This is after the change:
image

so1ve:
close #3599, close #3598, close #3620


packages/vue-test-workspace/vue-tsc/non-strict-template/#3598/main.vue
so1ve, line 1: Delete this later and rename #3612 to #3598

romainmenke:
Thank you @so1ve and @johnsoncodehk
Everything is green again on our end 🙇

PR #3603, opened 2023-09-28
fix type ComponentExposed

FlareZh:
Fix exposed type which is a ref.

### DIFF
--- a/packages/vue-component-type-helpers/package.json
+++ b/packages/vue-component-type-helpers/package.json
@@ -10,5 +10,8 @@
 		"type": "git",
 		"url": "https://github.com/vuejs/language-tools.git",
 		"directory": "packages/vue-component-type-helpers"
+	},
+	"dependencies": {
+		"@vue/reactivity": "^3.3.0"
 	}
 }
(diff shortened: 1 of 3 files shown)


packages/vue-component-type-helpers/package.json
```diff
@@ -10,5 +10,8 @@
 		"type": "git",
 		"url": "https://github.com/vuejs/language-tools.git",
 		"directory": "packages/vue-component-type-helpers"
+	},
+	"dependencies": {
+		"@vue/reactivity": "^3.3.0"
```
so1ve, line 15: `vue-component-type-helpers` should not depend on `@vue/reactivity`.

FlareZh: Thank you! But how could I ShallowUnwrapRef the exposed?

FlareZh: > `vue-component-type-helpers` should not depend on `@vue/reactivity`.

Sorry I don't knew that. Thank you very much for your PR so soon.

johnsoncodehk:
Thanks for the PR! But it has been resolved by #3604. 🤞

PR #3624, opened 2023-10-05
fix: make emits type correct when user assigns emit function a custom name

### LINKED ISSUE #3615: Generic components throwing errors since v1.8.14
Generic components started throwing errors once we update from 1.8.13 to 1.8.14 verison.

The problem is, emit type is not inherited from props generic type. In next example, emit is supposed to be `string` type but turns to be `any` now:

image

...

image

...

image

so1ve:
close #3615


filipbekic01:
I'd love to understand what's changed in PR but anyways, thanks!

johnsoncodehk:
Thanks!

PR #3633, opened 2023-10-07
fix: make `defineProps` work when reading a property from `defineProps()`

### LINKED ISSUE #3629: Breaking assigned props variable
Caused in db8857b40547949f7c45906272f3878ea546f0cc

`propsAssignName` is empty if it's not a statement like `const foo = defineProps()`

image

---

My scenario: it's breaking the syntax `defineProps().withDefaults()` in Vue Macros. 
See https://github.com/vue-macros/vue-macros/pull/517

so1ve:
close #3629


packages/vue-test-workspace/vue-tsc/non-strict-template/#3629/main.vue
```diff
@@ -0,0 +1,4 @@
+<script setup lang="ts">
+// @ts-expect-error
```
rchl, line 2: With `@ts-expect-error` this test will pass both with and without the fix so it's missing the point.

so1ve: Ugh, I'm not sure how to fix this since `foo` doesn't exist on `defineProps()`. Any ideas?

johnsoncodehk:
Thanks!

sxzz:
Another case broken
```vue
<script setup lang="ts">
const name = 'title'
defineProps([name])
</script>
```

image

PR #3636, opened 2023-10-07
fix: avoid reading `props` from `__VLS_ctx`

### LINKED ISSUE #3204: Calling prop fields in the template (like `props.foo`) doesn't work with `withDefaults`
This is issue originated from this [vue/core discussion](https://github.com/vuejs/core/issues/8310#issuecomment-1553755650). Apparently it's not a vue/core error.

This is the error:

gen-err

Reproducible:

```vue
<script setup lang="ts" generic="T">
const props = withDefaults(defineProps<{
	value?: T | null;
	list: T[];
}>(), {
	value: null,
});
</script>

<template>
	<select>
		<option v-for="item of props.list">

		</option>
	</select>
</template>
```

### LINKED ISSUE #3405: TS Error on props when using Generic SFC
I created a fresh repo using ```npm init vue@latest``` the repo can be found here, [vue-generic-test](https://github.com/KieranSunshine/vue-generic-test/).

The problematic file is [src/components/GenericComponent.vue](https://github.com/KieranSunshine/vue-generic-test/blob/main/src/components/GenericComponent.vue)

When I try to access a prop I'm my IDE shows the following error:
image

If I remove the usage of the generic from props, the error goes away.

Apologies if this is already a known issue, or if I'm just being dumb. Let me know if I can provide any more information.

To be clear I'm using VS code, volar 1.8.5, typescript 5.0.4, vue 3.3.4, vue-tsc 1.8.5.

so1ve:
close #3204, close #3405


johnsoncodehk:
Good idea 👍, thanks!

PR #3644, opened 2023-10-09
fix: generate `__VLS_props` just in place

### LINKED ISSUE #3643: `defineProps([vars])` is broken
```vue
<script setup lang="ts">
const name = 'title'
defineProps([name])
</script>
```

image

_Originally posted by @sxzz in https://github.com/vuejs/language-tools/issues/3633#issuecomment-1752450511_

so1ve:
close #3643


johnsoncodehk:
For the following source code.

```ts
const name = 'foo';
const foo = defineProps([name]).foo;
```

I think a better way is generate it to:

```ts
const name = 'foo';
const __VLS_props = defineProps([name])
const foo = __VLS_props.foo;
```

We need to return the `const foo = defineProps([name]).foo;` range from `parseScriptSetupRanges()` and then it should be easy to do.

so1ve:
I think so, but IMO this will require much more work.

johnsoncodehk:
Updated, it might be simpler than you expect. ;)

PR #3666, opened 2023-10-15
feat: enforce `v-bind` argument to be an object

### LINKED ISSUE #3625: `v-bind="props"` should type check `props` is an object
Since [vue uses the `in` operator on the value used for v-bind](https://github.com/vuejs/core/blob/main/packages/runtime-core/src/vnode.ts#L609) then I would expect that this example should fail type checks because `string` is not suitable for `in`

instead there is an error at runtime but no type check failure

https://play.vuejs.org/#eNp9kTFPwzAQhf/KcUtAKu3QrUorAaoEDIAAicVLSK7BxbEt2ymRovx3zo4aGKpu9r13776ze7yxdn5oCVeY+9JJG8BTaC2oQtdrgcEL3AgtG2tcgB4c7WCAnTMNZNyWCV0a7QM0voZ1VC+ze1LKwIdxqrrIroTOF2Mwx/AlUGNVEYhvALm3hYbD9afUFQ/jEIGw6fsUNwzcyTo788XUhjNm4pk7Wc/33mgG72OUwNI0VipyzzZIZhK4gqRErWCkn8dUC66l2bFeflH5faK+912sCXxx5MkdSOCkhcLVFEZ5+/ZEHZ8nsTFVq9h9Rnwlb1QbGUfbbasrxv7nS7QP6cmlrt/9tguk/XGpCBqdQ/IL5G+4O7P6H+5yvkx9Qg84/ALz6add

so1ve:
close #3625


dten:
Thanks!

PR #3672, opened 2023-10-16
fix(types): infer discriminated unions in child component props

davidmatter:
This fixes https://github.com/vuejs/core/issues/9381 

Like the fix in vue/core https://github.com/vuejs/core/pull/9336 we use an alternative Omit that doesn't break discriminated unions in props. More information can be found here: https://github.com/microsoft/TypeScript/issues/31501


packages/vue-language-core/src/generators/script.ts
```diff
@@ -377,8 +377,14 @@ export function generate(
 			);
 			//#endregion
 
+			codes.push(
+				`type OmitKeepDiscriminatedUnion<T, K extends keyof any> = T extends any`,
+  			`? Pick<T, Exclude<keyof T, K>>
+  			: never`,
+				`;\n`,
+			);
```
rchl, line 385: I believe this could be defined in https://github.com/vuejs/language-tools/blob/master/packages/vue-language-core/src/utils/globalTypes.ts since it doesn't depend on any dynamic code.

(I'm not super familiar with this project so you might wait for confirmation on that)

davidmatter: you're right, thanks. Done

johnsoncodehk:
LGTM, thanks!

PR #3674, opened 2023-10-17
chore: update vue 2 tests

### LINKED ISSUE #3673: Vue 2.7 strictTemplates error
https://github.com/xiaoxiangmoe/issue-vue-2.7-on-click-type-error.git

```
pnpm run type-check
```

```
src/App.vue:12:6 - error TS2559: Type '{ onClick: any; }' has no properties in common with type 'Readonly<Partial<{}> & Omit<Readonly<ExtractPropTypes<{ border: { type: PropType<boolean>; }; }>>, never>>'.

12     <HelloWorld @click="handleClick" />
        ~~~~~~~~~~


Found 1 error in src/App.vue:12
```

---
HelloWorld.vue
```vue
<script setup lang="ts">

interface ButtonProps {
    border?: boolean; 
}

const props = defineProps<ButtonProps>() 
defineEmits<{
    (event: 'click',payload: MouseEvent): void
}>()
</script>
<template>
    <div id="app">
        <button @click="$emit('click', $event)">Click me</button>
    </div>
</template>
```

so1ve:
fixes vuejs/vue#13104, based on vuejs/language-tools#3351

Actually this is a vue 2.7 type issue: it doesn't convert emits to props. However vue 2.7 seemed to be unmaintained, I fix it in language-tools. Also improves JSX compactibility


rchl:
I'm looking at the changes and I don't quite see how it's related to #3351. Maybe it does handle the "emits" case but I don't see how it would handle class, style, etc.

johnsoncodehk:
Thanks for investigating this, but I think we should avoid bypassing the problem in language tools, otherwise vue repo will never fix it.

so1ve:
Thanks, I'll rewrite this PR to contain some vue2-related test changes.

PR #3689, opened 2023-10-25
fix(class references): experimentalResolveStyleCssClasses regression

### LINKED ISSUE #3688: experimentalResolveStyleCssClasses regression (no class references in template)
Missing classes references in `<template>` when `<style>` block isn't `scoped` but `tsconfig` has 
```ts
'experimentalResolveStyleCssClasses' :  'always'
```

1. without `scoped`
image

2. with `scoped`
image

#### Notes:

Can be reproduced with extension Version: 1.8.20+
Works as expected with 1.8.19 and below.

[Reproduction](https://github.com/maIIady/volar-starter)

maIIady:
Resolves #3688

### DIFF
--- a/packages/language-core/src/plugins/vue-tsx.ts
+++ b/packages/language-core/src/plugins/vue-tsx.ts
@@ -160,7 +160,7 @@ function createTsx(fileName: string, _sfc: Sfc, { vueCompilerOptions, compilerOp
 
 		for (const style of _sfc.styles) {
 			const option = vueCompilerOptions.experimentalResolveStyleCssClasses;
-			if ((option === 'always' || option === 'scoped') && style.scoped) {
+			if (option === 'always' || (option === 'scoped' && style.scoped)) {
 				for (const className of style.classNames) {
 					classes.add(className.text.substring(1));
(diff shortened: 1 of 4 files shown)


packages/language-core/src/plugins/vue-tsx.ts
```diff
@@ -160,7 +160,7 @@ function createTsx(fileName: string, _sfc: Sfc, { vueCompilerOptions, compilerOp
 
 		for (const style of _sfc.styles) {
 			const option = vueCompilerOptions.experimentalResolveStyleCssClasses;
-			if ((option === 'always' || option === 'scoped') && style.scoped) {
+			if (option === 'always' || (option === 'scoped' && style.scoped)) {
```
maIIady, line 163: Just for reference [some lines above the code](https://github.com/vuejs/language-tools/pull/3689/files#diff-35e6594820018d2d5b7d52fc2dd8a2359445d24b2d54d5ae49cf1eaf742156f3R150) was initially correct

johnsoncodehk:
Thanks!

PR #3729, opened 2023-11-12
fix: syntax highlighting for `.prop` shorthand

### LINKED ISSUE #3727: Support for Vue's `.prop` modifier shorthand syntax
image

image

Even though both of the above are [functionally](https://github.com/vuejs/vue/issues/7582) [equivalent](https://vuejs.org/api/built-in-directives.html#v-bind), in VS Code the latter doesn't get syntax-highlighted, it doesn't get IDE autocomplete, and it doesn't get covered by Vue's ESLint plugin.

Note: TypeScript (in Volar takeover mode, haven't tested otherwise) does still check the contents though, at least.

image

so1ve:
fixes #3727

### DIFF
--- a/extensions/vscode/syntaxes/vue.tmLanguage.json
+++ b/extensions/vscode/syntaxes/vue.tmLanguage.json
@@ -953,7 +953,7 @@
 			]
 		},
 		"vue-directives-original": {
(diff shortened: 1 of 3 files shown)


johnsoncodehk:
LGTM, thanks!

PR #3731, opened 2023-11-14
WIP: Update to utility types pr

pikax:
https://github.com/vuejs/core/pull/9556 

Not sure if the snapshots generated are correct 🤔


johnsoncodehk:
I think it's better to keep language-tools and core loosely coupled in types, otherwise language-tools will have to add more version-specific compatibility code, which makes it increasingly difficult to maintain this project.

I will close this PR unless specific issues arise in the future that require relying on more core types.

PR #3743, opened 2023-11-21
fix: incremental insertion incorrect if input at block start

### LINKED ISSUE #3449: adding <script setup> to component with <script> creates error
This is a very minor bug with a simple workaround.
Adding a <script setup> tag (and closing it) to a component that already has a <script> tag will create an 'error' (in red) on the original script tag that says : 
`JSX element 'script' has no corresponding closing tag". ts(17008)`
image
A simple cut and paste on the first two lines will solve the problem. This error only occurs when typing manually the tag <script setup>, not on pasting it.

I'm using the 24th of july version of the extension, and there is no typescript in my project, only JS

so1ve:
fixes #3449


johnsoncodehk:
Thanks for investigating!

PR #3744, opened 2023-11-21
fix: allow data with the same name as components

### LINKED ISSUE #3411: Component names may not match other component properties
## Reproduction

I'm not sure if this is a bug or desired behavior, but it appears component names may not match other properties of that component.

https://play.vuejs.org/#eNp9U8Fu2zAM/RVOhzkFEhtrtkvnFeiKAtsO29D1qItnM45aWTIkOkkR+At22hfsF/cJo+TZSYC2F1skH6lH8mkvrto23XQoLkTuS6daAl2Y+oMU5KW4lEY1rXUEe6hwpQxeW7YNGoIeVs42kHBy8n7ClWulq/+RNLsOVigfENLgLmK4UtHp+D+uONtLA1COpr+A6GBXqDIP5z5+A6QjrE4Bs7PRBHBInTPw5nwJWQb5Amit/JQHrbMtOnoEUzTIbq1VhR62itaMDJ6xyQAYig5Xh29/Jk2eDcPiAbFB2LS6IGQLIH+1WAwpV67umlDFroAeW4Rk3yfARIwlKLxXtSl+agSy0BaObyJ0B2wg/xpusbSuyj05Zeo5dObB2K25TNKx1buXC0+1kpT87Hz59h2zB1gsBq7DujI28uyoCzHn9ZfWrFSd3ntrWB1xtlKE0SiN7ltLyhqWyLQEKQqt7fZL9JHrMA4s5qyxfHjCf+93wSfFd4ce3QalmGJUuBppCN/8+Io7Pk/BxladZvQLwVv0VneB4wD72JmKaR/hItvPUbU82jt/syM0fmwqED0oTgrWcBDqc60f6C7TZcxjnfAUpwcQ3tepTCq1iQeAT8hUhjcT1Rd3ctDg3z+/f8WMbEg53VT/D7oAP8M=

## Expected result

Either the child component should fail to render, or the compiler should throw no errors.

## Actual result

The child component renders successfully, but typescript errors are thrown due to the naming collision.

## Additional context

This appeared to work as expected prior to the release of 1.7 and 1.8.

For now it's easy to side-step by simply choosing a different name, but in large codebases, and specifically in large components, it adds a hurdle to migrating Vue 2 &rarr; 3 because the error messages can be difficult to diagnose. This problem will also start appearing without dependency upgrades in VSCode due to the extension updating automatically.

so1ve:
fixes #3411


johnsoncodehk:
This seems to break component type narrowing and fail the #3138 test of the latest commit: dc2432f4fe89e03209e3a6cdcec1a56a32f9f8f4

so1ve:
🤔 I'll take a look

so1ve:
Failing with another error :(

johnsoncodehk:
After thinking about it, I think this operation should not be done in Vue language tools, let's wait: https://github.com/vuejs/core/pull/3399

Thank you for your hard work!

PR #3750, opened 2023-11-22
fix: don't camelize attributes for plain elements

rchl:
Plain (native) elements should not have their props camelized as then the attribute names won't match the type definition if the element has those. For example `path` element defines type for the `stroke-width` attribute but since Volar "camelizes" the attribute name by default, it triggers type error due to `strokeWidth` not matching the `stroke-width` key.

I suppose the `ElementTypes.ELEMENT` type does not necessarily mean "native element" but hopefully it's close enough and correct to not camelize attributes by default for those elements.

### DIFF
--- a/test-workspace/tsc/vue3_strictTemplate/dataAttributes/main.vue
+++ b/test-workspace/tsc/vue3_strictTemplate/dataAttributes/main.vue
@@ -1,0 +1,16 @@
+<template>
+    <div>
+        <svg><path stroke="#efefef" stroke-width="123" /></svg>
+        <foo stroke-width="123" />
+    </div>
+</template>
+
+<script setup lang="ts">
+import { defineComponent } from 'vue';
+
+const Foo = defineComponent({
+    props: {
+        strokeWidth: { type: String, required: true },
+    }
+});
+</script>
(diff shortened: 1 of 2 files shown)


test-workspace/tsc/vue3_strictTemplate/dataAttributes/main.vue
```diff
@@ -0,0 +1,16 @@
...
+
+<script setup lang="ts">
+import { defineComponent } from 'vue';
+
+const Foo = defineComponent({
+    props: {
+        strokeWidth: { type: String, required: true },
+    }
+});
+</script>
```
so1ve, line 16: Nit: could you please use tabs to indent to match this repo's code style? Thanks

```suggestion
<template>
	<div>
		<svg><path stroke="#efefef" stroke-width="123" /></svg>
		<foo stroke-width="123" />
	</div>
</template>

<script setup lang="ts">
import { defineComponent } from 'vue';

const Foo = defineComponent({
	props: {
		strokeWidth: { type: String, required: true },
	}
});
</script>
```

johnsoncodehk:
I found that the type definition for `SVGAttributes` still has many props that are camelized (allowReorder, attributeName, attributeType, autoReverse...), and the current changes will not recognize those props.

Maybe we need to set a whitelist for the default value of `vueCompilerOptions.htmlAttributes` to solve this problem.

rchl:
> I found that the type definition for `SVGAttributes` still has many props that are camelized (allowReorder, attributeName, attributeType, autoReverse...), and the current changes will not recognize those props.

Seems like that's how the spec intended. Notice that they are listed like that in https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute

So the user should write them literally like that and Vue should not do any transformations.

johnsoncodehk:
@rchl that makes sense, thanks!

Issue #3526, opened 2023-08-31
[Performance]: .vue.d.ts are not used in TypeScript Project References


vidal7:
Hello.

**What I want to do**

I want to better organize my codebase. So I began to introduce TypeScript Project References to help cope with performance problems in vscode and also separating tests from implementation thus it was not possible for the implementation files to import the test files. I was inspired by the official TypeScript Project References documentation on https://www.typescriptlang.org/docs/handbook/project-references.html#an-example-project.

**What is my problem**

However, I keep getting this TypeScript error in vscode and vue-tsc

```
vue/main.unit.ts:1:18 - error TS6307: File 'vue/main.vue.ts' is not listed within the file list of project 'vue/tsconfig.unit.json'. Projects must list all files or use an 'include' pattern.
  The file is in the program because:
    Imported via './main.vue' from file 'vue/main.unit.ts'
    Imported via './main.vue' from file 'vue/main.vue.ts'

1 import Main from './main.vue';
                   ~~~~~~~~~~~~

  vue/main.vue.ts:42:45
    42 let __VLS_internalComponent!: typeof import('./main.vue')['default'];
                                                   ~~~~~~~~~~~~
    File is included via import here.


Found 1 error.
```

![image](https://github.com/vuejs/language-tools/assets/33458053/91cf6932-cef8-4806-926f-ab36f3798f30)

**What I expect to get**
No errors in vscode, npm script build:vue is working the same way as build:ts

**To reproduce**
* Clone https://github.com/vidal7/volar-issue-project-references
* npm install
* npm run build:ts => works
* npm run build:vue => doesn't work

**Details before reproducing**
* build:ts is building a test files project that is referring to a implementation project that contains only a ts file and it works
* build:vue is building a tests files project that is referring to a implementation project that contains only a vue file and it doesn't work.

Is this a bug? Am I missing something?

Thank you.

so1ve:
Add `*.vue` to the `include` pattern in your `tsconfig.unit.json` file

vidal7:
> Add `*.vue` to the `include` pattern in your `tsconfig.unit.json` file

It works but I don't think it is an acceptable solution. My understanding is that it defeats the purpose of separating tests from implementation. Your solution merge the test files with the implementation thus implementation can now depends on test files.

Am I right or I am missing something?

so1ve:
Ugh! Thanks for your clarification. This is likely a bug in volar.

vidal7:
[...]

PR #3751, opened 2023-11-23
perf(tsc): support vue files with project reference delcaration

blake-newman:
Support vue files with project reference output, so that typescript resolves the project reference vue file as the output `vue.d.ts` file rather than the source.

This drastically improves performance in large projects that have vue files imported from project references. As much as 50% performance improvements as ts no longer recompiles vue files from source.

Setup the real path to pass back the real path of vue files, as they are virtual files the inbuilt realpath always returns the non symlinked version (node_modules), which affects DX with vue-tsc as errors originating from upstream project reference packages are returned with the node modules path rather than the source. This change is also required to power the usage of `d.ts` files.

### DIFF
--- a/packages/tsc/src/index.ts
+++ b/packages/tsc/src/index.ts
@@ -64,7 +64,7 @@ export function createProgram(options: ts.CreateProgramOptions) {
 			modifiedTime: number,
 			scriptSnapshot: ts.IScriptSnapshot,
 		}>();
-		const languageHost: vue.TypeScriptLanguageHost = {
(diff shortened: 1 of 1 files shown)


blake-newman:
Solves #3526

blake-newman:
Note there is a similar patch here inside volar:

https://github.com/volarjs/volar.js/pull/93

As volar is meant to be framework agnostic it probably needs to be applied at vue-tsc layer, but i'm now investigating if this impacts vscode and should be resolved in volar.

PR #3766, opened 2023-11-30
fix(component-type-helpers): correct type inference for FunctionalComponent

### LINKED ISSUE #3764: `ComponentSlots<>` doesn't work with generic components (`FunctionalComponent` in general?)
Hello 🖖

## Sample

```vue
<template>
  <div>
    <slot name="child" />
  </div>
</template>

<script lang="ts" setup generic="T">
</script>
```
`ComponentSlots<typeof Sample>` always return `{}`.

## Generics

For generic components (`<script setup lang="ts" generic="...">`) Vue processor generate a `FunctionalComponent`.
And `ComponentSlots<typeof MyGenericComponent>` always returns `{}`.

## Fix

The original implementation of `ComponentSlots` is 
```typescript
export type ComponentSlots<T> =
	T extends new () => { $slots: infer S; } ? NonNullable<S> :
	T extends (props: any, ctx: { slots: infer S; }, ...args: any) => any ? NonNullable<S> :
	{};
```
If I change by adding `attrs` & `emit` fields in context:
```typescript
export type ComponentSlots<T> =
	T extends new () => { $slots: infer S } ? NonNullable<S> :
	T extends (props: any, ctx: { slots: infer S; attrs: any; emit: any }, ...args: any) => any ? NonNullable<S> :
	{}
```
it works

## ComponentEmit

Probably same error with `ComponentEmit`.

pinguet62:
fixes #3764

### DIFF
--- a/packages/component-meta/tests/index.spec.ts
+++ b/packages/component-meta/tests/index.spec.ts
@@ -535,6 +535,19 @@ const worker = (checker: ComponentMetaChecker, withTsconfig: boolean) => describ
 		expect(onBaz?.schema).toEqual([]);
 	});
 
+	test('reference-type-events for generic', () => {
+		const componentPath = path.resolve(__dirname, '../../../test-workspace/component-meta/generic/component.vue');
+		const meta = checker.getComponentMeta(componentPath);
+
+		expect(meta.type).toEqual(TypeMeta.Function);
+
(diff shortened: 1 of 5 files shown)


johnsoncodehk:
Thanks!

PR #3773, opened 2023-12-04
test: update event test case for improved functional events typing in 3.3.10

yyx990803:
This is needed for ecosystem-ci to pass after Vue core 3.3.10
Related: https://github.com/vuejs/core/pull/9234


johnsoncodehk:
I reverted the version bumping. This is to make the version matching more inclusive and prevent users from installing multiple vue versions in node_modules when they only upgrade vue-tsc without upgrading the vue version.

PR #3774, opened 2023-12-04
fix(language-core): camelize props for dynamic component

### LINKED ISSUE #3765: Dynamic component shows error for props with kebab-case
Hi,

since 1.8.24 Volar shows an error when using a dynamic component with an imported component object and a kebab-case required prop.

Example:
`FooBar` component has a required prop `fooBar`. The following leads to an error `Property 'fooBar' is missing in type`.
```
<template>
  <component :is="FooBar" :foo-bar="text"></component>
</template>

<script>
import { FooBar } from '#components'
</script>
```

Best regards
David

so1ve:
fixes #3765


test-workspace/tsc/vue3/#3765/main.vue
```diff
@@ -0,0 +1,15 @@
...
+
+<script setup lang="ts">
+import { defineComponent } from 'vue';
+
+const Foo = defineComponent({
+	props: {
+		strokeWidth: { type: String, required: true },
+	},
+});
+</script>
```
rchl, line 15: I would think that it makes sense to update `test-workspace/tsc/vue3_strictTemplate/dataAttributes/main.vue` so that it includes all cases in one place.

packages/language-core/src/generators/template.ts
```diff
@@ -1224,6 +1224,7 @@ export function generate(
 		codes.push(`}, `);
 
 		for (const prop of props) {
+			const shouldCamelize = !nativeTags.has(node.tag) || node.tag === 'component' || node.tag === 'Component';
```
rchl, line 1227: Feels hacky since the element can be called anything really. It can be a `div` even.

Wouldn't it make sense to force camelize when `node.tagType === CompilerDOM.ElementTypes.COMPONENT`? It appears to be a component type in this case.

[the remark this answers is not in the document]
so1ve: `isCamelizableElement`?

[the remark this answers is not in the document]
rchl: Or `canCamelize`?

packages/language-core/src/generators/template.ts
```diff
@@ -1260,7 +1261,7 @@ export function generate(
 				if (
 					(!prop.arg || (prop.arg.type === CompilerDOM.NodeTypes.SIMPLE_EXPRESSION && prop.arg.isStatic)) // isStatic
 					&& hyphenateAttr(attrNameText) === attrNameText
-					&& !nativeTags.has(node.tag)
+					&& shouldCamelize
```
rchl, line 1264: You can also now move this check to be done before the other ones as it's now the cheapest one (we can avoid `hyphenateAttr`)

packages/language-core/src/generators/template.ts
```diff
@@ -1224,6 +1224,7 @@ export function generate(
 		codes.push(`}, `);
 
 		for (const prop of props) {
+			const shouldCamelize = !nativeTags.has(node.tag) || node.tagType === CompilerDOM.ElementTypes.COMPONENT;
```
rchl, line 1227: In fact this check can be moved outside of the prop loop.

rchl (approved):

rchl:
BTW. My [initial fix](https://github.com/vuejs/language-tools/pull/3750) (the d9351d2959a228e4c8fe24aef85103ff564a6bb9 commit) for those cases used a variant of the `tagType` check. @johnsoncodehk you've changed it to check against the `nativeTags` without really saying why. Could you explain why you had to do that? Because at least based on the tests, the `nativeTags` seems unnecessary and having a hardcoded and potentially outdated list like that feels more hacky than relying on vue compiler itself. While I understand that the Vue compiler might not provide enough information in some cases, at least it's not reflected in the tests (I think) that the `nativeTags` check is necessary.

packages/language-core/src/generators/template.ts
```diff
@@ -1366,8 +1368,8 @@ export function generate(
 				let camelized = false;
 
 				if (
-					hyphenateAttr(prop.name) === prop.name
-					&& !nativeTags.has(node.tag)
+					canCamelize
+					&& hyphenateAttr(prop.name) === prop.name
```
johnsoncodehk, line 1372: Does it need `|| node.tagType === CompilerDOM.ElementTypes.SLOT`? (I'm not sure yet)

johnsoncodehk:
Just had a quick look, I'll have time to test it in a few hours.

packages/language-core/src/generators/template.ts
johnsoncodehk: `nativeTags` is used in the case of using vue-language-tools for projects other than vue, e.g. uni-app projects need to specify `nativeTags` to ignore the original tag names: https://github.com/uni-helper/uni-app-types#%E9%85%8D%E7%BD%AE

Originally #3750 judgment based on `tagType` would break this use case.

I plan to remove `nativeTags` from the separate PR and suggest users to modify template compiler options via plugin api to replace `nativeTags` option, so it should be possible to change to fully `tagType` based judgment.

packages/language-core/src/generators/template.ts
```diff
@@ -1223,6 +1223,8 @@ export function generate(
 		}
 		codes.push(`}, `);
 
+		const canCamelize = !nativeTags.has(node.tag) || node.tagType === CompilerDOM.ElementTypes.COMPONENT;
```
johnsoncodehk, line 1226: I will see if `|| node.tagType === CompilerDOM.ElementTypes.SLOT` is also need or not. (The previous comment in the wrong line has been deleted.)

rchl: Just a remainder that my original fix used `node.tagType !== CompilerDOM.ElementTypes.ELEMENT` which would maybe be even more accurate but I guess it has to be tested.

johnsoncodehk: Okay we don't need that, because the template AST given slot prop name is already camelize. I believe this has always been an edge issue with mapping on the other side and will investigate when I have time.

johnsoncodehk:
Thanks!

PR #3780, opened 2023-12-06
refactor(extension): Optimized the destruction of some events and refactor some code

Simon-He95:


extensions/vscode/src/features/nameCasing.ts
```diff
@@ -11,24 +12,25 @@ export async function activate(_context: vscode.ExtensionContext, client: BaseLa
 
 	await client.start();
 
+	const disposes: Disposable[] = [];
```
so1ve, line 15:
```suggestion
	const disposes: vscode.Disposable[] = [];
```

Simon-He95: Are there any benefits to using `vscode.Disposable` over directly importing types?

so1ve: I think there are no difference but we are already importing the `vscode` namespace: https://github.com/vuejs/language-tools/pull/3780/files#diff-55aa757c11cb962ca102a6bc49b4de9fddae8a4829f056f30a25246d48834311R1

Simon-He95: done

johnsoncodehk:
Thanks!

PR #3859, opened 2024-01-18
fix(language-core): special treatment for number elements inside `v-for`

### LINKED ISSUE #3845: Possible bug in v-for when used with number|number[]
I'm not sure if this is a bug or intended but this code example will give an error when used with `out(f)`:

```vue
<script setup lang="ts">
	import { ref, computed } from 'vue';
	const decide = ref<boolean>(true);
	const foo = computed(() => (decide.value ? 2 : [1]));

	function out(f: number) {
		return f;
	}
</script>

<template>
	<button @click="decide = !decide">Click to switch</button>
	<ul>
		<li v-for="f in foo" :key="f">{{ out(f) }}</li>
	</ul>
</template>
```

Inside the `v-for` I get two conflicting TS hints:

```ts
const f: ((radix?: number | undefined) => string) | {
    (locales?: string | string[] | undefined, options?: Intl.NumberFormatOptions | undefined): string;
    (locales?: Intl.LocalesArgument, options?: Intl.NumberFormatOptions | undefined): string;
} | (() => string) | (() => string)
```

```ts
(property) foo: 2 | number[]
```

Stackbliz example: https://stackblitz.com/edit/vitejs-vite-7atwbn?file=src%2FApp.vue

so1ve:
fixes #3845


johnsoncodehk:
LGTM, thanks!

PR #3882, opened 2024-02-01
fix: address transition type incompatibility with Vue 2.7.16

### LINKED ISSUE #3881: Type issues with Transition and latest Vue 2.x
With latest Vue 2 version (currently `2.7.16`) this code creates a type issue:

```vue
<template>
  <transition>
    <span v-show="true">Foo</span>
  </transition>
</template>
```

> Object literal may only specify known properties, and 'persisted' does not exist in type 'Readonly<Partial<{}> & Omit<Readonly<TransitionProps>, undefined>>'. ​ts(2353)

The issue appears to be due to `@vue/compiler-dom` [adding a `persisted` prop](https://github.com/vuejs/core/blob/d276a4f3e914aaccc291f7b2513e5d978919d0f9/packages/compiler-dom/src/transforms/Transition.ts#L40-L50) to the `transition` node when parsing such construct and Volar adding it to the generated template code which, combined with the latest version of Vue 2 which [includes proper types for the `Transition` component](https://github.com/vuejs/vue/issues/13002), causes this issue. Previously the type of Transition component was something less strict or `unknown` so there was no issue.

(there is also an issue with using a `key` attribute but that is maybe an issue with Vue types)

rchl:
Explained all in #3881

Fixes #3881

### DIFF
--- a/packages/language-core/src/generators/template.ts
+++ b/packages/language-core/src/generators/template.ts
@@ -1365,6 +1365,11 @@ export function generate(
 					continue;
 				}
 
+				if (vueCompilerOptions.target < 3 && node.tag === 'transition' && prop.name === 'persisted') {
+					// Vue 2 Transition doesn't support "persisted" property but `@vue/compiler-dom always adds it (#3881)
+					continue;
+				}
+
 				let camelized = false;
 
 				if (
--- a/packages/language-core/src/generators/template.ts
+++ b/packages/language-core/src/generators/template.ts
@@ -1365,7 +1365,11 @@ export function generate(
 					continue;
 				}
 
-				if (vueCompilerOptions.target < 3 && node.tag === 'transition' && prop.name === 'persisted') {
+				if (
+					vueCompilerOptions.target < 3
+					&& (node.tag === 'transition' || node.tag === 'Transition')
+					&& prop.name === 'persisted'
+				) {
 					// Vue 2 Transition doesn't support "persisted" property but `@vue/compiler-dom always adds it (#3881)
 					continue;
 				}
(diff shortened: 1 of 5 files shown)


packages/language-core/src/generators/template.ts
```diff
@@ -1365,6 +1365,11 @@ export function generate(
 					continue;
 				}
 
+				if (vueCompilerOptions.target < 3 && node.tag === 'transition' && prop.name === 'persisted') {
```
so1ve, line 1368: What will happen the tag is `Transition`?

rchl: Good point. I thought it's normalized by the compiler.

rchl:
Note that for some reason the error on this test does not reproduce when opening that file in VSCode and I totally don't get why. Seems like generated template doesn't include the `persisted` property there.

But regardless of that, it's a real issue that shows up with vue-tsc and also with LSP-volar in Sublime Text.

rchl:
> > Note that for some reason the error on this test does not reproduce when opening that file in VSCode and I totally don't get why. Seems like generated template doesn't include the `persisted` property there.
> > But regardless of that, it's a real issue that shows up with vue-tsc and also with LSP-volar in Sublime Text.
> 
> I have a feeling that the aggressive treeshaking might be accidentally removing code responsible for that in the vscode extension. That would be kinda bad if true.

Oh, it looks like different bundles of `@vue/compiler-dom` behave differently depending on the the `NODE_ENV` var. One doesn't include `transformTransition` (which is responsible for this behavior) in production and that's probably the one used when bundling code for the vscode extension.

Screenshot 2024-02-01 at 23 04 40

johnsoncodehk:
LGTM, thanks!

PR #3908, opened 2024-02-29
feat(language-service): migrate to named pipe server using TypeScript LanguageService

johnsoncodehk:
After migrating to the TS Plugin, the LSP Server no longer provides the semantic functionality of the TypeScript LanguageService instance. It needs to communicate with tsserver to invoke its LanguageService instance.

The communication between LSP Server and tsserver is based on named pipes instead of websockets. This is to avoid occupying local ports and dependencies on the ws package.

The following features have been migrated to the named pipe server that uses the TypeScript LanguageService:

- [x] Component Extract CodeAction
- [x] Template Twoslash Queries
- [x] Auto Insert `.value`
- [ ] Tag and Props Name Casing Detect / Convert
- [ ] Detect Missing `lang="ts"` / `"allowJs": true` / `jsconfig.json`
- [ ] Missing Required Props InlayHints
- [ ] Tag and Props Completion in Template

### DIFF
--- a/packages/language-service/src/plugins/vue-twoslash-queries.ts
+++ b/packages/language-service/src/plugins/vue-twoslash-queries.ts
@@ -1,7 +1,7 @@
 import { ServicePlugin, ServicePluginInstance } from '@volar/language-service';
 import * as vue from '@vue/language-core';
-import { Provide } from 'volar-service-typescript';
 import type * as vscode from 'vscode-languageserver-protocol';
+import { sendGetQuickInfoAtPosition } from 'typescript-vue-plugin/out/requests/client';
 
 const twoslashReg = /<!--\s*\^\?\s*-->/g;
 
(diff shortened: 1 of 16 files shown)


packages/language-service/src/plugins/vue-twoslash-queries.ts
```diff
@@ -1,7 +1,7 @@
 import { ServicePlugin, ServicePluginInstance } from '@volar/language-service';
 import * as vue from '@vue/language-core';
-import { Provide } from 'volar-service-typescript';
 import type * as vscode from 'vscode-languageserver-protocol';
+import { getQuickInfoAtPosition } from 'typescript-vue-plugin/out/namedPipe/client';
```
rchl, line 4: Should it rely on internal package build paths?

packages/typescript-plugin/src/namedPipe/server.ts
```diff
@@ -0,0 +1,62 @@
...
+export type Request = CollectExtractPropsRequest
+	| GetPropertiesAtLocationRequest
+	| GetQuickInfoAtPosition;
+
+let started = false;
+
+export function startNamedPipeServer() {
+	if (started) return;
+	started = true;
+	const server = net.createServer(connection => {
```
rchl, line 35: Is it not possible to communicate using node-ipc for example? At least that's what typescript-language-server is using to communicate with tsserver but not sure if that's relevant for this case...

johnsoncodehk: In my memory IPC communication requires obtaining a ChildProcess instance? Neither the LSP server nor the language client can obtain a process instance of tsserver so this may not be possible.

packages/language-service/src/plugins/vue-twoslash-queries.ts
johnsoncodehk: Not sure I understand. Just need to make sure that both node and esbuild can interpret the path.

rchl: just saying that `/out/namedPipe/client` path is an internal detail of the `typescript-vue-plugin` package and should not be hardcoded like that. But since you control both, I guess you can get away with it :)

PR #3914, opened 2024-03-01
fix(language-service): wait for tsserver to be ready when requesting auto insert `.value`

johnsoncodehk:
The call to `provideAutoInsertionEdit` is too eager (called 100ms after the document is changed). At this time, the tsserver may not have obtained the latest script changes, and the requested position parameter may be calculated at the wrong position.

Even if an additional version is used to check whether the tsserver script has been synchronized, it does not mean that the type checker is currently available, because the tsserver must wait for the diagnosis to be completed to synchronize the internal state. If the TS API is called in advance in the request handle, it will also cause the tsserver status to be abnormal.

It is currently difficult to find a way to reliably handle tsserver state synchronization, so this PR simply waits 250ms to bypass the problem.

### DIFF
--- a/packages/language-service/lib/plugins/vue-autoinsert-dotvalue.ts
+++ b/packages/language-service/lib/plugins/vue-autoinsert-dotvalue.ts
@@ -10,6 +10,7 @@ export function create(ts: typeof import('typescript')): ServicePlugin {
 	return {
 		name: 'vue-autoinsert-dotvalue',
 		create(context): ServicePluginInstance {
+			let currentReq = 0;
 			return {
 				async provideAutoInsertionEdit(document, position, lastChange) {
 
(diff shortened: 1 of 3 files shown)


packages/language-service/lib/plugins/vue-autoinsert-dotvalue.ts
```diff
@@ -19,6 +20,12 @@ export function create(ts: typeof import('typescript')): ServicePlugin {
...
 
+					const req = ++currentReq;
+					// Wait for tsserver to sync
+					await sleep(250);
+					if (req !== currentReq)
+						return;
+
 					const enabled = await context.env.getConfiguration?.<boolean>('vue.autoInsert.dotValue') ?? true;
 					if (!enabled)
 						return;
```
rchl, line 31: Wouldn't it make sense to check this configuration before the newly added code?

johnsoncodehk: getConfiguration has a send request cost before Volar has cached the option, so it is best to place it after sleep.

rchl: But now the unnecessary sleep will trigger for every document change, even if one is not using this option. Seems more wasteful than getting a configuration once that gets cached later.

PR #3950, opened 2024-03-03
Add LSP package name to the readme

WhyNotHugo:
The README contains no direct reference to the language server itself. Given past  renames and the legacy Vetur, it's not straightforward to fine the correct package online.

Include the LSP and its package name in the README. This will assist people trying to install the language server.

Additionally, as of 2.0.0, a plug-in is required for TypeScript support. This is mentioned in the changelog, but the name of the plug-in itself is not made evident anywhere. Also add this to the README.


Stanzilla:
I think you got that backwards? As of 2.0, the plugin is no longer required?

PR #3994, opened 2024-03-05
doc(component-meta): update demo

zzfn:
- replace createComponentMetaChecker with createChecker 
- replace src with lib


johnsoncodehk:
Thanks!

PR #4038, opened 2024-03-07
fix: allow folding root elements

### LINKED ISSUE #4007: There are so many questions that I don’t know where to start.
Including but not limited to the following issues：

Tags such as <script> and <template> cannot be folded normally
image
image
No folding sign

H5 tag prompt is too slow. When a function name is declared in <template> but the function is not declared in <script>, no error message appears.
Error messages are delayed. The vue file needs to be reloaded to prompt normal errors.

The component cannot be recognized, and the vue file needs to be reloaded to solve the problem.

Vue - Official: v2.0.5
VsCode: February 2024 (version 1.87)

### LINKED ISSUE #4023: Unable to fold/unfold top-level blocks in *.vue files
For some reason, I am no longer able to fold or unfold top-level blocks (template, script, and style) in *.vue files.

This issue started occurring recently and I have not been able to identify any changes that could have triggered it. I have tried updating to the latest version of the extension and resetting the extension settings to their default values, but the issue persists.

Here are some details about my setup:

- VS Code version: 1.87.0
- Vue - official extension version: **2.0.5**
- Operating System: Windows 11

Any help in resolving this issue would be greatly appreciated. 
Bug folding Volar

so1ve:
fix #4023, fix #4007

### DIFF
--- a/packages/language-core/lib/virtualFile/computedMappings.ts
+++ b/packages/language-core/lib/virtualFile/computedMappings.ts
@@ -1,7 +1,7 @@
-import { Mapping, Segment, replaceSourceRange } from '@volar/language-core';
+import { CodeMapping, Segment, replaceSourceRange } from '@volar/language-core';
 import { computed } from 'computeds';
 import type * as ts from 'typescript';
-import { enableAllFeatures } from '../generators/utils';
+import { disableAllFeatures, enableAllFeatures } from '../generators/utils';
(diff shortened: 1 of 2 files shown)


johnsoncodehk:
Thanks! But the original fix caused the end tag to be invisible, and we just fixed the mapping to solve it.

PR #4049, opened 2024-03-08
Basic attribute inheritance

A5rocks:
This PR adds [attribute inheritance](https://vuejs.org/guide/components/attrs.html#attribute-inheritance) irrelevant of the `inheritAttrs` configuration option (for now). This is in pursuit of making `strictTemplates` more useful.

I'm not sure how to check `inheritAttrs` as a follow up: I suppose probably `scriptSetupRanges`/`scriptRanges`?

---

Two things I noticed (not sure if these are already tracked or if I just am incorrect about these):
 - `@vue-expect-error` doesn't seem to error if the next statement doesn't error
 - components as a direct child of `<template>` just don't seem to get typed? (this must be expected behavior that I just don't get given that I don't use vue much...)


A5rocks:
Sorry, I should have run all the tests locally, not just the ones with `strictTemplate`. Some of the snapshot diffs look bad.

A5rocks:
I don't understand some of the test failures, because it looks like e.g. `Child` is just not being picked up as a type? Which is strange cause I don't see what I did that could cause that.

... There has to be a better way to approach this. This approach sucks.

so1ve:
Are you still working on this?

A5rocks:
Yeah I have scenario 1 and 2, I'm not planning on ever supporting scenario 3, and I'm not sure how to approach scenario 4. (I'm not sure how to reach in and get the value of `inheritAttrs`...)

But I'm running into roadblocks for the more trivial task of "always inherit attrs from the child, if there's only 1" as seen by failing test cases here... I tried fixing one thing (the latest commit) and it both breaks component typing in some cases and also breaks what initially worked w/r/t inheriting attributes.

Do you think there's a better way to approach implementing this?

so1ve:
I can give it a shot, but I prefer to work in a separate PR to avoid git reset. I can add you as a co-author. WDYT? :eyes:

A5rocks:
Sure, that would work! I only started this cause I made some dumb mistakes that would have been caught with `strictTemplates`... which I can't rely on due to false positives.

PR #4075, opened 2024-03-12
fix(typescript-plugin): fault tolerance for named pipe servers json file

Simon-He95:
After the upgrade, I suddenly found that the syntax in the `template` was invalid, and then tracked the problem and found that there was a problem reading the file in `searchNamedPipeServerForFile`. The format may have been broken when writing a certain version, resulting in JSON here. When `JSON.parse`, an error is reported directly, but no interception is done, and then there is no code prompt. I don’t know if I should delete the temp file directly when there is a problem with `parse`. At present, I have added `try {} catch (e) {}` to solve the problem.

The following is the format of the file that I found during troubleshooting.
image


johnsoncodehk:
Thanks! This may be caused by multiple tsserver processes overwriting json files at the same time.

PR #4081, opened 2024-03-12
chore(vscode): remove embedded language IDs

remcohaszing:
The language ID isn’t needed for injections. Only `scopeName` is needed. The language ID appears in the VSCode language selector, which isn’t desired behaviour.

VSCode language selector showing the various Vue related language IDs

This is based on https://github.com/microsoft/vscode/issues/207406#issuecomment-1991393649. Also I didn’t notice problems when trying this myself.


johnsoncodehk:
LGTM, thanks!

PR #4084, opened 2024-03-12
chore: add issue template

gegoune:
Hi @johnsoncodehk,

I have been astounded with quality of issue reports here following 2.0 release. I believe having issue template will help a bit by requiring users to put effort into high quality issues with replication steps and all other required information so maintainers and contributors don't have to do extra legwork.

This PR creates example issue template coming from [documentation](https://docs.github.com/en/communities/using-templates-to-encourage-useful-issues-and-pull-requests/configuring-issue-templates-for-your-repository#creating-issue-forms) and is intended to check whether you would be interested in implementing it at all. If so we can change the template to fit this repositories' requirements.

### DIFF
--- a/.github/ISSUE_TEMPLATE
+++ b/.github/ISSUE_TEMPLATE
@@ -1,0 +1,64 @@
+name: Bug Report
+description: File a bug report
+title: "[Bug]: "
+labels: ["bug", "triage"]
+projects: ["octo-org/1", "octo-org/44"]
+assignees:
+  - octocat
+body:
+  - type: markdown
+    attributes:
+      value: |
+        Thanks for taking the time to fill out this bug report!
+  - type: input
+    id: contact
+    attributes:
+      label: Contact Details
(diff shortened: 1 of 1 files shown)


rchl:
> https://twitter.com/johnsoncodehk/status/1649825147753488384

I don't get that point. A consistent structure to the reports would help to process the reports. The report is not meant to be a place to express myself but provide concise information that help with fixing the issue.

Also having a codesandbox (or whatever is cool these days) that users could fork would simplify reporting.

I would consider myself a more advanced user and would often provide reproduction (or even a fix) but I would still prefer to follow a template when reporting than having to think about how to write it myself.

lisonge:
maybe the file name should be `.github/ISSUE_TEMPLATE/bug_report.yml` instead of `.github/ISSUE_TEMPLATE`

and more information can be configured in `.github/ISSUE_TEMPLATE/config.yml`

johnsoncodehk:
We have added the issue template now, thank you!

PR #4093, opened 2024-03-13
fix(component-meta): `defineSlots` allows empty params

### LINKED ISSUE #4091: [BUG] vue-component-meta does not work correctly when using slots without defined props
Hi there!

I found a bug when using the `vue-component-meta` package inside Storybook. The autogenerated documentation inside storybook will break when there are slots without explicit props defined. 

See https://stackblitz.com/edit/github-admhcw?file=src%2Fstories%2FButton2.vue for a reproduction.

Works:
```ts
defineSlots<{
  /**
   * Use for custom content, instead of relying on 'icon' and 'label' props
   */
  default: (props: {}) => VNode[];
  /**
   * Override the default QSpinner when in 'loading' state
   */
  loading: (props: {}) => VNode[];
}>();
```

Does not work:
```ts
defineSlots<{
  /**
   * Use for custom content, instead of relying on 'icon' and 'label' props
   */
  default: () => VNode[];
  /**
   * Override the default QSpinner when in 'loading' state
   */
  loading: () => VNode[];
}>();
```

Here is the related issue: https://github.com/storybookjs/storybook/issues/26465

Evertvdw:
When using defineSlots without a function argument it would throw an error as `paramType` can be undefined in that case. By adding this check this can be prevented.

I used a `getAnyType` now in case of no types, but I'm not sure if that is the best approach. The equivelent using a template-slot will generate an empty object as type. I'm not sure why and how to replicate that in this code, feel free to improve.


Evertvdw:
@johnsoncodehk There are build errors introduced by the auto-fix workflow that you added. A heads up as you might want to know.

johnsoncodehk:
LGTM, thanks!

PR #4119, opened 2024-03-19
feat: reintroducing the complete language server

### LINKED ISSUE #3942: False positive errors at end-of-file
TS plugin cannot be used with Vue files if the project does not contain tsconfig/jsconfig, or the Vue files are not included by tsconfig/jsconfig.

### Steps to reproduce

1. Clone https://github.com/johnsoncodehk/volar-starter
	Since it is just for demonstration purposes, node_modules will not be installed.
2. Delete the `tsconfig.json` file in repo
3. Open the repo in VSCode, and open `src/components/HelloWorld.vue`
4. You should see TS errors reported due to missing dependencies, as shown below:
	image
5. Close `HelloWorld.vue`, open `src/main.ts`, then re-open `HelloWorld.vue`
6. Errors in the image above is now displayed in the wrong location (end of file)
	image

### Exploration

`@vue/typescript-plugin` calls `decorateLanguageService` and `decorateLanguageServiceHost` at following code to modify the original `languageService` and `languageServiceHost`:
https://github.com/vuejs/language-tools/blob/3efa52d3a1132c8a098bb35c7bd54b81ee93ee35/packages/typescript-plugin/index.ts#L62-L63

This works well for `ConfiguredProject`. But for `InferredProject`, if you open the TS file before opening the Vue file, the modification of `decorateLanguageService` is no longer valid, just like the original `languageService` is no longer used. Therefore the mapping behavior of `decorateLanguageService` is no longer valid.

### Temporary Solutions

1. Create a `jsconfig.json` file in project root with empty json object content `{}`
2. Downgrade to 1.8.27 for now

**Update:** If you have the "Svelte for VS Code" extension installed, disabling it may avoid this problem.

### LINKED ISSUE #3962: Vue server still crashed after upgrading to v2.0.4 in VSCode + Nuxt 3
Sorry to open a duplicated issue, I want to provide a reproduction here.

#### Environment

Apple M2
macOS Sonoma 14.2.1
Node v20.9.0
VSCode v1.87.0
pnpm v8.2.0
nuxt v3.2.2
Vue.volar v2.0.4

```
git clone https://github.com/Bernankez/blog.git
cd blog
pnpm install
```

After opening a Vue SFC, it alerts
image

#### outputs

```
null
null
null
undefined:1


SyntaxError: Unexpected end of JSON input
    at JSON.parse (<anonymous>)
    at Socket.<anonymous> (/Users/cole/.vscode/extensions/vue.volar-2.0.4/dist/server.js:10024:9777)
    at Socket.emit (node:events:526:35)
    at endReadableNT (node:internal/streams/readable:1359:12)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21)

Node.js v18.17.1
[Error - 14:27:08] Server process exited with code 1.
[Info  - 14:27:08] Connection to server got closed. Server will restart.
true
null
null
null
undefined:1
```

johnsoncodehk:
closes #3942, closes #3962

We migrated to the TS plugin in 2.0 to resolve many issues with v1, unfortunately it had many issues with compatibility:

- VSCode <= 1.87 bundled Node version breaks tsserver + `@vue/typescript-plugin` in some environments.
- Even if `@vue/typescript-plugin` only provides vue virtual code and handles mapping, it still causes performance problems in tsserver in some TS versions.
- Whenever any TS plugin crashes in a .vue file, it will cause `@vue/typescript-plugin` to be invalidated and make intellisense completely wrong. This is a fault-tolerance problem of the TS LS plugin architecture.

These issues are beyond our control, so we reintroduced the v1 complete LSP server and only switched to Hybrid Mode when explicitly enabled by the user.


altrusl:
@johnsoncodehk 

That was hard time
Thank you for your work

Stanzilla:
So what's the recommended setup for TypeScript users now? Do we need the extra VSC extension again?

so1ve:
@Stanzilla No, you just need the `Vue - Official` plugin.

Stanzilla:
> @Stanzilla No, you only need the `Vue - Official` plugin.

And Hybrid Mode enabled, right? Would make more sense if this was called something like enable TypeScript support maybe?

so1ve:
> And Hybrid Mode enabled, right? Would make more sense if this was called something like enable TypeScript support maybe?

If you have TS 5.3+ installed and have a `tsconfig.json` or `jsconfig.json`, feel free to enable it.

> if this was called something like enable TypeScript support maybe

Not exact. "Hybrid Mode" means Vue Language Tools makes the built-in TS service to run on `.vue` files, and if you disable hybrid mode we a two separate TS language server on `.vue` files.

gegoune:
Would be very helpful to have a comparison table of two solutions/approachers. Not being much experienced in tooling for JS I struggle to see how to chose what to do, enable hybrid mode? Disable it? What are the benefits of each approach? Downsides?

Stanzilla:
> Would be very helpful to have a comparison table of two solutions/approaches. Not being much experienced in tooling for JS I struggle to see how to chose what to do, enable hybrid mode? Disable it? What are the benefits of each approach? Downsides?

Yes exactly, it's very unclear atm

vidal7:
Maybe this can help : https://gist.github.com/johnsoncodehk/62580d04cb86e576e0e8d6bf1cb44e73

PR #4134, opened 2024-03-21
docs: update readme with neovim lspconfig setup

RayGuo-ergou:
### Description

The length of this issue continues to increase. I believe that posting the configuration directly in the README would be more straightforward for anyone seeking help.

### DIFF
--- a/README.md
+++ b/README.md
@@ -47,9 +47,9 @@ lspconfig.tsserver.setup {
   init_options = {
     plugins = {
       {
-	name = '@vue/typescript-plugin',
-	location = ts_plugin_path,
-	languages = { 'vue' },
+        name = '@vue/typescript-plugin',
+        location = ts_plugin_path,
+        languages = { 'vue' },
       },
     },
   },
--- a/README.md
+++ b/README.md
@@ -79,6 +79,7 @@ lspconfig.volar.setup {
     vue = {
       -- `hybridMode` provides full typescript functionality.
       -- If `hybridMode` is set to `true` `Volar` will run embedded `tsserver` therefore there is no need to run it separately.
+      -- `hybridMode` default to be `true` if not set.
       --- @see https://github.com/vuejs/language-tools/pull/4119
       hybridMode = true,
     },
--- a/README.md
+++ b/README.md
@@ -35,62 +35,48 @@
 Since version `2.0.0`, take over mode has been removed. You have to run `@vue/language-server` alongside a TypeScript server that is running `@vue/typescript-plugin`. Here is a minimal configuration for Neovim's LSP to make the language server work after upgrading to version `2.0.0`.
 
   ```lua
-local mason_registry = require('mason-registry')
-local ts_plugin_path = mason_registry.get_package('vue-language-server'):get_install_path() .. '/node_modules/@vue/language-server/node_modules/@vue/typescript-plugin'
+-- If you are using mason.nvim, you can get the ts_plugin_path like this
+-- local mason_registry = require('mason-registry')
+-- local ts_plugin_path = mason_registry.get_package('vue-language-server'):get_install_path() .. '/node_modules/@vue/language-server/node_modules/@vue/typescript-plugin'
 
-local servers = {
-  tsserver = {
-    init_options = {
-      plugins = {
-        {
-          name = '@vue/typescript-plugin',
-          location = ts_plugin_path,
-          -- If .vue file cannot be recognized in either js or ts file try to add `typescript` and `javascript` in languages table.
-          languages = { 'vue' },
-        },
+local ts_plugin_path = '/path/to/@vue/typescript-plugin'
+
+local lspconfig = require('lspconfig')
+
+lspconfig.tsserver.setup {
+  init_options = {
+    plugins = {
+      {
+	name = '@vue/typescript-plugin',
+	location = ts_plugin_path,
+	languages = { 'vue' },
       },
     },
-    filetypes = { 'typescript', 'javascript', 'javascriptreact', 'typescriptreact', 'vue' },
   },
-  volar = {},
+  filetypes = { 'typescript', 'javascript', 'javascriptreact', 'typescriptreact', 'vue' },
 }
 
-local mason_lspconfig = require('mason-lspconfig')
+lspconfig.volar.setup {}
 
 -- Auto cmd (LspAttach) to setup keybind, codelens, and formatting stuff
 -- I assume everyone should have this configured already but just for reference
 -- @see https://github.com/nvim-lua/kickstart.nvim/blob/65a5ac404b56c4718d79f65ac642e19e89346eda/init.lua#L451-L522
-Util.lsp.lsp_autocmd()
-
-local capabilities = vim.lsp.protocol.make_client_capabilities()
--- If you need cmp
-capabilities = vim.tbl_deep_extend('force', capabilities, require('cmp_nvim_lsp').default_capabilities())
-mason_lspconfig.setup({
-  ensure_installed = vim.tbl_keys(servers),
-  handlers = {
-    function(server_name)
-      local server = servers[server_name] or {}
-      server.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
-      require('lspconfig')[server_name].setup(server)
-    end,
-  },
-})
   ```
-
-However, after version `2.0.7`, you can explicitly disable hybrid mode therefore there's two more options to config vue-language-server.
+configuration for 2.0.7 or newer. (You can still use configuration above, `2.0.7` gives you more options)
 
 *Make sure you have typescript installed globally or pass the location to volar*
 
 ```lua
 -- If you would like to keep only vue-language-server for all vue, typescript and javascript files you can have the configuration simliar with take over mode
-local servers = {
-  -- tsserver = {},
-  volar = {
-    filetypes = { 'typescript', 'javascript', 'javascriptreact', 'typescriptreact', 'vue' },
-    init_options = {
-      vue = {
-        hybridMode = false,
-      },
+local lspconfig = require('lspconfig')
+
+-- lspconfig.tsserver.setup {}
+lspconfig.volar.setup {
+  filetypes = { 'typescript', 'javascript', 'javascriptreact', 'typescriptreact', 'vue' },
+  init_options = {
+    vue = {
+      -- Turn the hybrid mode on so you do not need to run tsserver
+      hybridMode = true,
     },
   },
 }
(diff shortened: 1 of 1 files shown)


gegoune:
I think best would be to avoid things like mason, not every one uses it and those who do know how to adopt configs to it.

There could be an argument made that even lspconfig is something that could be avoided here, but its adoption is so widespread that it's harder to justify it.

Thanks for that PR though!

README.md
```diff
@@ -27,8 +27,100 @@
...
+local ts_plugin_path = mason_registry.get_package('vue-language-server'):get_install_path() .. '/node_modules/@vue/language-server/node_modules/@vue/typescript-plugin'
+
+local servers = {
+  tsserver = {
+    init_options = {
+      plugins = {
+        {
+          name = '@vue/typescript-plugin',
+          location = ts_plugin_path,
+          -- If .vue file cannot be recognized in either js or ts file try to add `typescript` and `javascript` in languages table.
```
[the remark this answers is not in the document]
RayGuo-ergou: thanks for pointing that out that's a old comment from take over mode. I will remove it.

README.md
```diff
@@ -27,8 +27,100 @@
...
+})
+  ```
+
+However, after version `2.0.7`, you can explicitly disable hybrid mode therefore there's two more options to config vue-language-server.
+
+*Make sure you have typescript installed globally or pass the location to volar*
+
+```lua
+-- If you would like to keep only vue-language-server for all vue, typescript and javascript files you can have the configuration simliar with take over mode
+local servers = {
```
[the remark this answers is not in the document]
RayGuo-ergou: This section is under `neovim/nvim-lspconfig`. So I think I will use `lspconfig[server].setup` and remove mason. Thanks

README.md
```diff
@@ -27,8 +27,100 @@
...
+          location = ts_plugin_path,
+          languages = { 'vue' },
+        },
+      },
+    },
+  },
+  volar = {
+    init_options = {
+      vue = {
+        hybridMode = false,
```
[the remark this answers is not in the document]
RayGuo-ergou: ops my bad

README.md
```diff
@@ -27,8 +27,100 @@
...
+
+<details>
+  <summary>How to configure vue language server with neovim and lsp?</summary>
+
+Since version `2.0.0`, take over mode has been removed. You have to run `@vue/language-server` alongside a TypeScript server that is running `@vue/typescript-plugin`. Here is a minimal configuration for Neovim's LSP to make the language server work after upgrading to version `2.0.0`.
+
+  ```lua
+local mason_registry = require('mason-registry')
+local ts_plugin_path = mason_registry.get_package('vue-language-server'):get_install_path() .. '/node_modules/@vue/language-server/node_modules/@vue/typescript-plugin'
+
```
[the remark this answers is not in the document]
RayGuo-ergou: I would like to keep this part but as comment. This could save someone's few mins to figure out the path.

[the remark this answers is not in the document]
gegoune: It';s fine to keep the path variable, but maybe without mason?

README.md
```diff
@@ -35,89 +35,75 @@
...
-      local server = servers[server_name] or {}
-      server.capabilities = vim.tbl_deep_extend('force', {}, capabilities, server.capabilities or {})
-      require('lspconfig')[server_name].setup(server)
-    end,
-  },
-})
   ```
-
-However, after version `2.0.7`, you can explicitly disable hybrid mode therefore there's two more options to config vue-language-server.
+configuration for 2.0.7 or newer. (You can still use configuration above, `2.0.7` gives you more options)
```
gegoune, line 65: I really think having a single configuration for latest version is the best option. It will only get older and older with barely anyone not using latest version.

Also, what other options? :)

README.md
```diff
@@ -35,89 +35,75 @@
...
-        hybridMode = false,
-      },
+local lspconfig = require('lspconfig')
+
+-- lspconfig.tsserver.setup {}
+lspconfig.volar.setup {
+  filetypes = { 'typescript', 'javascript', 'javascriptreact', 'typescriptreact', 'vue' },
+  init_options = {
+    vue = {
+      -- Turn the hybrid mode on so you do not need to run tsserver
```
gegoune, line 78:
```suggestion
-- `hybridMode` provides same functionality as `tsserver` with `@vue/typescript-plugin` plugin.
-- If `hybridMode` is set to `true` `Volar` will run embedded `tsserver` therefore there is no need to run it separately.
```

Is that correct? If so, maybe description like that could be more useful?

gegoune (changes requested):

README.md
gegoune: Or even state something like that and say that there are two ways of running `Volar`, with and without additional `tsserver` and then provide two configurations just like you are doing here.

README.md
```diff
@@ -27,8 +27,86 @@
...
 
 [neovim/nvim-lspconfig](https://github.com/neovim/nvim-lspconfig) ⚡ 🤝 \
-*Vue language server configuration for Neovim* \
-[[Volar 2.0 version set up tutorial](https://github.com/vuejs/language-tools/issues/3925)]
+*Vue language server configuration for Neovim*
+
+<details>
+  <summary>How to configure vue language server with neovim and lsp?</summary>
+
+Since version `2.0.0`, take over mode has been removed. You have to run `@vue/language-server` alongside a TypeScript server that is running `@vue/typescript-plugin`. Here is a minimal configuration for Neovim's LSP to make the language server work after upgrading to version `2.0.0`.
```
gegoune, line 35:
```suggestion
Configuration for versions:  `2.0.0` to `2.0.6`

Note: take over mode has been removed. You have to run `@vue/language-server` alongside a TypeScript server that is running `@vue/typescript-plugin`. Here is a minimal configuration for Neovim's LSP to make the language server work after upgrading to version `2.0.0`.
```

README.md
```diff
@@ -27,8 +27,86 @@
...
+  filetypes = { 'typescript', 'javascript', 'javascriptreact', 'typescriptreact', 'vue' },
+}
+
+lspconfig.volar.setup {}
+
+-- Auto cmd (LspAttach) to setup keybind, codelens, and formatting stuff
+-- I assume everyone should have this configured already but just for reference
+-- @see https://github.com/nvim-lua/kickstart.nvim/blob/65a5ac404b56c4718d79f65ac642e19e89346eda/init.lua#L451-L522
+  ```
+configuration for 2.0.7 or newer. (You can still use configuration above, `2.0.7` gives you more options)
```
gegoune, line 65:
```suggestion
Configuration for versions: `2.0.7` or newer.
```

gegoune (changes requested):

README.md
RayGuo-ergou: I would not say the first config is only for `2.0.0` to `2.0.6` it still work in newest version and is one of biggest reason why version 2 exist. I probably state the config for version 2 and newer

README.md
RayGuo-ergou: Rather saying the config is for `2.0.7` state it as "hybrid mode config" would be more accurate.

RayGuo-ergou:
Thanks @gegoune for your recommendations. 💖

fengT-T:
@RayGuo-ergou but in my nvim config, if i set hybridMode = false, i dont need run tsserver

RayGuo-ergou:
> but in my nvim config, if i set hybridMode = false, i dont need run tsserver

Shoot! My bad, it should be another way around, thanks so much!

gegoune:
> Thanks @gegoune for your recommendations. 💖

Very happy to be of help! Thank you for taking initiative here! 🦸 

You can ping me when you think it's ready and I can take another look.

README.md
```diff
@@ -27,8 +27,91 @@
...
+local ts_plugin_path = '/path/to/@vue/typescript-plugin'
+
+local lspconfig = require('lspconfig')
+
+lspconfig.tsserver.setup {
+  init_options = {
+    plugins = {
+      {
+        name = '@vue/typescript-plugin',
+        location = ts_plugin_path,
```
johnsoncodehk, line 53: As I said at https://github.com/vuejs/language-tools/issues/3925#issuecomment-1987733530, specifying location as @vue/language-server path should be able to directly reference @vue/language-server on which it depends. @vue/typescript-plugin, users do not need to install @vue/typescript-plugin globally and can ensure that it is the same version.

Does the documentation not suggest this because it doesn't work for you?

RayGuo-ergou: Ops missed that, yes just tried and it worded. Thanks!

README.md
```diff
@@ -27,8 +27,91 @@
...
+  filetypes = { 'typescript', 'javascript', 'javascriptreact', 'typescriptreact', 'vue' },
+}
+
+lspconfig.volar.setup {}
+
+-- Auto cmd (LspAttach) to setup keybind, codelens, and formatting stuff
+-- I assume everyone should have this configured already but just for reference
+-- @see https://github.com/nvim-lua/kickstart.nvim/blob/65a5ac404b56c4718d79f65ac642e19e89346eda/init.lua#L451-L522
+  ```
+### Hybrid mode(similar to takeover mode) configuration (Requires `@vue/language-server` version `^2.0.7`)
```
johnsoncodehk, line 67: Here says the opposite:
Disabling Hybrid Mode means that TS/CSS/HTML language support in .vue files is provided by @vue/language-server.
Enabling Hybrid Mode means that TS support within .vue files is provided by tsserver + @vue/typescript-plugin, CSS/HTML support is still provided by @vue/language-server,
If you want the behavior of v1 takeover mode, you actually need to disable Hybrid Mode.

RayGuo-ergou: Yeah I realized that, my brain does not work properly atm

README.md
gegoune: @johnsoncodehk With both `@vue/language-server` and `@vue/typescript-plugin` installed globally (pnpm) there is nothing in `/Users/gegoune/.pnpm-global/5/node_modules/@vue/language-server/node_modules`. Doesn't that mean that I need to specify path to `/Users/gegoune/.pnpm-global/5/node_modules/@vue/typescript-plugin` in `location`?

RayGuo-ergou: You should be fine with `npm install` I think pnpm uses it link magic so there's no `.../@vue/language-server/node_modules`

CofCat456:
Perhaps we should establish a community for Neovim users in the forum, as we have done in the past. In addition to providing correct configurations, it can also facilitate effective communication.

Thanks for that PR though!

RayGuo-ergou:
@gegoune Do not if you have received the notification that I have request a review again but just in case.

I have updated the docs and should be more clear in my opinion.

The config title changed from 'xxx version' to 'Hybrid/none-hybrid'.

Would you review it again when you have time 🙇‍♂️ thanks!

gegoune (approved):
Very nice, thank you!

Would be nice to see what are pros and cons of each configuration approach, but that's out of scope for this PR and not related to Neovim only.

Thank you for taking your time to make it so easy to understand!

README.md
```diff
@@ -27,8 +27,89 @@
...
+        location = vue_language_server_path,
+        languages = { 'vue' },
+      },
+    },
+  },
+  filetypes = { 'typescript', 'javascript', 'javascriptreact', 'typescriptreact', 'vue' },
+}
+
+-- No need to set `hybridMode` to `true` as it's the default value
+lspconfig.volar.setup {}
```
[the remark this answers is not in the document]
gegoune: This change would have to be submitted to https://github.com/neovim/nvim-lspconfig. Code above just uses configuration from lspconfig.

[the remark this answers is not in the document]
so1ve: Good to know, thanks!

johnsoncodehk:
LGTM, thanks everyone!

PR #4157, opened 2024-03-25
fix: Revert the dark pattern advertisement in the statusbar

basdp:
This PR reverts the advertisement that is being shown in the status bar. The status bar is not a place to advertise. We should stick to the regular spaces for asking for sponsorships, including the about page, the extension information page, the repository readme.md file, etc. Using the status bar for this purpose is also explicitly against the Visual Studio Code UX Guidelines for extensions (https://code.visualstudio.com/api/ux-guidelines/overview).

We should respect the developers that use our product. They are the reason we can exist.


johnsoncodehk:
Please respect this product! PR is not for expressing personal opinions. I've answered your point at https://github.com/vuejs/language-tools/discussions/4127#discussioncomment-8908550.

If you disagree, you can definitely fork and release another version.

PR #4181, opened 2024-03-31
Added simple tests for vue-tsc and custom SFC extensions.

cabal95:
Adds two simple tests for vue-tsc to ensure volarjs/volar.js#155 is working correctly. So this should not be merged until that is merged.

These basically duplicate two of the existing `component.vue` test files with a custom extension and then update the dts suite to enable that custom extension.


johnsoncodehk:
Thanks!

PR #4184, opened 2024-03-31
feat(language-service): drag and drop import respects tsconfig path aliases

### LINKED ISSUE #4125: [Feature] Drag and Drop Component Import respects tsconfig path aliases
When tsconfig is configured with `"paths": { "@/*": ["src/*"] }`, drag `/src/components/HelloWorld.vue` into `/src/App.vue` and it should be imported` @/components/HelloWorld.vue` instead of `./components/HelloWorld.vue`.

johnsoncodehk:
Closes #4125

TODO:

- [x] Respect `typescript.preferences.importModuleSpecifier` setting


packages/typescript-plugin/lib/requests/getImportPathForFile.ts
```diff
@@ -1,14 +1,66 @@
...
+		): {
+			getModuleSpecifierForBestExportInfo(
+				exportInfo: readonly SymbolExportInfo[],
+				symbolName: string,
+				position: number,
+				isValidTypeOnlyUseSite: boolean,
+				fromCacheOnly?: boolean
+			): { exportInfo?: SymbolExportInfo, moduleSpecifier: string, computedWithoutCacheCount: number; } | undefined;
+		};
+	} = (ts as any).codefix;
```
newVincentFong, line 47: How does it work? 🧐
Will `typescript` secretly export some internal services?

johnsoncodehk: Yes, some APIs are not exposed in dts.

PR #4195, opened 2024-04-02
chore: add reply new issue workflow 

jahnli:
image

### DIFF
--- a/.github/workflows/reply-new-issue.yml
+++ b/.github/workflows/reply-new-issue.yml
@@ -1,0 +1,29 @@
+name: Reply New Issue
+
+on:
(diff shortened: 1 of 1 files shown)


rchl:
This seems pointless. The same could be conveyed in the initial issue template (which @johnsoncodehk said before that he doesn't want to have).

so1ve:
We don't have an issue template. And this message is not helpful.

jahnli:
github templates can be used

jahnli:
In response to a question, actually
Questions are answered automatically, first-time contributors are welcome, and questions are regularly closed at......

PR #4196, opened 2024-04-02
fix(tsc): remove fake global types holder

blake-newman:
composite projects currently do not rebuild efficently because we attach a fake place holder vue file. This is used to generate global types once for the project.

As the fake placeholder file is added the the roots file, the second build will detect a change because the root file does not exist for discovery with tsc.

> Project 'packages/design-system/tsconfig.json' is out of date because buildinfo file
> 'packages/design-system/tsconfig.tsbuildinfo' indicates that file '[.../path/to/fake-file.vue]'
> was root file of compilation but not any more.

remove adding the fake file to the roots, and instead apply the globals to the first vue file and then remove them when writing .d.ts.


johnsoncodehk:
@blake-newman Made some adjustments, please let me know if this causes performance regression again in your project.

johnsoncodehk:
Thanks!

Issue #3976, opened 2024-03-04
Using extensing under VS Code Run/Debug does not work.


cabal95:
This might be related to #3928 in that it might be a windows only issue, I won't be able to test on Linux for a couple days. The named pipe JSON file never seems to be created when the extension is executed from inside VS Code - like when you are trying to develop and/or debug the extension.

The named pipe JSON file is created when the extension is used normally. But if I delete it from my AppData folder and then run the extension under VS Code development, it does not get re-created.

I tried setting a VS Code breakpoint in `startNamedPipeServer` but it never seems to resolve the breakpoint, even after attaching to the language server. So there must be some other piece I would need to attach to in order to step through that code.

I am assuming this named pipe issue is why it can't contact the language server and thus why everything comes up red.

![image](https://github.com/vuejs/language-tools/assets/1119863/a4a1798a-da65-4a4c-bdba-531b859e6235)

Note: This was discovered while trying to debug custom extensions being broken in the 2.x release. I will open a new issue for that and see if I can debug that further under Linux.

Ilanaya:
Experiencing the same issue on the windows machine. Everything works fine on the linux.

[Minimal reproduction repo](https://github.com/Ilanaya/volar-starter)
- Step to reproduce:
Preconditions: windows machine
1. `Launch Client`
2. Open reproduction repo in opened window
3. Open any `vue` file

- Actual result:
![image](https://github.com/vuejs/language-tools/assets/74474615/2a31c2fb-93f9-4af1-bd44-cce12baa4d6a)

- Expected result:
No errors

DanielViberg:
This might be the cause for my issues on linux:

Distributor ID:	Ubuntu
Description:	Ubuntu 22.04.2 LTS
Release:	22.04
Codename:	jammy

Using vim as client:
![image](https://github.com/vuejs/language-tools/assets/27137435/8ec9c207-f3da-4b5a-b1f0-73543ff02c7c)

cabal95:
@Ilanaya I re-tested this today, 2024-03-31, from master branch and I am unable to reproduce this now. Can you verify if you are still having this same issue? If not I'll close this.

Ilanaya:
@cabal95 yep, you can close

cabal95:
Closing as unable to replicate now so it may have been fixed or was related to something to else.

PR #4201, opened 2024-04-03
chore(server): use `os.tempdir()` instead of hardcoded `pipeFile` paths

Ilanaya:
As [this change](https://github.com/vuejs/language-tools/commit/72e38d52d95cdbd1aa29011a6a60657eb619deef#diff-cec35c50723f4febc3d6909d21749dbe907bc561d74f6e52b8a0de6dd9d9f7e4R13-R15) was indroduces in 2.0.2 it looks logical to use the same logic to create `pipeFiles` too.
 
P.S.: I found this inconsistency during debugging #3976 (see [this comment](https://github.com/vuejs/language-tools/issues/3976#issuecomment-1979105927) for more information) as this still was an issue on my windows machine. At some moment I mentioned a warning in the `tsconfig.json` which complains on the absense of the `dist/schemas/vue-tsconfig.schema.json` file.

image

So I manually run `npm run build` in the `extensions/vscode` directory and then the issue was resolved, so I have some related questions.

1.  How to debug `packages/typescript-plugin/index.ts`, `packages/typescript-plugin/lib/server.ts` etc? Breakpoints do not work there if using `Launch client`/`Attach to server`
2. As I can see `npm run watch` in root do not create `dist` dir in `extensions/vscode`. Is this desired behaviour? Every time I open project with `tsconfig` in dev mode I see a warning like on screenshot above.

### DIFF
--- a/packages/typescript-plugin/lib/server.ts
+++ b/packages/typescript-plugin/lib/server.ts
@@ -8,6 +8,8 @@ import { getPropertiesAtLocation } from './requests/getPropertiesAtLocation';
 import { getQuickInfoAtPosition } from './requests/getQuickInfoAtPosition';
 import { NamedPipeServer, connect, readPipeTable, updatePipeTable } from './utils';
 import type { Language, VueCompilerOptions } from '@vue/language-core';
+import * as path from 'path';
+import * as os from 'os';
 
 export interface Request {
 	type: 'containsFile'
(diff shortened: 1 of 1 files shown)


so1ve:
This won't work IMO since named pipes on Windows require sockets begin with `\\.\pipe`

Ilanaya:
> This won't work IMO since named pipes on Windows require sockets begin with `\\.\pipe`

Thanks! I didn't know about this restriction. Should I close and maybe move my questions to discussions? I am really interested in debugging information.

johnsoncodehk:
The debugging configuration of TS plugin/tsserver is a bit troublesome. I just use `console.log` for debugging. But if you want to do it, please refer to https://github.com/microsoft/TypeScript/wiki/Debugging-Language-Service-in-VS-Code.

johnsoncodehk:
Close since as @so1ve said, this doesn't work on Windows.

PR #4243, opened 2024-04-10
docs(vscode): remove outdated formatters section

BBboy01:
I've got 404 while visiting https://github.com/vuejs/language-tools-plugins, maybe is the repo name has changed


johnsoncodehk:
The repo link is https://github.com/volarjs/services, but since v2.0 custom formatters is not support for now, so I will remove this section. Thanks!

PR #4245, opened 2024-04-10
fix(tsc): re-introduce global types removal check

blake-newman:
The issue with the guard clause prior was that the out directory for the .d.ts file is likely to be a different place from the global types holder file. This meant the clause was not accurately detecting if it should remove the global types.

The guard clause is needed to some degree as large projects with many emitted files would run this removal which is fairly slow. vue-tsc is unusable without this guard clause.

Instead just get the file name from the file being written and the global types holder, if the dts file starts with the global types holder then perform the global types removal logic.

This allows for the output and source paths to be different. The removal may run more than once if similar names are present but it won't run on every file to be emitted.


johnsoncodehk:
Thanks for the PR! Can you share how much the build time improved by introducing this condition in your case?

blake-newman:
Our application: 
- 3340 (ts/x) files
- 851 vue files

Full build no guard check `4981s` (1 hour 23 min 1 second) 
Full build guard check `82s` (1 min 22 seconds)

sawmurai:
Another data point: Our build slowed down by about 50% with the latest version.

johnsoncodehk:
I haven't been able to reproduce the performance difference caused by this change. I would merge it first, but we'd better have a test case for this in case it breaks again in the future.

messenjer:
For my project: 

Before 2.0.12:

```
> vue-tsc --build --force

npm run type-check  216.44s user 9.80s system 116% cpu 3:13.77 total
```

After 2.0.13:

```
> vue-tsc --build --force

npm run type-check  35.12s user 1.35s system 177% cpu 20.588 total
```

80% faster

PR #4248, opened 2024-04-10
📝 correct location string in tsserver plugin initializationOptions

friedow:
This broke during the revert in https://github.com/vuejs/language-tools/commit/1f543969c31c23a44bdfae0a7d57c99cdfce0a2b.


johnsoncodehk:
Thanks for the PR, but `.../@vue/language-server` is indeed the correct location, see https://github.com/vuejs/language-tools/pull/4134#discussion_r1533451116.

PR #4260, opened 2024-04-12
fix(tsc): improve regexp performance for global type removal

blake-newman:
The issue with the removal of global types was the RegExp, for certain files written like `.tsbuildinfo` which is a single line long contents file it would scan the whole line scanning for the characters.

By specifying `m` flag it causes `^` and `$` to match the begin/end of each line (not only begin/end of string).

This removes the need for the check of the filename to do the replacement check.

### DIFF
--- a/packages/tsc/index.ts
+++ b/packages/tsc/index.ts
@@ -1,6 +1,5 @@
 import { runTsc } from '@volar/typescript/lib/quickstart/runTsc';
 import * as vue from '@vue/language-core';
-import * as path from 'path';
 
 const windowsPathReg = /\\/g;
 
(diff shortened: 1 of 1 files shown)


packages/tsc/index.ts
```diff
@@ -65,7 +58,7 @@ export function run() {
 	}
 }
 
-const removeEmitGlobalTypesRegexp = /[^\n]*__VLS_globalTypesStart[\w\W]*__VLS_globalTypesEnd[^\n]*\n/g;
+const removeEmitGlobalTypesRegexp = /^[^\n]*__VLS_globalTypesStart[\w\W]*__VLS_globalTypesEnd[^\n]*\n?$/mg;
```
rchl, line 61: I think those initial/trailing `\n` make no sense with `m` flag. Those should be ignored anyway when matching with `m` flag.

```suggestion
const removeEmitGlobalTypesRegexp = /^__VLS_globalTypesStart[\w\W]*__VLS_globalTypesEnd.+$/mg;
```

johnsoncodehk: cc @blake-newman

PR #4312, opened 2024-04-27
feat(language-core): support defineSlots destructuring

zhiyuanzmj:
This is a syntax error:
```ts
let __VLS_own!: (new ()=> { $slots: typeof { ...slots } })
//                                  ^^^^^^^^^^^^^^^^^^^
```
https://github.com/vuejs/language-tools/blob/master/packages/language-core/lib/codegen/script/template.ts#L106


johnsoncodehk:
Just make some refactoring, thanks!

PR #4326, opened 2024-04-30
fix(language-core): correct logic for defineSlots destructuring

zhiyuanzmj:
## Before:

image

## After: 
```ts
const { bottom } = __VLS_slots; 
const __VLS_slots = defineSlots<{
	bottom: (props: { num: number }) => any[],
}>()
```


johnsoncodehk:
LGTM, thanks!

PR #4358, opened 2024-05-09
chore: specify `packageManager`

so1ve:
Not sure why my local environment always uses pnpm v8 for volar, and it causes pnpm-lock.yaml to be changed.


johnsoncodehk:
Thanks!

PR #4365, opened 2024-05-10
feat(typescript-plugin): add emit JSDoc support

### LINKED ISSUE #1894: JSDoc intellisense for Vue 3 setup events and component description
### Discussed in https://github.com/johnsoncodehk/volar/discussions/1473

<div type='discussions-op-text'>

<sup>Originally posted by **vincerubinetti** June 17, 2022</sup>
I'm using Vue 3 and the composition API with `<script setup>`. Just upgraded to the latest version of Vue and and updated my Volar extension.

All of this works great (note the comments):

**`TestComponent.vue`**

```vue
<script setup lang="ts">
/** fun test component */

interface Props {
  /**  text to show in label */
  text: string;
}

const props = defineProps<Props>();

props.text; // hovering over this shows "text to show in label"

interface Emits {
  /** test event */
  (event: "test"): void;
}

const emit = defineEmits<Emits>();

emit("test"); // hovering over this shows "test event"
</script>
```

**But if I use the component in another template, there are some holes:**

```vue
<template>
  <TestComponent
    text="test text"
    @test="() => null"
  />
</template>
```

Hovering over the `text` attribute shows "text to show in label", but hovering over `@test` does not show "test event". It would also be cool if hovering over the `TestComponent` tag showed "fun test component" (and any other JSDoc info I suppose).

I did try different JSDoc tags like `@description` and `@function` to see if those would show up, but they didn't.

It'd be cool if Volar could support this. I think there are other tools/plugins that do this, but they all seem to be more on the side of doc generators. I don't need a doc generator because my app is small enough, but I do want/need intellisense for components/props/events because I forget what they are.

Sorry if this is the wrong place for this, I'm not sure if this falls under the scope of Vue/Volar/VS Code/TypeScript/etc.</div>

johnsoncodehk:
Close #1894

It is not possible to generate virtual code that can display emit JSDoc. As an alternative, JSDoc is now calculated from the original type through TypeChecker and attach to the hover message via proxy getQuickInfoAtPosition.

### DIFF
--- a/packages/language-core/lib/codegen/template/element.ts
+++ b/packages/language-core/lib/codegen/template/element.ts
@@ -34,6 +34,7 @@ export function* generateComponent(
 	const var_originalComponent = matchImportName ?? ctx.getInternalVariable();
 	const var_functionalComponent = ctx.getInternalVariable();
 	const var_componentInstance = ctx.getInternalVariable();
+	const var_componentEmit = ctx.getInternalVariable();
 	const var_componentEvents = ctx.getInternalVariable();
 	const var_defineComponentCtx = ctx.getInternalVariable();
 	const isComponentTag = node.tag.toLowerCase() === 'component';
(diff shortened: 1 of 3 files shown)


redhat:
Are there any specific usage examples? I still can't get the comment hints here.

PR #4392, opened 2024-05-16
fix(language-service): wrong vue-i18n url for language blocks

kazupon:


so1ve:
Those files are automatically generated from vuejs.org I guess?

kazupon:
Sorry, I may not understand the internals of this repository. 😓 

If so, is this the place to fix it?
https://vuejs.org/api/sfc-spec.html#custom-blocks

so1ve:
Yes, we use an automated script to update those data: https://github.com/vuejs/language-tools/blob/master/packages/language-service/scripts/update-html-data.js

kazupon:
Thanks!
I've got it. I've already opened the fix PR https://github.com/vuejs/docs/pull/2875
when that PR will be merge, that data will be updated with that script.

Thank you yor your quick triages!

PR #4413, opened 2024-05-24
fix(language-core): set __VLS_template as an asynchronous function

zhiyuanzmj:
### `__VLS_template` should be an asynchronous function.
Prevent the compile error:
```text
 'await' expressions are only allowed within async functions and at the top levels of modules.
```
image


zhiyuanzmj:
Or maybe we should move `__VLS_internalComponent` outside the `__VLS_template` function

johnsoncodehk:
Thank you for your PR! But this resulted in failure to detect misuse of await in template, I just added a test in master branch.

johnsoncodehk:
> Or maybe we should move `__VLS_internalComponent` outside the `__VLS_template` function.

This is a good point, but requires special handling for class component codegen, so it's better not to do this to avoid further complexity.

PR #4430, opened 2024-05-31
fix(language-core): use defineEmits calls instead of type infer

zhiyuanzmj:
## It will be a type error when I pass the compiled ts file to twoslash.
image


johnsoncodehk:
LGTM, thanks!

PR #4463, opened 2024-06-12
docs(nvim): add nvim-cmp integration

RayGuo-ergou:
Hi 👋

This is a documentation update for neovim user, `nvim-cmp` is a popular neovim plugin used for auto complete code, I know this might be a little bit off topic but I hope this can help some neovim users. 

As mentioned in the content, I was inspired by [this post](https://twitter.com/youyuxi/status/1799701740000035136).

Basically this changes:

image
image

to:

image
image

### DIFF
--- a/README.md
+++ b/README.md
@@ -124,6 +124,28 @@ sources = cmp.config.sources({
         return true
       end
 
+      -- If you want to be more accurate and have nvim-treesitter installed, you can check if the cursor is in a start tag
+      -- This would be helpful for the following code
+      -- type ExampleType = '@a' | '@b' | '@c' | '1@c'
+      -- const v: ExampleType = ''
+
+      -- local ts_utils = require('nvim-treesitter.ts_utils')
+      -- local function is_in_start_tag()
+      --   local node = ts_utils.get_node_at_cursor()
+      --   while node do
+      --     if node:type() == 'start_tag' then
+      --       return true
+      --     end
+      --     node = node:parent()
+      --   end
+      --   return false
+      -- end
+      --
+      -- -- If not in start tag, return true
+      -- if not is_in_start_tag() then
+      --   return true
+      -- end
+
       local cursor_before_line = ctx.cursor_before_line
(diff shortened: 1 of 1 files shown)


gegoune:
Please note that this will affect all LSP servers which might not be desirable. It could be wrapped in conditional statement based on LSP server name, what do you think?

RayGuo-ergou:
> Please note that this will affect all LSP servers which might not be desirable. It could be wrapped in conditional statement based on LSP server name, what do you think?

Thanks for pointing that out! You are absolutely right! Since javascript/typescript does not use `:` and `@` I think check buffer type would be enough.

gegoune:
I think there is `entry.source.name`, which is what I had in mind of using here. Not sure whether by saying ' javascript/typescript does not use…' you mean `script` part of Vue file or completely separate files?

Reason I am asking is because I think using filetype check will make it work outside of `template` section, which is the only ona that this change is relevant to, isn't it? (in which case my `entry.source.name` check still won't be enough).

Need to be careful here though as it might have wide performance radius blast.

RayGuo-ergou:
>  think there is entry.source.name, which is what I had in mind of using here.

It will return the name like `nvim_lsp` or `buffer` so I dont think it's the right thing to use here.

> . Not sure whether by saying ' javascript/typescript does not use…' you mean script part of Vue file or completely separate files?

What I mean is neither JavaScript nor Typescript use `@` or `:` as keyword. So e.g. lua has `a:b`

> Need to be careful here though as it might have wide performance radius blast.

I am not worry to much about this to be honestly, as first this will only apply in vue files, second `/^:/` or `/^:-on/` should be very fast as it just check the start of the string.

johnsoncodehk:
Hi @RayGuo-ergou, do you want to open a [discussion](https://github.com/vuejs/language-tools/discussions) and add the discussion link in the readme instead? This may be easier to manage Q&A.

johnsoncodehk:
Thank you!

PR #4474, opened 2024-06-16
feat(language-core): support Vue 3.5

zhiyuanzmj:


johnsoncodehk:
Thank you for your efforts!

I reverted the changes to GlobalComponents because that was what needed to be resolved in vue core types.

Issue #4497, opened 2024-06-24
Error: VirtualCode id must be lowercase: scriptSetupFormat


davidmatter:
### Vue - Official extension or vue-tsc version

2.0.22

### VSCode version

-

### Vue version

3.4.30

### TypeScript version

5.5.2

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (8) arm64 Apple M1 Pro
    Memory: 57.80 MB / 16.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 22.3.0 - ~/Code/data-visualization/packages/rds/node_modules/.bin/node
    Yarn: 1.22.19 - /opt/homebrew/bin/yarn
    npm: 10.8.1 - ~/.nvm/versions/node/v22.3.0/bin/npm
    pnpm: 9.4.0 - ~/.nvm/versions/node/v22.3.0/bin/pnpm
    bun: 1.1.15 - /opt/homebrew/bin/bun
  Browsers:
    Chrome: 126.0.6478.63
    Safari: 17.5
```

### Steps to reproduce

1. `pnpm create vite@latest` -> Vue project with TS
2. `pnpm update`
3. `vue-tsc --declaration --emitDeclarationOnly -p tsconfig.app.json`

Throws:
`Error: VirtualCode id must be lowercase: scriptSetupFormat`

## Workaround
Until the fix is merged: Add the following to your `package.json` if you're using `pnpm`

  ```json
"pnpm": {
    "overrides": {
      "@volar/typescript": "2.3.1",
      "@volar/language-core": "2.3.1"
    }
 }
```

For NPM, Yarn etc. you'll find other solutions.

davidmatter:
Regression from https://github.com/volarjs/volar.js/commit/1f6e2bf4a0bad217607e196f378d9a1198293c3b

chaozwn:
same error.
![image](https://github.com/vuejs/language-tools/assets/51357674/211d9256-30f3-4b3c-8cc0-2e456c557180)

lancewilhelm:
Same error as @chaozwn also on neovim, mac M2 (and I believe my Ubuntu (noble) machine suffered the same error). I have tried different node LTS versions as well as the latest with no success. Tried creating a new Vue project with `npm create vue@latest` with the hope that my project with the issue, but same error on a fresh project (default options, 'no' to everything).

JianGuoHt:
+1

davidmatter:
[...]

PR #4498, opened 2024-06-24
fix(vue-tsc): update required volar version

davidmatter:
CI runs with --frozen-lockfile which contains dependency versions that don't match with what will be installed in the end in userland because the lockfile is not shipped.

This results in tests not running on the same code as will be installed for the user, causing e.g. https://github.com/vuejs/language-tools/issues/4497

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -543,7 +543,7 @@
 	"devDependencies": {
 		"@types/semver": "^7.5.3",
 		"@types/vscode": "^1.82.0",
-		"@volar/vscode": "~2.3.1",
+		"@volar/vscode": "~2.3.3",
 		"@vue/language-core": "2.0.22",
 		"@vue/language-server": "2.0.22",
 		"@vue/typescript-plugin": "2.0.22",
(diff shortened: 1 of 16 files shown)


Ciscodefran:
Why don you update this to 2.3.3
https://github.com/volarjs/volar.js/releases/tag/v2.3.3

davidmatter:
I've created this PR before 2.3.3 was released

davidmatter:
Not sure if @johnsoncodehk wants to pin to an exact volar version? I imagine you'd like to be flexible in updating volar without needing to re-publish vue-tsc. The problem is just the possibility of having upstream regressions when volar is updated. Feel free to close this one.

johnsoncodehk:
@davidmatter I think we can add a CI test to test without `--frozen-lockfile` flag. Would you want to update this PR for it?

davidmatter:
Well, if I think about it: Even if we run CI without `--frozen-lockfile`, updating volar can still break the language tools when it's not pinned to a specific version as it's a direct dependency of vue-tsc that users will automatically fetch when they update their dependencies. Removing `--frozen-lockfile` would only help when vue-tsc is re-published :)

johnsoncodehk:
I just updated the PR to update the required version to `~2.3.4` to avoid installing the regression `2.3.2`.

> If semver is used, this should not be a problem in general - we just need to be careful not to introduce breaking changes when not pinned.

This is what https://github.com/volarjs/ecosystem-ci for, it got a little damaged recently and I'm going to get it working again.

PR #4506, opened 2024-06-26
feat: better sfc-level tag name completion

KermanX:
Typing `lang="ts"` every time when start writing a new SFC is very annoying. This PR adds completion for that.

| **Before** | image |
| -- | -- |
| **After** | image |

**Improvements**
- Removed `!DOCTYPE` and `Custom Blocks`
- Add `script lang="ts"` and `script setup lang="ts"`

**Notes**
- The description of `script lang="ts"` is the same as `script`, and the description of `script setup lang="ts"` is the same as `script setup`. I am not sure how to update this.
- Exact matching is used in this PR. I don't know if a looser matching method should be used.
- Should the order of the completion items be changed?

### DIFF
--- a/packages/language-service/lib/plugins/vue-sfc.ts
+++ b/packages/language-service/lib/plugins/vue-sfc.ts
@@ -172,6 +172,17 @@ export function create(): LanguageServicePlugin {
 						return result;
 					});
 				},
+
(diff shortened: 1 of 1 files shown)


johnsoncodehk:
LGTM, thanks!

PR #4513, opened 2024-06-29
fix(tsc): bump peer typescript version

so1ve:


packages/tsc/package.json
```diff
@@ -21,7 +21,7 @@
 		"semver": "^7.5.4"
 	},
 	"peerDependencies": {
-		"typescript": "*"
+		"typescript": ">=5.3.0"
```
johnsoncodehk, line 24: >=5.3.0 is a requirement of @vue/typescript-plugin, vue-tsc only requires >=5.0.0.

PR #4515, opened 2024-07-01
chore: add 'expected' and 'actually happening' to bug report template

davidmatter:
Some issues don't provide enough information on what is expected vs happening.


johnsoncodehk:
Thanks!

PR #4542, opened 2024-07-07
test: global components prop validation

davidmatter:
Add test case for https://github.com/vuejs/language-tools/wiki/Global-Component-Types


johnsoncodehk:
LGTM, thanks!

PR #4206, opened 2024-04-04
feat(vscode): automatically enable Hybrid Mode

johnsoncodehk:
Closes #4131

This introduces the default value of `"auto"` for `vue.server.hybridMode`, which automatically enables it based on the compatibility of installed TS plugin extensions with Hybrid Mode.

The following TS plugin extensions are currently known to be compatible:

- unifiedjs.vscode-mdx
- astro-build.astro-vscode
- ije.esm-vscode
- johnsoncodehk.vscode-tsslint
- VisualStudioExptTeam.vscodeintellicode

The following TS plugin extensions are currently known to be incompatible:

- styled-components.vscode-styled-components
- svelte.svelte-vscode
- Divlo.vscode-styled-jsx-languageserver

If you know of any other compatible TS plugin extensions, please report them here so that we can update the whitelist.

To verify compatibility, simply set `"vue.server.hybridMode": true` in your VSCode settings and reload VSCode to check if TS support in .vue files is working as expected.

PR #4543, opened 2024-07-08
fix(vscode): add vue vine to hybrid mode compatible list

so1ve:
https://github.com/vuejs/language-tools/pull/4206#issuecomment-2212537184


johnsoncodehk:
Thanks!

PR #4544, opened 2024-07-08
fix(language-core): generate codes for `:key` prop on `v-for` and `v-if` node

### LINKED ISSUE #4539: key is not type-checked on `template`

2.0.26

### VSCode version

1.92.0-insider

### Vue version

3.4.31

### TypeScript version

5.5.3

### System Info

_No response_

### Steps to reproduce

1. `git clone git@github.com:brc-dd/volar-template-key-bug.git`
2. `pnpm i`
3. `pnpm vue-tsc --noEmit`

### What is expected?

Error on both line 4 and line 5 of `Foo.vue`

Error is only on line 5. There should be an error saying `Type '{}' is not assignable to type 'PropertyKey | undefined'.` for `:key` prop of `<template>` too.

https://github.com/brc-dd/volar-template-key-bug

### Any additional comments?

_No response_

so1ve:
I created a fake component to generate codes since `IF_BRANCH` nodes are not valid elements. Not sure if it is the best approach.

fixes #4539


johnsoncodehk:
I temporarily changed the issue to a draft because we should wait for https://github.com/vuejs/core/issues/11322 to be fixed by the upstream. Forcibly handling unrecognized props in language tools will make the code we need to maintain too complex.

PR #4609, opened 2024-07-23
fix(language-core): support parse method to access ctx var in object

linghaoSu:
fix https://github.com/vuejs/language-tools/issues/4604


johnsoncodehk:
LGTM, thanks!

Issue #4537, opened 2024-07-05
Certain CSS classes causes Typescript generic types to fail


Pentadome:
### Vue - Official extension or vue-tsc version

2.0.26

### VSCode version

Version: 1.92.0-insider Commit: 82104a3a6cf8713b81e5bbd97960dbf5c82a816a Date: 2024-07-05T00:00:00.000Z Electron: 29.4.0 ElectronBuildId: 9728852 Chromium: 122.0.6261.156 Node.js: 20.9.0 V8: 12.2.281.27-electron.0 OS: Windows_NT x64 10.0.22631

### Vue version

3.4.31

### TypeScript version

5.5.3

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (16) x64 AMD Ryzen 7 7800X3D 8-Core Processor
    Memory: 20.36 GB / 31.22 GB
Binaries:
    Node: 22.4.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.22 - C:\Program Files\nodejs\yarn.CMD
    npm: 10.8.1 - C:\Program Files\nodejs\npm.CMD
Browsers:
    Edge: Chromium (126.0.2592.87)
    Internet Explorer: 11.0.22621.3527
```

### Steps to reproduce

This works:
```vue
<script setup lang="ts" generic="T extends 'checkbox' | 'radio'">
defineProps<{
  type: T;
}>();

defineModel<boolean>({ required: true });
</script>

<template>
  <div class="relative size-5">

  </div>
</template>
```

This causes error:
```vue
<script setup lang="ts" generic="T extends 'checkbox' | 'radio'"> <-- Cannot find name T
defineProps<{
  type: T; <-- Cannot find name T
}>();

defineModel<boolean>({ required: true });
</script>

<template>
  <div class="relative size-5">

    <div
      aria-hidden="true"
      class="pointer-events-none absolute size-full border border-primary2 p-1 peer-checked:*:block peer-[&[type='radio']]:rounded-full peer-[&[type='radio']]:*:rounded-full">
      <div class="hidden size-full border border-primary2 bg-primary2" />
    </div>
  </div>
</template>

```

<details>
  <summary><s>This causes another error:</s></summary>

```vue
<script setup lang="ts" generic="T extends 'checkbox' | 'radio'">
defineProps<{
  type: T;
}>();

defineModel<boolean>({ required: true });
</script>

<template>
  <div class="relative size-5">
    <input
[...]

PR #4619, opened 2024-07-29
fix(language-core): escape \ and ' in className avoid type error

linghaoSu:
fix https://github.com/vuejs/language-tools/issues/4537


johnsoncodehk:
LGTM, thanks!

PR #4623, opened 2024-07-30
fix: semantic highlight of the end tag of namespaced elements

### LINKED ISSUE #3341: Wrong highlighting for namespaced components
Ref: https://vuejs.org/api/sfc-script-setup.html#namespaced-components

image

KermanX:
fix #3341

| Previous | Now |
| - | - |
| image | image |

The remaining problem is that renaming doesn't work. If you rename `C` in the script block, everything is fine. However, if `C` in `<A.B.C` is renamed to `C123`, it will become `<C123123`; if it's renamed to `D`, an error message "Rename failed to apply edits" will show.


johnsoncodehk:
LGTM, thanks!

PR #4626, opened 2024-07-30
fix(language-plugin-pug): should cache proxyed object

### LINKED ISSUE #4335: `language-plugin-pug` unused prop with `v-slot` and `v-for`
We are encountering the following issue when using Pug, with `v-slot` and `v-for`.
The goal was to take all slots passed into our wrapper component, and pass them down into a child component. To do this, we use `v-for` with `$slots` and create a `v-slot` for each of them, where the content is a `slot` of the same name.
The `ts-plugin` gives us the following error: `'mySlot' is declared but its value is never read`
```
template(v-slot:[mySlot]="slotProps" v-for="(_, mySlot) in $slots" :key="mySlot")
			slot(:name="mySlot" v-bind="slotProps")
```
image

<br>
<br>

We have also tried with a less complex example, and it appears the issue always occurs when a `template` has both a `v-slot` and a `v-for` on it.
image

<br>

We have noticed that this does not occur when using HTML, only Pug. Here is the HTML equivalent:
```
<template v-slot[mySlot]="slotProps" v-for="(_, mySlot) in $slots" :key="mySlot">
    <slot :name="mySlot" v-bind="slotProps"></slot>
</template>
```

KermanX:
fix #4335.

This PR caches proxyed objects created by the Pug plugin, so that logic involving equality checking would work.

Previously, the following line doesn't remove `v-for` node because every time `node.props` generates a different array.
https://github.com/vuejs/language-tools/blob/9ef235b3cabe1639c3e048bcf347c76575ecb09f/packages/language-core/lib/codegen/template/templateChild.ts#L138

| Prev | image |
| - | - |
| **Now** | image |

### DIFF
--- a/packages/language-plugin-pug/index.ts
+++ b/packages/language-plugin-pug/index.ts
@@ -34,8 +34,9 @@ const plugin: VueLanguagePlugin = ({ modules }) => {
 					return createProxyObject(completed);
 
 					function createProxyObject(target: any): any {
+						const proxys = new WeakMap();
 						return new Proxy(target, {
-							get(target, prop) {
+							get(target, prop, receiver) {
 								if (prop === 'offset') {
 									const htmlOffset = target.offset;
 									const nums: number[] = [];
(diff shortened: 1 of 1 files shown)


johnsoncodehk:
LGTM, thanks!

PR #4634, opened 2024-07-31
feat(language-core): inlay hints for destructured props

### LINKED ISSUE #4622: Inlay hints for destructured props

Reactive Props Destructure 语法糖可能对理解变量是否为 prop 带来额外的心智负担

在解构的 prop 变量前添加 `props.` 的嵌入提示有助于解决这个问题

image

KazariEX:
resolve #4622 

It will look like:

image


KazariEX:
Which is better?

image

image

packages/language-core/lib/parsers/scriptSetupRanges.ts
```diff
@@ -14,6 +15,8 @@ export function parseScriptSetupRanges(
 
 	const props: {
 		name?: string;
+		destructured?: ts.Identifier[];
+		destructuredReferences?: [ts.Identifier, boolean /* isShorthand */][];
```
so1ve, line 19: Can we use named tuple here?

test-workspace/tsc/vue3.5/components/script-setup-destructured-props.vue
```diff
@@ -0,0 +1,44 @@
+<script setup lang="ts">
```
so1ve, line 1: Please change to use tabs to indent

packages/language-core/lib/parsers/scriptSetupRanges.ts
```diff
@@ -254,6 +257,10 @@ export function parseScriptSetupRanges(
 				};
 
 				if (ts.isVariableDeclaration(parent)) {
+					if (ts.isObjectBindingPattern(parent.name)) {
```
so1ve, line 260: Perhaps we'd better move the logic into https://github.com/KazariEX/vuejs-language-tools/blob/8379263ecc99ef550cc61f9567d70467633a8131/packages/language-core/lib/parsers/scriptSetupRanges.ts#L236 and use https://github.com/vuejs/language-tools/blob/dc6f943c6ccb26778906a49532966aaf69fdc5a5/packages/language-core/lib/codegen/common.ts#L44-L68

KazariEX: Do you mean to move the all logic into the main function?

KazariEX:
I don't know why it can't get any inlayHints... Hope to know the correct solution.

johnsoncodehk:
> Which is better?
> 
> image
> 
> image

`: props.foo` is better. 👍

johnsoncodehk:
This is close to complete! However, executing `findDestructuredProps` in codegen affects the performance of vue-tsc. I am refactoring to move the logic related to inlay hints to the language-service package to avoid unnecessary computations in vue-tsc

KazariEX:
Actually I implemented it in the `language-service` at first, and only tried to implement it in codegen after being reminded about the magic of `__hint`.

Additionally, I noticed a feature about tracking reactive variables in the Insiders version. Does this imply that they can extract reusable logic?

johnsoncodehk:
> Additionally, I noticed a feature about tracking reactive variables in the Insiders version. Does this imply that they can extract reusable logic?

My impression is that the logic is different, it's okay let's see if we can reduce the code after merging this.

Mrlilili:
Is there a way to also display  inlay hints for destructured props in the template?

image

KazariEX:
> Is there a way to also display inlay hints for destructured props in the template?
> 
> image

I don't think it's necessary because all variables within the template will be automatically unwrapped.

Mrlilili:
> > Is there a way to also display inlay hints for destructured props in the template?
> > image
> 
> I don't think it's necessary because all variables within the template will be automatically unwrapped.

In the template, it is necessary to distinguish the source of this variable as being from props or a custom ref value，this can increase the readability of the code.

KazariEX:
If Props Reactive Destructure is not used, we cannot distinguish whether the variables in the template are props. Providing inlay hints only for the former not only yields little benefit, but also causes inconsistency.

PR #4640, opened 2024-08-03
feat(language-service): better completion for directives

KazariEX:
It provides completion for `v-cloak`, `v-once` and `v-pre` without `=""`, just like `v-else`


KazariEX:
Revert changes to `v-on` as this may cause inconsistent behavior with `v-bind`

KazariEX:
Revert changes to `v-on` as this may cause inconsistent behavior with `v-bind`

johnsoncodehk:
LGTM, thanks!

PR #4642, opened 2024-08-04
feat(language-core): document links for classname within `:class`

### LINKED ISSUE #4641: Document links for classname within `:class`

Currently, the document links for `scoped` classname only takes effect on the `class` attribute without `v-bind`

Apply document links on string literal keys of arrays or objects with `v-bind`

KazariEX:
resolve #4641 

image

- [x] document links
- [x] rename support


johnsoncodehk:
I think rename support can be left out for now. The current class name rename solution is unreliable due to a upstream issue (https://github.com/volarjs/volar.js/issues/176).

KazariEX:
This commit treats repeated class declarations as references to the first class, ensuring synchronization when renaming classes with the same name. 

This requires ignoring the CSS renaming behavior when using the typescript-semanitc renaming feature. Currently, since no interception is implemented, class renaming in the `<style>` will be overridden by the CSS functionality. We need to address this issue later.

KazariEX:
Since we have extracted the classnames as much as possible, the logic previously used for renaming support is no longer necessary.

johnsoncodehk:
Awesome, thanks! I tested it manually, it would be great if you could add a test case for it.

I plan to extract this feature to the language service plugin instead of relying on TS virtual code, which will improve the robustness of this feature.

PR #4643, opened 2024-08-04
fix(language-service): avoid converting internal id of special tags

KazariEX:
fix #4596


KazariEX:
Sorry about mistaken click... I've found a deeper logical error and will continue working on it.

PR #4647, opened 2024-08-05
feat(examples): init example projects

davidmatter:
To improve our documentation, we should start to create example projects that we can gradually extend. The main goal is to help users with configuring their editor, tsconfig, build scripts etc. as there is quite a lot of complexity involved.

Let's use this basic example as a starting point.


README.md
```diff
@@ -4,28 +4,34 @@
 
 💬 **#language-tools** on our [Discord Server](https://discord.gg/vue)
 
+## Example Projects
+
+- [Basic Example](./examples/basic/) A basic example project that highlights some aspects of the Vue language tools. This example is a good starting point for new users.
+- [[Todo] nvim](./examples/with-nvim/) If you are using Neovim, feel free to contribute an example project to help others.
```
rchl, line 10: Why would there be a editor-specific example here? The editor configuration is a separate concern IMO.

All in all I'm not convinced I see the point of those examples.

README.md
```diff
@@ -4,28 +4,34 @@
...
+- [[Todo] Vue2](./examples/with-vue2/) If you are using Vue 2, feel free to contribute an example project to help others.
+
 ## Packages
 
 - [Vue Language Features](https://github.com/vuejs/language-tools/tree/master/extensions/vscode) \
-*Vue, Vitepress, petite-vue language support extension for VSCode*
+  _Vue, Vitepress, petite-vue language support extension for VSCode_
```
rchl, line 16: The less changes you do the more likely your PR will be reviewed and accepted. Are those formatting changes really relevant to this PR?

README.md
davidmatter: Editor-specific examples can be used to highlight certain intricacies of different environments. I'm not saying it should be focused on specific editors only, but there are important points in that regard, e.g.

* explain how the hybrid version setting works in vscode
* nvim config files for a good editing experience
* vscode debugging and launcher profiles
* how to configure the project's typescript version in contrast to the typescript version of the vscode workbench

IDE unspecific:
* How to configure global components
* How to setup type declaration in libraries vs Vue applications
* How to setup generic components and what do you need to be aware of in regards to the language tools?
* How to implement different tsconfig setups with project references etc.
* How to configure your nuxt/element-plus/(?) libraries to work well with vscode

etc.

rchl: IMO editor specific quirks should be documented separately in text or delegated to other projects that take care of that. There is not that much you can show within a project regarding editor configuration.

As for your other (non-editor) points, I suppose some might make sense but it might be encroaching on being typescript/vue territory rather than something that should be documented within a vue language server.

PR #4650, opened 2024-08-05
fix(language-core): local name support for prop using runtime api

### LINKED ISSUE #4649: Type check error when using local name different from prop name with runtime api

2.0.28

### VSCode version

1.92.0-Insider

### Vue version

3.5.0-alpha.5

### TypeScript version

5.5.4

### System Info

_No response_

### Steps to reproduce

```
<script lang="ts" setup>
import Child from './component.vue';

const model = '1';
const foo = '1';
const bar = '1';
</script>

<template>
    <Child
        v-model="model"
        v-model:foo="foo"
        v-model:bar="bar"
    />
</template>
```

```
<script lang="ts" setup>
const modelAlias = defineModel({
    type: Number,
    required: true
});
const fooAlias = defineModel('foo', {
    type: Number,
    default: 1
});
const barAlias = defineModel('bar', {
    default: 1
});
</script>
```

### What is expected?

Correct type inference.

All props got `any` type.

_No response_

### Any additional comments?

_No response_

KazariEX:
fix #4649


KazariEX:
I need some help. I've built the correct virtual code, however, as long as the local variable is named different from the prop name, the type will be inferred as `any`

johnsoncodehk:
LGTM, thanks!

PR #4652, opened 2024-08-06
fix(language-plugin-pug): compute offset correctly of pug class

### LINKED ISSUE #1767: Pug class literal underline range incorrect
```html
<template lang="pug">
.foo
^^^
</template>

<style scoped>
.foo {}
</style>
```

KazariEX:
fix #1767


KazariEX:
It should be noted that this currently only applies to cases where classnames are completely consecutive. Due to the code parsing logic, it cannot obtain the source offset for each class. I'm trying to find a solution.

PR #4669, opened 2024-08-08
fix: should try casting dynamic slot name into constant

### LINKED ISSUE #4668: Dynamic slot name Template error: Type 'QTableSlots' has no matching index signature for type 'string'.ts-plugin(2537)

2.0.28

### VSCode version

1.91.1

### Vue version

3.4.29

### TypeScript version

5.4.5

### System Info

_No response_

### Steps to reproduce

As of version 2.0.28 of Vue Official, we get the following syntax highlighting if we use a dynamic slot name. In this case it is for QTableSlots of the Quasar Component Library.
```vue
<template #[`body-cell-${actionsKey}`]="data">
```

image

### What is expected?

No syntax highlighting as described in https://vuejs.org/guide/components/slots.html#dynamic-slot-names

Syntax highlighting and error in Problems window. When using Vue Official 2.0.26 this error does not occur.

_No response_

### Any additional comments?

_No response_

KermanX:
fix #4668.

Introduced a new helper:

```ts
function __VLS_tryAsConstant<const T>(t: T): T;
```

which is used to infer the type of an arbitrary expression like `as const`, but not throwing errors when this can't be done. For example, `("1" + "2") as const` will be an error, but `__VLS_tryAsConstant("1" + "2")` will just be of type `string`.

btw, I think `__VLS_tryAsConstant` can be used in other places to provide more precise type info - but it may cause performance issues.

### DIFF
--- a/packages/language-core/lib/codegen/script/globalTypes.ts
+++ b/packages/language-core/lib/codegen/script/globalTypes.ts
@@ -100,6 +100,7 @@ declare global {
 		: false;
 
 	function __VLS_normalizeSlot<S>(s: S): S extends () => infer R ? (props: {}) => R : S;
+	function __VLS_tryAsConstant<const T>(t: T): T;
 
 	/**
 	 * emit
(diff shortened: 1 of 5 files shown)


davidmatter:
Nice work

johnsoncodehk:
Thanks!

PR #4671, opened 2024-08-08
fix(language-service): drop unnecessary completion of event

### LINKED ISSUE #4670: Unnecessary completion of event

2.0.28

### VSCode version

1.92.0-insiders

### Vue version

3.5.0-alpha.5

### TypeScript version

5.5.4

### System Info

_No response_

### Steps to reproduce

Type some event name on any element.

### What is expected?

provide completion as `:onEvent`, `v-on:event` and `v-bind:onEvent`

The provided completion includes `onEvent` that cannot be of type `string`

_No response_

### Any additional comments?

_No response_

KazariEX:
fix #4670

### DIFF
--- a/packages/language-service/lib/plugins/vue-template.ts
+++ b/packages/language-service/lib/plugins/vue-template.ts
@@ -555,7 +555,9 @@ export function create(
 								const isGlobal = !propsSet.has(prop);
 								const name = casing.attr === AttrNameCasing.Camel ? prop : hyphenateAttr(prop);
 
-								if (hyphenateAttr(name).startsWith('on-')) {
+								const isEvent = hyphenateAttr(name).startsWith('on-');
+
+								if (isEvent) {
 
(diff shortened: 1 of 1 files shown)


johnsoncodehk:
@KazariEX Could you add a minimal test case?

KazariEX:
I have revised this PR and now it will do the following in the completion:

- Remove events in the formats of `onX`, `:onX` and `v-bind:onX`. IMO the standard practice for using component events should be to only use `@` or `v-on:`
- Consider `class`, `is`, `key`, `ref` and `style` as special props of all elements and components, and arrange them in the following order at the beginning of the completion list:
	1. props
	2. specialProps
	3. :props
	4. :specialProps
	5. @events
	6. v-bind:props
	7. v-bind:specialProps
	8. v-on:events
- Restore the display of the `vnode` events and rank it at the end of the list with the lowest priority.

PR #4674, opened 2024-08-09
feat(language-core): rename and navigation support for css module name

### LINKED ISSUE #4673: Rename and navigation support for css module name

Renaming CSS objects does not take effect on module name when using CSS Modules.

As the title suggests.

KazariEX:
resolve #4673


KazariEX:
Now it can do the following things:

image

image

Issue #4677, opened 2024-08-11
Renaming a class with multiple class names fails


L33Z22L11:
### Vue - Official extension or vue-tsc version

v2.0.28

### VSCode version

1.92.0

### Vue version

3.4.36

### TypeScript version

5.5.4

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (16) x64 AMD Ryzen 7 5800H with Radeon Graphics
    Memory: 2.62 GB / 15.86 GB
  Binaries:
    Node: 22.5.1 - C:\Program Files\nodejs\node.EXE
    npm: 10.8.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.7.0 - ~\AppData\Local\pnpm\pnpm.CMD
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.22621.3527
```

### Steps to reproduce

```vue
<template>
    <div class="class1 classname" />
</template>

<style scoped>
.class1 {}
</style>
```

1. Rename `class1` to `class12`.

### What is expected?

- `class1` is renamed to `class 12`.

### What is actually happening?

- The whole class attribute is overwritten to `class122`.

### Any additional comments?

Only when the length of the new class name is longer than that of the old one is renaming permitted.

PR #4681, opened 2024-08-12
fix(language-service): ignore css rename when use typescript-semanitc rename

linghaoSu:
fix https://github.com/vuejs/language-tools/issues/4677

When performing a rename, if the rename action is done on the name itself rather than on the . (dot), it triggers both the typescript-semantic and css plugins. However, if the rename is performed on the . (dot),  only the css plugin's rename operation is triggered.

1. When both typescript-semantic and css plugins are triggered:

The typescript-semantic plugin correctly generates the modification based on the TypeScript files generated by Vue.

On the other hand, the css plugin analyzes the CSS block and extracts the range including the . for the rename operation. When we trigger the rename, the content we input does not include the ., but the target range for replacement does include it, leading to a misalignment issue.

```vue

  .foo
// typescript-semantic
->.foo1

// css
      ! This is the leftover content from the TypeScript plugin above
->foo11
  ^^^^
   Generated by the CSS plugin
```

2. When only the css plugin is triggered:

In this case, the input needs to include the entire selector, including the .. However, triggering the rename only for the CSS part does not update the class names in the template area, so the desired effect is not achieved.

### DIFF
--- a/packages/language-service/lib/plugins/css.ts
+++ b/packages/language-service/lib/plugins/css.ts
@@ -9,6 +9,9 @@ export function create(): LanguageServicePlugin {
 			const baseInstance = base.create(context);
 			return {
 				...baseInstance,
+				async provideRenameEdits() {
+					return undefined;
+				},
 				async provideDiagnostics(document, token) {
 					let diagnostics = await baseInstance.provideDiagnostics?.(document, token) ?? [];
 					if (document.languageId === 'postcss') {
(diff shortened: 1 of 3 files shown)


linghaoSu:
I'm not entirely sure if this is the best approach, but it seems to resolve the issue effectively.

KazariEX:
This solves another problem: renaming variables in v-bind does not take effect on the source definition.

PR #4688, opened 2024-08-13
fix(language-core): generate `ref` as identifier instead of interpolation

KazariEX:
When using some characters like `-` as identifiers for `ref`, the incorrect virtual code will be generated.


johnsoncodehk:
LGTM, thanks!

PR #4689, opened 2024-08-13
fix(language-core): correct type narrowing from script to template

### LINKED ISSUE #4391: Type assertions in setup are not narrowed in template
After upgrading to vue-tsc 2.0.19 previous code no longer works:

```ts
const value: number | undefined = 1

if (!value) {
	throw new Error('Undefined')
}
```
```
<Component must-be-number={value} />
```
I get error
```
'__VLS_ctx.value' is possibly 'undefined'.
```

Verified working version 2.0.7

KazariEX:
fix #4391


johnsoncodehk:
LGTM, thanks!

PR #4691, opened 2024-08-14
fix(language-core): should camelize prop name in `experimentalModelPropName`

### LINKED ISSUE #4414: vue-tsc 2.0.18/19: Cannot find name 'xyz' property when  using experimentalModelPropName config
Hello, I am using the vuetify2-component-types dev dependency.
There is a recommended setting to configure where vuetify 2 deviates from the standard value/input configuration for v-model:

        "experimentalModelPropName": {
            "input-value": {
                "v-checkbox": true,
                "v-switch": true,
                "v-chip": true,
                "v-btn": true,
                "v-list-item": true,
                "v-bottom-navigation": true
            },
            "": { "input": true },
            "value": {
                "input": { "type": "text" },
                "textarea": true,
                "select": true
            }

Pre 2.0.18, this was working as intended. Now, in 2.0.18/19 I am getting error messages from vue-tsc
image

I can fix this a bit by putting v-model as the last thing in the template
image

But still, the tooltip descriptions / intellisense info are missing. Only, if I would remove v-model completely, they are currently appearing:
image

KermanX:
fix #4414.


johnsoncodehk:
Thanks!

PR #4715, opened 2024-08-23
fix(typescript-plugin): unknown request type warning

davidmatter:
This one ensures that

```typescript
console.warn('[Vue Named Pipe Server] Unknown request type:', request.type);
```

is not falsely executed for existing request types.


johnsoncodehk:
LGTM, thanks!

PR #4722, opened 2024-08-25
chore: add .DS_Store to .gitignore

davidmatter:
https://en.wikipedia.org/wiki/.DS_Store


johnsoncodehk:
Thanks for this PR, but I think .DS_Store should be configured by Mac developers in global .gitignore rather than in project space.

PR #4765, opened 2024-08-30
fix: Whitelist ms-dynamics-smb.al extension for Vue Hybrid Mode.

kyleweishaupt:
The Microsoft AL Language Extension for Microsoft Dynamics 365 Business shows up as potentially incompatible when starting VS Code and opening .vue files. These are unrelated to to the AL language plugin. 

Changing hybrid mode to true does not cause any interference with Vue files as far as I can tell. 

image


johnsoncodehk:
Thanks!

PR #4768, opened 2024-08-31
fix(language-core): prevent circular reference of templateRef

zhiyuanzmj:


test-workspace/tsc/passedFixtures/vue3.5/templateRef/template-ref.vue
KazariEX, line 1: I think it's best to keep consistency in code style.

packages/language-core/lib/codegen/script/template.ts
```diff
@@ -257,3 +261,15 @@ export function getTemplateUsageVars(options: ScriptCodegenOptions, ctx: ScriptC
 
 	return usageVars;
 }
+
+function getRefsType(options: ScriptCodegenOptions, scriptSetupRanges: ScriptSetupRanges) {
```
KazariEX, line 265: Can we use the generator here?

zhiyuanzmj: I tried. But `const exps = []` is a string[] type, Can you try it?

KazariEX: I found that `generateTemplateCtx` also needs to be converted into a pure generator-based form. In this case, its readability doesn't seem to differ much from using array operations :)

KazariEX (approved):

johnsoncodehk:
Thanks!

Issue #4583, opened 2024-07-12
End of file breaking autocompletion


unpseudocomplique:
### Vue - Official extension or vue-tsc version

2.0.26

### VSCode version

1.91.0

### Vue version

3.4.0

### TypeScript version

idk how to get it

### System Info

```shell
System:
    OS: Windows 11 10.0.22631
    CPU: (16) x64 AMD Ryzen 7 5800X 8-Core Processor
    Memory: 19.45 GB / 47.94 GB
  Binaries:
    Node: 20.2.0 - C:\Program Files\nodejs\node.EXE
    Yarn: 1.22.19 - C:\Program Files\nodejs\yarn.CMD
    npm: 9.6.6 - C:\Program Files\nodejs\npm.CMD
    pnpm: 9.4.0 - C:\Program Files\nodejs\pnpm.CMD
    bun: 1.1.17 - ~\.bun\bin\bun.EXE
  Browsers:
    Chrome: 126.0.6478.127
    Edge: Chromium (126.0.2592.87), ChromiumDev (127.0.2610.3)
    Internet Explorer: 11.0.22621.3527
```

### Steps to reproduce

having an unexpected end of file break the autocompletion forever.
Like having a component without end tag, the only solution is to do a reload window.

It's also annoying because sometimes i want volar to import my component but because of the end of file it break before importing.

![image](https://github.com/user-attachments/assets/5073818d-a434-44a6-b2bf-60a8ca85af34)

### What is expected?

Autocompletion should still work even if an issue is present on the component

### What is actually happening?

Autocompletion is break and won't come back even when component is fixed !

### Link to minimal reproduction

[...]

PR #4771, opened 2024-08-31
fix(language-core): auto-completion for the last line of template block

zhiyuanzmj:
close: #4583


johnsoncodehk:
Great work, thanks!

PR #4795, opened 2024-09-03
fix(language-core): make `expose` of non-generic template ref required

zhiyuanzmj:
Should be required when is not a generic component.
image


packages/language-core/lib/codegen/template/element.ts
```diff
@@ -229,7 +229,7 @@ export function* generateComponent(
 		ctx.templateRefs.set(refName, [varName, offset!]);
 		ctx.usedComponentCtxVars.add(var_defineComponentCtx);
 
-		yield `var ${varName} = {} as (Parameters<typeof ${var_defineComponentCtx}['expose']>[0] | null)`;
+		yield `var ${varName} = {} as (Parameters<Required<typeof ${var_defineComponentCtx}>['expose']>[0] | null)`;
```
KazariEX, line 232:
```suggestion
		yield `var ${varName} = {} as (Parameters<NonNullable<typeof ${var_defineComponentCtx}['expose']>>[0] | null)`;
```

I guess this way is more efficient.

KazariEX (approved):

zhiyuanzmj:
> Can you add a test case?

It can correct infer type, Only affecting twoslash.
Such as this:
image

johnsoncodehk:
I see, thanks!

PR #4819, opened 2024-09-05
feat(language-core): enhance single root nodes collection

KazariEX:
This is an enhancement which enabling it to collect all single root nodes under the template branch, and selectively collect single root child node through specific components.

It will also infer the type of `$el` as the union of all single root nodes.


KazariEX:
There is an upstream issue I caused that needs to be fixed :)

PR #4855, opened 2024-09-18
fix(language-service): initialize scope with null prototype object

KazariEX:
Prevent mistakenly identifying properties such as `toString` inherited from `Object` as props.

### DIFF
--- a/packages/language-service/lib/plugins/vue-inlayhints.ts
+++ b/packages/language-service/lib/plugins/vue-inlayhints.ts
@@ -107,7 +107,7 @@ export function findDestructuredProps(
 	ast: ts.SourceFile,
 	props: Set<string>
 ) {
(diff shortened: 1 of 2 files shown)


packages/language-service/lib/plugins/vue-inlayhints.ts
```diff
@@ -107,7 +107,7 @@ export function findDestructuredProps(
 	ast: ts.SourceFile,
 	props: Set<string>
 ) {
-	const rootScope: Scope = {};
+	const rootScope: Scope = Object.create(null);
```
davidmatter, line 110: Can we test this somehow?

davidmatter (approved):
LGTM

zhiyuanzmj (approved):

PR #4864, opened 2024-09-21
fix(language-core): generate correct reference for `v-on` on `<slot>`

### LINKED ISSUE #4862: Slot doesn't reference variables correctly when registering events with @xxx syntax

2.1.6

### VSCode version

1.93.1

### Vue version

3.5.7

### TypeScript version

5.6.2

### System Info

_No response_

### package.json dependencies

```json
"dependencies": {
  "typescript": "^5.6.2",
  "vue": "^3.5.7"
}
```

### Steps to reproduce

```
<script setup lang="ts">
const foo = 'foo'
const bar = () => {}
</script>

<template>
    <slot :foo="foo" @bar="bar"></slot>
</template>
```

### What is expected?

Variable references in the generated code should be aligned with those in the template code.

image

_No response_

### Any additional comments?

_No response_

KazariEX:
fix #4862

### DIFF
--- a/packages/language-core/lib/codegen/template/elementProps.ts
+++ b/packages/language-core/lib/codegen/template/elementProps.ts
@@ -24,8 +24,7 @@ export function* generateElementProps(
 		suffix: string;
 	}[]
 ): Generator<Code> {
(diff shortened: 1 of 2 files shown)


davidmatter (approved):
LGTM

PR #4870, opened 2024-09-23
feat(language-core): auto infer `$event.currentTarget` type

KazariEX:
I think this is not a very good implementation, it would be great if we could introduce `EventHandler` similar to Solid in core.


KazariEX:
I think this is not a very good implementation, it would be great if we could introduce `EventHandler` similar to Solid in core.

PR #4871, opened 2024-09-23
chore: remove side effects

vikingair:
This would allow bundlers to tree shake the module code when imports are not done via `import type`.

E.g., 

```ts
import { ComponentProps } from 'vue-component-type-helpers';
```

Also, I would like to know why the `code` needs to be default exported as a string containing the same TS code again. Is it maybe some artifact from the past and or only relevant to Vue2?

### DIFF
--- a/packages/component-meta/package.json
+++ b/packages/component-meta/package.json
@@ -6,6 +6,7 @@
 		"**/*.js",
 		"**/*.d.ts"
 	],
+	"sideEffects": false,
 	"repository": {
 		"type": "git",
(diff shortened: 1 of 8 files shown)


KazariEX (approved):

johnsoncodehk:
LGTM, Thanks!

> Also, I would like to know why the `code` needs to be default exported as a string containing the same TS code again. Is it maybe some artifact from the past and or only relevant to Vue2?

That's because the code snippet needs to be used as virtual code in vue-component-meta.

PR #4878, opened 2024-09-25
fix(language-core): should not make `$props` optional

### LINKED ISSUE #4876: JS defineProps() does not mark props as required

2.1.6

### VSCode version

1.93.1

### Vue version

3.5.8

### TypeScript version

5.5.4

### System Info

```shell
System:
    OS: Linux 6.8 Ubuntu 24.04.1 LTS 24.04.1 LTS (Noble Numbat)
    CPU: (6) arm64 unknown
    Memory: 16.18 GB / 22.27 GB
    Container: Yes
    Shell: 5.2.21 - /bin/bash
  Binaries:
    Node: 20.17.0 - ~/.nvm/versions/node/v20.17.0/bin/node
    npm: 10.8.2 - ~/.nvm/versions/node/v20.17.0/bin/npm
  Browsers:
    Brave Browser: 128.1.69.168
```

### package.json dependencies

```json
"dependencies": {
    "vue": "^3.4.29"
  },
  "devDependencies": {
    "@vitejs/plugin-vue": "^5.0.5",
    "vite": "^5.3.1"
  }
```

### Steps to reproduce

JS (*not* TS) `defineProps` with a required prop:

```vue
<script setup>
defineProps({
  msg: {
    type: String,
    required: true
  }
})
</script>
```
In another component, import this component and use it in the template.

### What is expected?

The prop cannot be omitted.

The prop is falsely marked as optional.

This image is from the repro repo:

image

When you define the component globally via `components.d.ts`, the typing becomes more obvious:

image

There's no error for the missing `required_prop` on the left. When I add `lang="ts"` to the script tag on the right, the error appears.

https://github.com/phil294/vue-bug-repo

### Any additional comments?

When you define props as

```vue
<script>
export default {
  props:{
    msg: {
      type: String,
      required: true
    }
  },
  setup() {
  }
}
</script>
```
nothing works at all. This was already reported in https://github.com/vuejs/language-tools/issues/4074  and is allegedly because of https://github.com/vuejs/language-tools/issues/3690 (which to me it doesn't seem like it but idk) which is labelled as a "minor" bug. In my opinion this isn't minor as arguably the main point of Volar, IntelliSense in Vue templates, is pretty much unusable if you're not using TS :-S

KazariEX:
fix #4876


davidmatter (dismissed):
LGTM

packages/language-core/lib/codegen/script/component.ts
```diff
@@ -52,8 +52,7 @@ export function* generateComponent(
 export function* generateComponentSetupReturns(scriptSetupRanges: ScriptSetupRanges): Generator<Code> {
 	// fill $props
 	if (scriptSetupRanges.props.define) {
-		// NOTE: defineProps is inaccurate for $props
-		yield `$props: __VLS_makeOptional(${scriptSetupRanges.props.name ?? `__VLS_props`}),${newLine}`;
+		yield `$props: ${scriptSetupRanges.props.name ?? `__VLS_props`},${newLine}`;
```
rchl, line 55: Not sure if this is correct in general.

Try with a component with props:

```
  msg: {
    type: String,
    default: null
  }
```

In that case the prop should be optional but won't.

I feel like that might need coordination with changes in Vue types.

johnsoncodehk: I just add a test case in https://github.com/vuejs/language-tools/pull/4878/commits/3557d8b17a9e5e78c18013262823649410effada for this.

johnsoncodehk:
Closed due to not being a suitable solution. (making https://github.com/vuejs/language-tools/blob/master/test-workspace/tsc/passedFixtures/vue3/%234878/main.vue test fail)

PR #4880, opened 2024-09-25
feat(language-core): split `strictTemplates`

### LINKED ISSUE #4879: split `strictTemplate` into two distinct options for props and component import

First, thanks again for your amazing work on vue-tsc, nothing is greater than to be type safe in a Vue environnement.
In our codebase we want to gradually make component imports safer, and currently an undeclared Component is marked as `unknown`.
Activating `strictTemplates: true` resolve this, but also activates the strict prop typechecking, which we currently can't handle in our codebase. 
We are then currently blocked on this subject as we would like to gradually upgrade a safer props types but will be too much work for now.

The solution could be to split the `strictTemplate` into two options, or to add a two new ones `strictProps` and `strictComponentImports`, to cherry pick which one we want.

KazariEX:
resolve #4879 

```ts
const strictTemplates = typeof vueOptions.strictTemplates === 'boolean' ? {
	attributes: vueOptions.strictTemplates,
	components: vueOptions.strictTemplates
} : vueOptions.strictTemplates ?? {
	attributes: false,
	components: false
};
return {
	strictTemplates
};
```


rchl:
There is also schema: https://github.com/vuejs/language-tools/blob/master/packages/language-core/schemas/vue-tsconfig.schema.json

Ashishsonavane (approved):

PR #4897, opened 2024-09-30
chore: remove `importsNotUsedAsValues`

KazariEX:


davidmatter (approved):
Should we eventually aim for supporting `verbatimModuleSyntax=true`? I've looked into it and there would be hundreds of necessary changes.

KazariEX:
It seems that these changes are not destructive, can try doing it this way.

johnsoncodehk:
> Should we eventually aim for supporting `verbatimModuleSyntax=true`? I've looked into it and there would be hundreds of necessary changes.

I've tried several times and the current conclusion is that verbatimModuleSyntax is not available for CJS projects.

As an alternative to importsNotUsedAsValues, we need to use the equivalent ESLint rule in tsslint.config.ts.

PR #4902, opened 2024-10-02
fix(language-core): should wrap item with `Reactive` on `v-for`

KazariEX:
fix vuejs/core#12095


bmulholland:
In case anyone else comes here to check, Github says this PR "first appeared in v2.1.8." It does not, and has not been released yet.

PR #4907, opened 2024-10-04
fix(language-core): correctly obtain the index of style modules

KazariEX:


packages/language-core/lib/codegen/script/styleModulesType.ts
```diff
@@ -8,13 +8,12 @@ export function* generateStyleModulesType(
 	options: ScriptCodegenOptions,
 	ctx: ScriptCodegenContext
 ): Generator<Code> {
-	const styles = options.sfc.styles.filter(style => style.module);
+	const styles = options.sfc.styles.map((style, i) => [style, i] as const).filter(([style]) => style.module);
```
Gehbt, line 11: use `Array.reduce` can eliminate one layer of looping.  ~~the type needs manual assign.~~
```suggestion
	const styles = options.sfc.styles.reduce((accumulator, style, i) => {
		if (style.module) {
			accumulator.push([style, i]);
		}
		return accumulator;
	}, []);
```

KazariEX: This way indeed requires manually specifying the type, the code will actually become more complex.

PR #4908, opened 2024-10-05
fix(language-core): correctly infer union props to options

KazariEX:
fix vuejs/core#12012 , fix vuejs/core#12110

### DIFF
--- a/packages/language-core/lib/codegen/globalTypes.ts
+++ b/packages/language-core/lib/codegen/globalTypes.ts
@@ -87,6 +87,7 @@ export function generateGlobalTypes(lib: string, target: number, strictTemplates
 		'__ctx' extends keyof __VLS_PickNotAny<K, {}> ? K extends { __ctx?: infer Ctx } ? Ctx : never : any
 		, T extends (props: any, ctx: infer Ctx) => any ? Ctx : any
 	>>;
+	type __VLS_IsUnion<T, U = T> = T extends U ? [U] extends [T] ? false : true : false;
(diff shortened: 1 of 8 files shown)


darkbasic:
Thanks, I've tested your patch and it fixed my issue.

davidmatter (approved):

hendrikheil:
@KazariEX I think this would be a great addition. Currently there are typescript errors when you do the following:

```
<script setup lang="ts">
import { useCssModule } from 'vue';

const styles = useCssModule();

defineProps<
  | {
      option: 'a';
      aValue: string;
    }
  | {
      option: 'b';
      bValue: number;
    }
>();
</script>

<template>
  <div :class="styles.example">Option: {{ option }}</div>
</template>

<style lang="css" module>
.example {
  color: red;
}
</style>
```

As soon as you add the style module tag, vue and typescript are unable to properly see props and things like imported components. I tested your code locally and it seemed to have made the typescript errors in the template tag go away immediately. Do you think it makes sense to reconsider this PR?

KazariEX:
According to Johnson's suggestion, a better approach is to place union types into generics to skip redundant type inferring and directly use the original types, which avoids injecting too much difficult to maintain code.

darkbasic:
There are many issues with that approach. I started writing a comment about that but then I stumbled upon an even bigger bug regarding union props and unfortunately I still have to file it.

hendrikheil:
I did try the generics approach as well, but that does cause other issues for me as well. I'll try to create a minimal component reproduction of the errors I'm encountering with it and post it here.

PR #4930, opened 2024-10-18
fix(component-meta): error when signatures is undefined

Hannesrasmussen:
In certain cases, `signatures[0].parameters[0]` returns undefined. We need to prevent this from causing an error.

### DIFF
--- a/packages/component-meta/lib/base.ts
+++ b/packages/component-meta/lib/base.ts
@@ -540,7 +540,7 @@ function createSchemaResolvers(
 	function resolveSlotProperties(prop: ts.Symbol): SlotMeta {
 		const propType = typeChecker.getNonNullableType(typeChecker.getTypeOfSymbolAtLocation(prop, symbolNode));
 		const signatures = propType.getCallSignatures();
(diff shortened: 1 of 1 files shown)


packages/component-meta/lib/base.ts
```diff
@@ -540,7 +540,7 @@ function createSchemaResolvers(
 	function resolveSlotProperties(prop: ts.Symbol): SlotMeta {
 		const propType = typeChecker.getNonNullableType(typeChecker.getTypeOfSymbolAtLocation(prop, symbolNode));
 		const signatures = propType.getCallSignatures();
-		const paramType = signatures[0].parameters[0];
+		const paramType = signatures.length > 0 ? signatures[0].parameters[0] : [];
```
KazariEX, line 543:
```suggestion
		const paramType = signatures[0]?.parameters[0];
```

KazariEX (changes requested):

KazariEX (approved):

johnsoncodehk (approved):

PR #4971, opened 2024-10-31
feat(language-core): support `@vue-generic`

KazariEX:
resolve vuejs/core#8015

### DIFF
--- a/test-workspace/tsc/passedFixtures/vue3/v-generic/main.vue
+++ b/test-workspace/tsc/passedFixtures/vue3/v-generic/main.vue
@@ -4,8 +4,11 @@ import Comp from './comp.vue';
 </script>
 
 <template>
-    <Comp v-generic="number" @foo="(val) => exactType(val, {} as number)"/>
-    <Comp v-generic="string" @foo="(val) => exactType(val, {} as string)"/>
+    <Comp v-generic="string" @foo="(bar) => exactType(bar, {} as string)"/>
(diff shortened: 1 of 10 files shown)


test-workspace/tsc/passedFixtures/vue3/v-generic/main.vue
```diff
@@ -0,0 +1,11 @@
+<script setup lang="ts">
+import { exactType } from '../../shared';
+import Comp from './comp.vue';
+</script>
+
+<template>
+    <Comp v-generic="number" @foo="(val) => exactType(val, {} as number)"/>
```
xiaoxiangmoe, line 7: Maybe we should also add a example for multiple generic params

johnsoncodehk:
I think writing types in `v-generic` is wrong in Vue's specification, according to the documentation custom directives only receive JS statements, the `v-generic` convention may break compatibility with other tools/libraries.

We should use `generic`, or convention to pass type parameters in HTML comment through `<!-- @vue-generic {xxx} -->`.

johnsoncodehk:
@KazariEX I see, but this will still break Vue versions prior to https://github.com/vuejs/core/pull/12301, especially since Vue 2 has stopped being maintained. Since this is a language tools specific feature, we need an implementation that doesn't have to change core.

KazariEX:
Can we consider it as a feature of the next minor version?

PR #4175, opened 2024-03-29
feat(component-meta): add typing resolution for defineModel modifiers

stafyniaksacha:
This PR aims to retrieve types of model modifiers from `defineModel` macro.

```ts
// default
const [modelValue, modelModifiers] = defineModel<string, 'lazy' | 'trim'>()

// modelValue: import("vue").PropType<string>
// modelModifiers: import("vue").PropType<Record<'lazy' | 'trim', true>>


// named
const [named, modifiers] = defineModel<string, 'foo'>('named')

// named: import("vue").PropType<string>
// namedModifiers: import("vue").PropType<Record<'foo', true>>
```

Here is a playground where we can see property being generated: https://play.vuejs.org/#eNqFkk1PwzAMhv+KlUtBqrYDnKZuEqAdQOJDgLgQDlHrlow0iZJ0DEr/O26qlQlN2y1+/dp+bKVlF9ZO1g2yGct87qQN4DE0FpTQ1Zyz4DlbcJ0b7QO81qZA9SJUgynE960pZCnR+TeYQ4Gl1EgSqswHJ3WVQqLE91cCP5CQUCeLk9OxmRY1FrHPkRalMVSYRH9C9dl0ICUuCgLWVomAFAFk1uGibaHtAxgY9/Kmg+E/Q5Q76Lps2nfqZ+30Zyndg+hLWU1W3mg6WhzEWW5qKxW6exskbcfZbIvAmVDKfN5ELTgi2er5O+Yfe/SV3/QaZw8OPbo1cjbmgnAVhiG9fLrDDb3HJG3RKHIfSD6iN6rpGQfbZaMLwt7xRdrr2hoX6PrPfrkJqP12qR40nij6OaOfc3Vg9T/cs8l5rOO6Y90vLZfTGA==

PR #4978, opened 2024-11-03
fix(language-core): make model modifiers optional

stafyniaksacha:
This pull request includes changes to improve the handling of component property modifiers in the `generateComponentProps` function and its associated tests.

It aims to better reflect `modelModifiers` type marking keys as optional:

```html
<!-- MyComp.vue -->
<script setup lang="ts">
const [modelValue, modelModifiers] = defineModel<string, 'foo' | 'bar'>()

// modelModifiers.foo?: true
// modelModifiers.bar?: true
</script>
```

With this update, we can omit keys when manually defining `model-modifiers` prop
```html
<template>
  <!-- Type '{ foo: true; }' is missing the following properties from type 'Record<"foo" | "bar", true>': bar -->
  <MyComp :model-modifiers="{ foo: true }" />
</template>
```

Note that it will also fix type when using `vitest` and `@vue/test-utils`:
```ts
import { MyComp } from './MyComp.vue'
import { mount } from '@vue/test-utils'
import { it } from 'vitest'

it('should foo modifier', async () => {
  const component = mount(MyComp, {
    props: {
	  // Type '{ foo: true; }' is missing the following properties from type 'Record<"foo" | "bar", true>': bar
      'modelModifiers': { foo: true },
    },
  })
})
```

### DIFF
--- a/packages/component-meta/tests/index.spec.ts
+++ b/packages/component-meta/tests/index.spec.ts
@@ -39,6 +39,8 @@ const worker = (checker: ComponentMetaChecker, withTsconfig: boolean) => describ
 		expect(bar).toBeDefined();
 		expect(qux).toBeDefined();
 		expect(quxModifiers).toBeDefined();
+		// @todo: order vary depending on withTsconfig option
+		expect(quxModifiers?.type).toMatch(/Partial<Record<(?:"lazy" \| "trim"|"trim" \| "lazy"), true>> \| undefined/);
 		expect(onUpdateFoo).toBeDefined();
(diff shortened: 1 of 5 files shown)


KazariEX:
Could you please add a test case?

stafyniaksacha:
> Could you please add a test case?

Yes sure, where should I put them ? I only contributed to components-meta env (time ago https://github.com/vuejs/language-tools/pull/4175)

KazariEX:
Added.

KazariEX (approved):

johnsoncodehk (approved):

PR #4990, opened 2024-11-06
chore(vscode): use tsup for tree-shaking

zhiyuanzmj:


johnsoncodehk:
It seems that `dist/client.js` does not bundle the modules it depends on (`reactive-vscode`, `@volar/vscode`...)

zhiyuanzmj:
Oh, I was wrong. The PR can't tree-shaking. It's only use require instead of build-in bundle. But the VSCode plugin don't have node_modules. So I closed the PR.

johnsoncodehk:
No problem, thank you for exploring!

PR #4994, opened 2024-11-07
feat(build): integrate tree-shaker for improved bundle optimization

johnsoncodehk:
Tree shaking power by https://github.com/KermanX/tree-shaker.

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -555,6 +555,7 @@
 		"postinstall": "vscode-ext-gen --scope vue"
 	},
 	"devDependencies": {
+		"@kermanx/tree-shaker": "https://pkg.pr.new/KermanX/tree-shaker/@kermanx/tree-shaker@67a133e",
 		"@types/semver": "^7.5.3",
 		"@types/vscode": "^1.82.0",
 		"@volar/vscode": "~2.4.9",
--- a/extensions/vscode/scripts/build.js
+++ b/extensions/vscode/scripts/build.js
@@ -1,6 +1,7 @@
 // @ts-check
 const path = require('path');
 const fs = require('fs');
+const treeShake = require('@kermanx/tree-shaker').treeShake;
 
 require('esbuild').context({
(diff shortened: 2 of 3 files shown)


Ashishsonavane (approved):

johnsoncodehk:
@KermanX I think this PR is almost successful, but tree-shaker seems to have some template string/regex conversion problems, can you take a look?

KermanX:
> @KermanX I think this PR is almost successful, but tree-shaker seems to have some template string/regex conversion problems, can you take a look?

Yes, I've noticed that bug, but I can't immediately solve it because I need to investigate how to escape the template literal in oxc 🥲.

KermanX:
I've used a workaround to solve the template literal escaping issue.

Currently, this PR is tree-shaking thousands of input files, which may lose many import chances compared to tree-shaking the bundled js.

I found that the `server.cjs` produced by esbuild will make the tree shaker stuck for a long time. However, tree-shaking works well on the Rollup output in #4998.

johnsoncodehk:
Unfortunately the output file is currently corrupt and still needs to be investigated.

#4998 changes the project dependencies too much, so it's relatively difficult to merge (and I haven't finished reviewing all the changes yet), and if we can successfully merge this PR first, and move on to #4998 in the future, I think there is no harm.

KermanX:
> Unfortunately the output file is currently corrupt and still needs to be investigated.

I've fixed another bug. And I can't find more problems myself now.

KermanX:
I just found that the filter used to be `/\.js/`, which accepts `*.json`. And tree-shaker can't parse JSON, which results in an empty output 🥲

Sadly, after this fix, tree-shaker can only make the bundle size less than 0.01 MB smaller. This is quite expected because tree-shaker aims to remove unused code, but individual source modules always don't have useless code themselves. I will work on making tree-shaking the bundled code work.

johnsoncodehk:
Due to the switch to rolldown, this PR is obsolete.

PR #4998, opened 2024-11-08
chore(vscode): use `unjs/unbuild` for bundling

kermanx:
Use https://github.com/unjs/unbuild to bundle the VSCode extension, also as a preparation for #4994.

#### Notes:
- To dev the extension: run `pnpm dev:vue`, the command will exit after the stub files generated (no need to watch file changes), then F5. (Should the tasks defined in `.vscode/tasks.json` and `.vscode/launch.json` be restored?)
- To build the extension: like before.
- Other packages except the VSCode extension remain unchanged.


sxzz:
Hey, what do you think about switching to [tsdown](https://github.com/sxzz/tsdown), powered by Rolldown, for faster speed?

extensions/vscode/package.json
```diff
@@ -573,7 +574,9 @@
 		"esbuild": "latest",
 		"esbuild-visualizer": "latest",
 		"reactive-vscode": "^0.2.9",
+		"rollup-plugin-visualizer": "^5.12.0",
```
filipsobol, line 577: I noticed that you've replaced `sonda` with `rollup-plugin-visualizer`. I will not try to convince you to use `sonda`, but if this was a technical decision, could you share the reason? I would appreciate your feedback so I can improve it.

KermanX: @johnsoncodehk uwu

johnsoncodehk: Hi @filipsobol, first of all thank you for developing sonda. One of the reasons is that we want the PR to keep the most relevant changes so that it has a greater chance of being merged. Another reason is that the output file treemap is paginated in Sonda, and we cannot visually check whether the modules that need to be shared by multiple files have been correctly split.

filipsobol: Thank you, this gives me some interesting ideas :thinking: 

> Another reason is that the output file treemap is paginated in Sonda, and we cannot visually check whether the modules that need to be shared by multiple files have been correctly split.

By visual check, do you mean manual or automated visual check? You should be able to switch between chunks using the dropdown menu in the upper-left corner, but I can imagine that it may require extra work or even problems in automated tests (e.g. if it visually compares two reports).

PR #5011, opened 2024-11-17
feat(language-core): type support for `useAttrs` and `$attrs`

### LINKED ISSUE #5008: ``$attrs`` or ``ReturnType<typeof useAttrs()>`` is always ``Data``when ``fallthroughAttributes: true``

2.1.10

### VSCode version

1.95.3

### Vue version

3.5.12

### TypeScript version

5.6.3

### System Info

_No response_

### package.json dependencies

```json
"dependencies": {
  "vue": "3.5.12"
},
"devDependencies": {
  "@types/node": "22.9.0",
  "@vitejs/plugin-vue": "5.2.0",
  "typescript": "5.6.3",
  "vite": "5.4.11",
  "vue-tsc": "2.1.10"
}
```

### Steps to reproduce

1. Open the plaground link
2. Hover over ``attrs``: the type will be ``Data`` instead of ``ButtonHTMLAttributes``
```ts
const attrs = useAttrs();
```
3. Run ``npm run check``: You will see that

### What is expected?

The type of the fallthrough attributes should be `Data` when there is no `v-bind="$attrs"` assigned to an element. But when it is assigned, it should be of the assigned element's attributes type.

The type of the fallthrough attributes always is ``Data``, both on ``useAttrs`` and ``$attrs``.

https://stackblitz.com/edit/vue3-vite-typescript-starter-aahszk?file=src%2Fcomponents%2FVButton.vue

### Any additional comments?

_No response_

KazariEX:
resolve #5008


zhiyuanzmj (approved):

packages/language-core/lib/codegen/template/index.ts
```diff
@@ -91,11 +92,12 @@ function* generateSlotsType(options: TemplateCodegenOptions, ctx: TemplateCodege
 }
 
 function* generateInheritedAttrs(ctx: TemplateCodegenContext): Generator<Code> {
-	yield 'var __VLS_inheritedAttrs!: {}';
+	yield 'let __VLS_inheritedAttrs!: {}';
 	for (const varName of ctx.inheritedAttrVars) {
 		yield ` & typeof ${varName}`;
 	}
 	yield endOfLine;
+	yield `var $attrs!: Partial<typeof __VLS_inheritedAttrs> & Record<string, unknown>${endOfLine}`;
```
davidmatter, line 100: Is the `var` (rather than `let`) an oversight or does it need to be a var?

davidmatter (dismissed):

packages/language-core/lib/codegen/template/index.ts
KazariEX: Since it is generated at the end but used in the template, it requires `var` that can be declared as a hoisting variable.

PR #5043, opened 2024-12-11
fix(language-core): always use destructure syntax to pass props type

### LINKED ISSUE #1232: Default export of the module has or is using private name 'Props'
when upgrade volar to ^0.34.9 ,  there is a ts warning `Default export of the module has or is using private name 'Props'` at the top of  template

```js
//error info -> Default export of the module has or is using private name 'Props'. ts(4082)
<template>test</template>
<script lang="ts" setup>
  //step1: declear type
  interface Props {
    foo: string;
  }
 // setp2: defineProps
  defineProps<Props>(); 
</script>
```

but if definePorps like below, the warning disappear
```js
<template>test</template>
<script lang="ts" setup>
  defineProps<{
    foo: string;
  }>();
</script>
```

KazariEX:
re-fix #1232

### DIFF
--- a/packages/language-core/lib/codegen/script/component.ts
+++ b/packages/language-core/lib/codegen/script/component.ts
@@ -173,20 +173,14 @@ export function* generatePropsOption(
 	const useOption = !useTypeOption || scriptSetupRanges.props.withDefaults;
 
 	if (useTypeOption) {
-		if (codes.length === 1) {
-			yield `__typeProps: `;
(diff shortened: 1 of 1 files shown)


KermanX (approved):

johnsoncodehk:
Since we have discovered that destructuring can significantly impact type evaluation performance, we need to consider this solution carefully.

PR #5055, opened 2024-12-16
feat(language-core): type support of `useSlots` and `$slots`

KazariEX:


darkbasic:
image

It used to infer `string | number` before the 2.2.0 update while now I get this error.

darkbasic:
image

@KazariEX it used to infer `string | number` before the 2.2.0 update while now I get this error.

darkbasic:
@KazariEX thanks, I'll use `Record<string, unknown>` otherwise I get this error:

image

FaustineG:
Thank @KazariEX ! 

In my case, I didn't need to specify a type, simply `defineSlots();` removed the error. 

@darkbasic You could probably also use the `Slots` type from vue, it worked on another typing error for me.
image

KazariEX:
You can also use:

```vue
<template v-for="render in $slots">
  <component :is="render" />
</template>
```

instead of `<slot>`.

PR #5067, opened 2024-12-18
fix(language-core): resolve components with various name cases correctly

KazariEX:
Due to the kebab case component in the template causing its pascal case component name to also be accepted, remove the `__VLS__resolvedLocalAndGlobalComponents` that integrates all components (including components that will not actually be retrieved from it).

For example:

```vue
<script setup lang="ts">
import myComp from './my-comp.vue';
</script>

<template>
	<my-comp />
	<MyComp />
</template>
```

Due to `my-comp` being converted to pascal case and added into `__VLS_resolvedLocalAndGlobalComponents`, `MyComp` will be mistakenly considered as an existing component, however in fact `myComp` from the script block cannot be used in this way.

At the same time, it avoids a large amount of redundant information that may occur when displaying unknown components due to its type being unwrapped.

Before:

image

After:

image


johnsoncodehk:
For context, the purpose of `__VLS_resolvedLocalAndGlobalComponents` is to reuse the WithComponent type evaluation results for the same tags, but there is currently no evidence that `__VLS_resolvedLocalAndGlobalComponents` effectively improves performance, so let's merge it.

PR #5070, opened 2024-12-19
perf(typescript-plugin): use named pipe servers more efficiently

johnsoncodehk:
- [x] Named pipe server search number dropped from 40 to 20
- [x] Instead of re-doing the server connect every time for every request, a valid server connect is retained and reused for all requests.
- [x] containsFile results are now cached.
- [x] Server status polling interval changed from 1 second to 2.5 seconds
- [x] Pushed by the named pipe server when component names/props change rather than pulled by the client everytime

### DIFF
--- a/packages/typescript-plugin/lib/server.ts
+++ b/packages/typescript-plugin/lib/server.ts
@@ -65,7 +65,7 @@ export async function startNamedPipeServer(
 		getFileId: (fileName: string) => fileName,
 	};
 	const dataChunks: Buffer[] = [];
-	const componentNamesSubscriptions = new Map<string, [Awaited<ReturnType<typeof getComponentAndProps>>, Set<net.Socket>]>();
+	const componentNamesSubscriptions = new Map<string, [ReturnType<typeof getComponentAndProps> | Awaited<ReturnType<typeof getComponentAndProps>>, Set<net.Socket>]>();
 	const server = net.createServer(connection => {
 		connection.on('data', buffer => {
 			dataChunks.push(buffer);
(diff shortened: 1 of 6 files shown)


packages/typescript-plugin/lib/server.ts
```diff
@@ -21,10 +20,27 @@ export interface Request {
 	| 'getComponentProps'
 	| 'getComponentEvents'
 	| 'getTemplateContextProps'
-	| 'getComponentNames'
-	| 'getElementAttrs';
-	args: [fileName: string, ...rest: any];
-}
+	// | 'getComponentNames'
```
knakmitch, line 23: You left in commented code

packages/typescript-plugin/lib/server.ts
```diff
@@ -130,6 +104,155 @@ export async function startNamedPipeServer(
...
+						componentNamesSubscriptions.set(fileName, [newData, subscriptions]);
+						// Notify
+						for (const connection of connections) {
+							notify(connection, 'componentAndPropsUpdated', fileName, newData);
+						}
+					}
+				}
+			}
+		}
+	}
```
knakmitch, line 134: there is a high amount of nesting here. Would be a good idea to refactor.

packages/typescript-plugin/lib/utils.ts
```diff
@@ -3,211 +3,239 @@ import * as net from 'node:net';
...
+				resolve(data);
+			});
+			// setTimeout(() => {
+			// 	if (this.requestHandlers.has(seq)) {
+			// 		console.error(`[${seq}] ${requestType} ${fileName} timeout`);
+			// 		this.requestHandlers.delete(seq);
+			// 		resolve(undefined);
+			// 	}
+			// }, 5000);
+			this.socket!.write(JSON.stringify([seq, requestType, fileName, ...args] satisfies RequestData) + '\n\n');
```
knakmitch, line 171: lot of commented out code here

PR #5071, opened 2024-12-19
fix(language-core): generate macros after script setup content

KazariEX:


zhiyuanzmj:
Hi @KazariEX , When I update twoslash to volar 2.2.0,  I got a ts error:
image

declare expression can't using in function.

KazariEX:
It should not affect vue-tsc, can it be solved by adding `// @ts-ignore`?

zhiyuanzmj:
Ok it works well, let's do it.

PR #5101, opened 2024-12-30
feat(component-meta): collect destructured props defaults

Akryum:


packages/language-core/lib/codegen/utils/index.ts
```diff
@@ -41,15 +41,16 @@ export function collectVars(
 export function collectIdentifiers(
 	ts: typeof import('typescript'),
 	node: ts.Node,
-	results: [id: ts.Identifier, isRest: boolean][] = [],
-	isRest = false
+	results: [id: ts.Identifier, isRest: boolean, initializer: ts.Expression | undefined][] = [],
```
KazariEX, line 44: I think this parameter can be modified to object type.

PR #5136, opened 2025-01-20
feat: experimental resolve external stylesheets

### LINKED ISSUE #3505: vue style module does not prompt

image
I have a prompt for writing styles in the Vue file. But if you use src Imports，
image

style doesn't prompt anymore

### What does the proposed API look like?

How can external styles also have prompts?, If the cssmoodule of React introduces external files, it can also prompt

### LINKED ISSUE #3720: Suggestion: CSS @import statements intellisense
Right now I'm very satisfied with using CSS modules in single file components, but the intellisense doesn't seem to extend to imported modules:
```vue
<template>
    <div :class="$style['my-class']"></div>
</template>
<style module>
/* imagine this contains .my-class */
@import url("./styles/animation.css"); 

.some-class {
...
}
</style>
```
It works correctly and the language tools don't scream because `$style` is `Record<string, string> & { 'some-class': string }`, but it would be nice if the `$style` type takes static imports into consideration.

KazariEX:
resolve #3505, resolve #3720


johnsoncodehk:
I think we can detect if the current file is referencing external css, then suggest to the user to install css-modules-kit extension.

PR #5137, opened 2025-01-21
feat(language-core): type support of slot children

KazariEX:
image

### DIFF
--- a/packages/language-core/lib/codegen/globalTypes.ts
+++ b/packages/language-core/lib/codegen/globalTypes.ts
@@ -1,7 +1,7 @@
 import { getSlotsPropertyName } from '../utils/shared';
 
 export function generateGlobalTypes(lib: string, target: number, strictTemplates: boolean) {
-	const fnPropsType = `(K extends { $props: infer Props } ? Props : any)${strictTemplates ? '' : ' & Record<string, unknown>'}`;
+	const fnPropsType = `(T extends { $props: infer Props } ? Props : any)${strictTemplates ? '' : ' & Record<string, unknown>'}`;
 	let text = ``;
 	if (target < 3.5) {
 		text += `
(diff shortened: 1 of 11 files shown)


Anoesj:
Awesome!!

Anoesj:
If only text content is expected in a slot, how would one type that? I tried numerous things:
```ts
const slots = defineSlots<{
  default?: () => HTMLElement;
}>();

const slots = defineSlots<{
  default?: () => Text;
}>();

const slots = defineSlots<{
  default?: () => string;
}>();
```

I tried all of these separately, but text content in a slot is typed as `[]`, so all of these actually fail. Does `[]` indicate: no HTML elements or components? So is this the correct form? It works, but it does look a bit weird...
```ts
const slots = defineSlots<{
  default?: () => [];
}>();
```

KazariEX:
We currently do not generate virtual code for text nodes, so this cannot be type checked yet. You can indeed use `[]` to indicate that no native elements or components are accepted.

Anoesj:
Hmm that's too bad, because even though this is a very very welcome addition, it's not entirely *strict* that way. See, you can do this without getting any TS errors:

#### `LoadingSpinner.vue`
```vue
<template>
  <div class="loading-spinner">
    <slot><span>Loading...</span></slot>
  </div>
</template>

<script setup lang="ts">
  const slots = defineSlots<{
    default: () => [HTMLElement]
  }>();
</script>
```

#### `dashboard.vue`
```html
<LoadingSpinner v-if="loading">
  <span></span>Loading dashboard...
</LoadingSpinner>
```

It's not the perfect example, but I think you understand what I mean. Strictly speaking, the slot children are wrong here, while the types state they're fine.

I also noticed that even though you state the slot children should be `[HTMLSpanElement]`, it accepts any other html tag in there anyway. So you can't really narrow down what type of html elements can be used as slot children.
 
-------

Probably also good if this new feature is documented a little more extensively on the Vue website. The new `defineSlots` syntax isn't documented yet either on vuejs.org, right now. Do you know if someone is working on that? I didn't read any announcements for this new syntax (which also seems to be available without `strictSlotChildren: true`) anywhere either.

This syntax is documented:
```ts
<script setup lang="ts">
const slots = defineSlots<{
  default(props: { msg: string }): any
}>()
</script>
``` 

This one isn't:
```ts
<script setup lang="ts">
const slots = defineSlots<{
  default: (props: { msg: string }) => [HTMLElement]
}>()
</script>
```

Sorry if I should have created issues instead of replying on your pull request!

KazariEX:
After discussing with other members, we think that this feature is not yet perfect. We hope to eventually achieve an effect similar to [`renders`](https://flow.org/en/docs/react/render-types) in Flow, which requires cooperation with core. You can express your opinions on vuejs/rfcs#733 to promote the progress of this proposal.

Anoesj:
Oh I see! So it's experimental? Will the syntax change in the future? Because in that case, probably good to mention that in the Wiki too.

KazariEX:
You are right. I'll consider making this feature experimental for now.

Anoesj:
Thanks, good idea. By the way, as an outsider, it was not very clear what every `vueCompilerOptions` option does. Maybe some before/after code examples per option would help clarify what exactly every option does. Like, I tried `inferComponentDollarRefs: true`, but I really don't know what changed after turning it on.

It also seems like the deepwiki is outdated about this topic: https://deepwiki.com/vuejs/language-tools/2.2-typescript-integration#vue-compiler-options

Anoesj:
I cannot find out how to suggest changes to the wiki or make a PR for it, so I created a discussion for it here: https://github.com/vuejs/language-tools/discussions/5509

Note that I'm still working on it, but accidentally submitted the discussion already (😅), so I'll add more to it in the following hours.

PR #5169, opened 2025-02-10
feat(component-meta): extend component meta

### LINKED ISSUE #5168: Extend component meta

It would be usefull for library authors to be able to extend component metas. In `@nuxt/ui` for example we're using something similar for our devtools, allowing us to pass default property values and examples to preview components (Here's an [example](https://github.com/nuxt/ui/blob/v3/src/runtime/components/Button.vue#L36)).

We could define a new macro `defineComponentMeta(meta: Record<string, any>)` that can be used within a component. Its content would be parsed and injected into the component's meta by the meta checker.

```vue
<script setup lang="ts">
const props = defineProps<{ foo: string }>()

extendComponentMeta({
    bar: 'baz'
})
</script>

<template>
    <span />
</template>
```

Which would result in the following meta:
```ts
{
    bar: 'baz',
    props: [ /* ... */ ],
    // ...
}
```

I can look into the implementation if needed.

romhml:
Resolves #5168 

Introduces a new macro `defineComponentMeta` to extend component meta with arbitrary values.

```vue
<script setup lang="ts">
defineComponentMeta({
  foo: 'bar'
})
</script>

<template>
  <span />
</template>
```

Results in the following meta:

```ts
{
    bar: 'baz',
    props: [ /* ... */ ],
    // ...
}
```

### DIFF
--- a/packages/component-meta/lib/base.ts
+++ b/packages/component-meta/lib/base.ts
@@ -266,8 +266,9 @@ ${commandLine.vueOptions.target < 3 ? vue2TypeHelpersCode : typeHelpersCode}
 		let _events: ReturnType<typeof getEvents> | undefined;
 		let _slots: ReturnType<typeof getSlots> | undefined;
 		let _exposed: ReturnType<typeof getExposed> | undefined;
+		let _extended_meta: ReturnType<typeof getExtendedMeta> | undefined;
 
-		return {
+		const meta = {
 			get type() {
 				return _type ?? (_type = getType());
 			},
@@ -282,8 +283,16 @@ ${commandLine.vueOptions.target < 3 ? vue2TypeHelpersCode : typeHelpersCode}
 			},
 			get exposed() {
 				return _exposed ?? (_exposed = getExposed());
-			},
-		};
+			}
+		}
+
+		return new Proxy(meta, {
+			get(target, prop) {
+				if (prop in target) return target[prop as keyof typeof target]
+				const extendedMeta = _extended_meta ?? (_extended_meta = getExtendedMeta())
+				return extendedMeta?.[prop as string]
+			}
+		})
 
 		function getType() {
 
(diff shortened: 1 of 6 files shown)


packages/component-meta/lib/base.ts
```diff
@@ -297,6 +306,36 @@ ${commandLine.vueOptions.target < 3 ? vue2TypeHelpersCode : typeHelpersCode}
...
+
+			function traverse(node: ts.Node): void {
+				if (ts.isIdentifier(node) && node.text === identifier) {
+					const argument = node.parent.arguments[0];
+					if (ts.isObjectLiteralExpression(argument)) {
+						argument.properties.forEach(property => {
+							if (!ts.isPropertyAssignment(property)) return
+							const key = property.name.getText();
+							const valueNode = resolveDefaultOptionExpression(property.initializer, ts);
+							meta[key] = printer?.printNode(ts.EmitHint.Expression, valueNode, ast) ?? valueNode.getText(ast)
```
romhml, line 328: @KazariEX I'm wondering if it's a good idea to parse nested objects and arrays by traversing the rest of the AST here. I tried to match the parser for prop default values.

On one side it would be nice for end users to receive plain objects / arrays, but might be a bit heavy on performances.

packages/component-meta/lib/base.ts
```diff
@@ -297,6 +306,36 @@ ${commandLine.vueOptions.target < 3 ? vue2TypeHelpersCode : typeHelpersCode}
 			return 0;
 		}
 
+
+		function getExtendedMeta() {
+			let meta: Record<string, any> = {};
+			const snapshot = language.scripts.get(componentPath)?.snapshot!;
+			const fileText = snapshot.getText(0, snapshot.getLength());
+			const ast = ts.createSourceFile('/temp', fileText, ts.ScriptTarget.Latest, true);
```
KazariEX, line 314: See: https://github.com/vuejs/language-tools/blob/317b0cc71bf1a2e8cbf8e1670d21440d89d73ec4/packages/component-meta/lib/base.ts#L334-L344

We can read ast from `vue.tsCodegen` if it's a `.vue` file, or try sharing parsed ast if it's a `.ts` file.

KazariEX: I'm not sure if we can parse `defineComponentMeta` at `language-core` so that we can read the node directly, since it's not useful for IDE.

packages/component-meta/lib/base.ts
```diff
@@ -297,6 +306,36 @@ ${commandLine.vueOptions.target < 3 ? vue2TypeHelpersCode : typeHelpersCode}
...
+					if (ts.isObjectLiteralExpression(argument)) {
+						argument.properties.forEach(property => {
+							if (!ts.isPropertyAssignment(property)) return
+							const key = property.name.getText();
+							const valueNode = resolveDefaultOptionExpression(property.initializer, ts);
+							meta[key] = printer?.printNode(ts.EmitHint.Expression, valueNode, ast) ?? valueNode.getText(ast)
+						});
+					}
+				}
+				return ts.forEachChild(node, traverse)
```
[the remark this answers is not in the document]
KazariEX: I'm not sure if we can parse defineComponentMeta at language-core so that we can read the node directly, since it's not useful for IDE.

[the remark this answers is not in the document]
romhml: I'll look into!

PR #5171, opened 2025-02-11
fix(language-core): add `undefined` to first param type of optional model emits

KazariEX:
fix vuejs/core#12817


packages/language-core/lib/codegen/script/scriptSetup.ts
```diff
@@ -524,6 +524,9 @@ function* generateModelEmit(
 			const [propName, localName] = getPropAndLocalName(scriptSetup, defineModel);
 			yield `'update:${propName}': [value: `;
 			yield* generateDefinePropType(scriptSetup, propName, localName, defineModel);
+			if (!defineModel.required) {
```
RylanBueckert-Broadsign, line 527: Shouldn't there also be a check for a default value? If there is a default value provided, then it should not be allowed to emit with `undefined`.

KazariEX: You are right.

PR #5174, opened 2025-02-13
chore: use pnpm catalog to manage dependencies

KazariEX:


pnpm-workspace.yaml
```diff
@@ -1,4 +1,63 @@
 packages:
-  - 'extensions/*'
-  - 'packages/*'
-  - 'test-workspace'
+  - extensions/*
+  - packages/*
+  - test-workspace
+catalog:
+  "@lerna-lite/cli": &lerna ^3.11.0
```
ghiscoding, line 6: hi there, I'm the maintainer of Lerna-Lite and I brought full `catalog:` support in v3.12.0 only (both lerna commands) and so you should consider updating Lerna-Lite to latest. The summary is that v3.11.0 only had `lerna version` catalog support and `lerna publish` with catalog support was only added in v3.12.0

KazariEX: It seems that lerna only updated the version of a private package and did not update the catalog.

image

And the updated package is not from the root directory.

image

ghiscoding: > I'm not sure about `&lerna` and `*lerna` though since this is new syntax to me, consider testing lerna with `--dry-run` first

like I said, I'm not sure this would work, I've never seen this syntax. However I did set a few catalog entries within Lerna-Lite itself (see [pnpm-workspace.yaml](https://github.com/lerna-lite/lerna-lite/blob/6a115f6629358da060c0968fad98893f4fbb74f0/pnpm-workspace.yaml#L9-L13)) and I also tried it on another project and it works fine with basic catalog definitions. 

There is however an important note to be aware, Lerna will **not** bump any version in the catalog, if you want version bump then you should use `workspace:`. In other words, these entries shown below will probably fail and you should move them out of your workspace catalog and convert them to use `workspace:`... that important note is mentioned in the [docs](https://github.com/lerna-lite/lerna-lite/blob/main/packages/publish/README.md#catalog-protocol)

```sh
"@volar/language-core": *volar
"@volar/language-server": *volar
"@volar/language-service": *volar
```

KazariEX: Got it, thanks.

KazariEX: I changed all references to local dependencies to `workspace:`, but running `lerna version` still doesn’t update these package versions. What could be the reason?

KazariEX: It is anchor and alias syntax in the yaml, I think it should not affact the version bump.

This is the output with `--dry-run`:

image

After I changed `packages/*` to `packages/language-core`, the corresponding package appeared in the output:

image

ghiscoding: > After I changed `packages/*` to `packages/language-core`, the corresponding package appeared in the output:

from where? the pnpm workspace or from lerna config? I don't really know how to help with this. In Lerna-Lite, I changed from `globby` to `tinyglobby` and there was recent tinyglobby versions recently, not sure if that could have an impact but the unit tests are all passing like before so... does it work correctly in your main branch (outside of this PR)?

KazariEX: Ah you are right, it happens on the `tinyglobby^0.2.11`.

KazariEX: Now we should use `packages/**` instead of `packages/*`.

ghiscoding: oh that seems like a breaking change, I might open a tinyglobby issue then, at least we know what it is now :)

johnsoncodehk:
After more thought, I decided that switching to catalog is not without cost. While this reduces duplication of configuration, it also introduces complexity, and `package.json` is no longer compatible with npm (we sometimes need to publish a single package using `npm publish` when the lerna publishing process is interrupted due to network reasons).

At present, catalog is still not a good enough solution. We can see if there are better methods or tools in the future.

ghiscoding:
> (we sometimes need to publish a single package using `npm publish` when the lerna publishing process is interrupted due to network reasons).

@johnsoncodehk @KazariEX 
Perhaps you didn't know but Lerna-Lite can now recover from such failures, it was added 2 years ago in this Lerna-Lite PR [publish: recover from network failure](https://github.com/lerna-lite/lerna-lite/pull/469). You simply have to recall the `lerna publish` command and let it resume from where it previously stopped, which allows you to recover from a possible failure. So you don't have to go with npm to publish, but it's a bit more tricky do deal with that in a CI workflow.

So with that in consideration, you have 2 ways of dealing with this and I recently (yesterday actually) encountered these problems and I'm also using Lerna-Lite with `catalog:` in a large project of mine. I had such issue because I use OTP (one-time-password) and it expired before every packages had time to be published (only 80% ended up being published), so I tried these 2 ways to fix my issue:
1. create a separate workflow to retry lerna publish only 
   - which is what I've done here [publish-npm-publish-retry.yml](https://github.com/ghiscoding/slickgrid-universal/blob/master/.github/workflows/publish-npm-publish-retry.yml) 
2. or modify your CI workflow with bash to automatically detect publish error and ask for a new OTP then retry lerna publish
  - which is what I also changed in my [publish worflow](https://github.com/ghiscoding/slickgrid-universal/blob/d1f553d9bbaa61ad850a9e5467fc03656fb404dd/.github/workflows/publish-npm-latest.yml#L119-L167) 

I tried the number 1 first and it totally worked, I was able to resume my publish. As for the number 2, I didn't fully try it yet since it didn't fail since then but I'm pretty sure it will work... but in the end, in my case the publish problem I had with OTP expiration was caused by Husky running ESLint when lerna publish starts which delayed my publish and caused my OTP to expire and my publish to fail. 

So you can totally use this pnpm `catalog:` if you have proper CI workflow code to deal with possible lerna publish errors. There's no reasons to stick with the old approach if catalog can improve your maintenance (it helps in my project to remove duplicates and have Renovate update only 1 location).

Cheers

PR #5181, opened 2025-02-15
refactor: update alien-signals to 1.0.3

KazariEX:


johnsoncodehk:
Relying on `pauseTracking` and `resumeTracking` is not a good pattern. We can explore refactoring to use finer-grained computeds instead in the future.

PR #5193, opened 2025-02-17
chore(language-core): adjust regex match for @vue-generic comments to improve offset calculation

Gehbt:
before
before

after
after

### DIFF
--- a/packages/language-core/lib/codegen/template/templateChild.ts
+++ b/packages/language-core/lib/codegen/template/templateChild.ts
@@ -46,7 +46,7 @@ export function* generateTemplateChild(
 			yield* ctx.expectError(prevNode);
 		}
 		else {
-			const match = prevNode.loc.source.match(/^(<!--\s*@vue-generic\b\s*\{)(?<content>[^}]*)\}/);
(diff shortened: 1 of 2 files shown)


packages/language-core/lib/codegen/template/templateChild.ts
```diff
@@ -46,12 +46,12 @@ export function* generateTemplateChild(
 			yield* ctx.expectError(prevNode);
 		}
 		else {
-			const match = prevNode.loc.source.match(/^<!--\s*@vue-generic\b\s*\{(?<content>[^}]*)\}/);
+			const match = prevNode.loc.source.match(/^(<!--\s*@vue-generic\b\s*\{)(?<content>[^}]*)\}/);
```
KazariEX, line 49: Try `(?<=<!--\s*@vue-generic\b\s*\{)(?<content>[\s\S]*)(?=\}\s*-->)`?

Gehbt: Yes, and should we also leave some place for comments?

Like `(?<=^<!--\s*@vue-generic\b\s*\{)(?<content>[\s\S]*)(?=\}[\s\S]*-->$)`

KazariEX (approved):

PR #5198, opened 2025-02-19
fix(language-core): do not emit undefined for model with falsy default value

RylanBueckert-Broadsign:
https://github.com/vuejs/core/issues/12817#issuecomment-2669056595


packages/language-core/lib/codegen/script/scriptSetup.ts
```diff
@@ -536,7 +536,7 @@ function* generateModelEmit(
 			const [propName, localName] = getPropAndLocalName(scriptSetup, defineModel);
 			yield `'update:${propName}': [value: `;
 			yield* generateDefinePropType(scriptSetup, propName, localName, defineModel);
-			if (!defineModel.required && !defineModel.defaultValue) {
+			if (!defineModel.required && !('defaultValue' in defineModel)) {
```
KazariEX, line 539:
```suggestion
			if (!defineModel.required && defineModel.defaultValue === undefined) {
```

PR #5248, opened 2025-03-04
refactor: remove hybrid mode configuration

johnsoncodehk:


packages/language-server/package.json
```diff
@@ -16,6 +16,7 @@
 		"directory": "packages/language-server"
 	},
 	"dependencies": {
+		"@typescript/server-harness": "latest",
```
machty, line 19: maybe this should be devDependency?

johnsoncodehk: Fixed in master, thanks!

PR #5256, opened 2025-03-06
fix(vscode): syntax highlighting error if type declaration at the end of script block

### LINKED ISSUE #2060: Syntax highlighting requires new-line or semi-colon between script blocks.
**Issue: script tag syntax highlighting is broken**
Screen Shot 2022-10-24 at 7 51 52 PM

**Workaround: use semi-colon or add new line.**
Screen Shot 2022-10-24 at 7 52 25 PM

### Reproduction

```vue
<script lang="ts">
export type ButtonType = 'primary' | 'secondary' | 'tertiary'
</script>

<script setup lang="ts">
defineProps<{
  type: ButtonType
}>()
</script>

<template>
  <button>
    <slot />
  </button>
</template>
```

zhiyuanzmj:
closed #2060

Before:
image

After:
image


johnsoncodehk:
Thanks for the PR, but this is an upstream issue: https://github.com/microsoft/TypeScript-TmLanguage/issues/873

We cannot add this patch fix on our side because we cannot maintain of it's complexity. Most importantly, since this introduces inconsistencies in TS syntax highlights, we cannot verify whether it will cause other regressions.

PR #5262, opened 2025-03-10
fix(language-core): correct error mapping when prop exp is arrow function

KazariEX:
fix vuejs/core#13012

### DIFF
--- a/test-workspace/tsc/passedFixtures/vue3/#5262/main.vue
+++ b/test-workspace/tsc/passedFixtures/vue3/#5262/main.vue
@@ -1,0 +1,8 @@
+<script setup lang="ts">
+import Comp from './comp.vue';
+</script>
+
+<template>
+	<!-- @vue-expect-error -->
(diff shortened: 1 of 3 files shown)


test-workspace/tsc/passedFixtures/vue3/#5262/main.vue
```diff
@@ -0,0 +1,8 @@
+<script setup lang="ts">
+import Comp from './comp.vue';
+</script>
+
+<template>
+	<!-- @vue-expect-error -->
+	<Comp :foo="() => []" />
```
jacekkarczmarczyk, line 7: the original issue was about not reporting errors for `:foo="() => 'string'"`, passing function that returns non-string value was correctly throwing an error, that's why I think this test should use a function that returns a string

KazariEX: You are right.

PR #5309, opened 2025-04-09
fix(ci): Run auto-fix from the correct branch

AaronDewes:
By default, `pull_request_target` checks out the main branch of vuejs/language-tools and runs the auto fix there. This is very likely not what you wanted to achive.

Also, the permissions pull_request_target has are very broad and can be dangerous.

### DIFF
--- a/.github/workflows/auto-fix.yml
+++ b/.github/workflows/auto-fix.yml
@@ -4,7 +4,7 @@ on:
   push:
     branches:
       - master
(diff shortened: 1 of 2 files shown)


KazariEX:
I am also confused about its behavior🧐

KazariEX:
It seems that `pull_dequest` only applies to PR from the branch of the repository itself?

PR #5318, opened 2025-04-14
chore(deps): replace `path-browserify` with `pathe`

OrbisK:
Stumbling across https://github.com/vuejs/language-tools/pull/5304 I saw that `path-browserify` could be replaced by unjs `pathe`. `path-browserify` has not been updated for 5 years, is only commonjs and from its docs `path-browserify currently matches the Node.js 10.3 API.

I used pnpm catalog to synchronise versions across the monorepo. We can replace this with the current version if needed.
We can even think about replacing `node:path` with `pathe` too.

All vitest tests pass.

If this PR is accepted, I can open a backport to v2.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ dist
 node_modules
 *.tsbuildinfo
 *.vsix
+.idea
 .vscode-test-web
 extensions/*/meta.json
 extensions/*/stats.html
--- a/packages/component-meta/lib/base.ts
+++ b/packages/component-meta/lib/base.ts
@@ -1,6 +1,6 @@
 import { TypeScriptProjectHost, createLanguageServiceHost, resolveFileLanguageId } from '@volar/typescript';
 import * as vue from '@vue/language-core';
-import { posix as path } from 'path-browserify';
+import { posix as path } from 'pathe';
 import type * as ts from 'typescript';
 import { code as typeHelpersCode } from 'vue-component-type-helpers';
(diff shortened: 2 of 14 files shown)


KazariEX:
If I remember correctly, directly replacing it with `pathe` actually results in importing the wrong function type.

OrbisK:
> If I remember correctly, directly replacing it with `pathe` actually results in importing the wrong function type.

Results or results? 😅 

`pathe` should be comparable to `node:path`. There may be a mismatch between `path-browserify` and `pathe`. Can you still reproduce this bug or show me where i can find it?

KazariEX:
image

OrbisK:
> image

you are right. the `* as path` is the problem. Let me replace this 😅

KazariEX:
I plan to switch to the pnpm catalog in this PR #5174, so let's stick with the original dependency management for now.

KazariEX (approved):

johnsoncodehk:
Thanks for your PR! But I'm going to revert this first, let me explain.

We used `upath` in the past, and switched to `path-browserify` for reasons of clarifying module responsibilities: Environment-agnostic modules should not need to consider path consistency.

Since `path-browserify` only supports POSIX, this made it clear that modules that depend on `path-browserify` should only accept POSIX format paths. When `pathe` was introduced, we reintroduced the ambiguity of whether a module might handle win32 format paths.

PR #5326, opened 2025-04-18
fix(component-meta): attach namespace prefix correctly on generated types

### LINKED ISSUE #5323: [vue-component-meta] Type should include namespaces

When using vue-component-meta to analyze component props that utilize namespaced types the library drops the namespace and returns only the base type name.

When I define component props like this:

```typescript
import { Namespace } from './types';

const props = defineProps<{
  testProp: Namespace.Data;
}>();
```
I would expect the `getComponentMeta('').props ` to return a type `Namespace.Data` for this particular property, however the returned type is only `Data` (all the namespaces are dropped).

I found out that this behaviour can be fixed by changing the type generation function from `typeChecker.typeToString(subtype)` to `typeChecker.typeToString(subtype, undefined, ts.TypeFormatFlags.UseFullyQualifiedType)`.

I am not sure if dropping the namespaces is a feature of the library, but I definitely dont find this behaviour helpful. I would be glad if I can turn this setting on in MetaCheckerOptions.

KazariEX:
fix #5323


OndrejHj04 (approved):
Lgtm

Issue #5026, opened 2024-12-05
Syntax highlighting is absent in GitHub for .vue files with comment at top


matthew-white:
### Vue - Official extension or vue-tsc version

N/A

### VSCode version

N/A

### Vue version

N/A

### TypeScript version

N/A

### Steps to reproduce

In a GitHub repository I work on, I noticed that as of the last week, .vue files no longer have syntax highlighting when viewed in GitHub. I think that syntax highlighting in GitHub is determined by [github-linguist/linguist](https://github.com/github-linguist/linguist). They had a release last week (v9.0.0) that updated the Vue grammar to use the one in this repository instead of the one from [vuejs/vue-syntax-highlight](https://github.com/vuejs/vue-syntax-highlight) (PR: github-linguist/linguist#7086).

I think the issue has to do with comments in .vue files. In my repository, there is a file header with copyright information at the top of every file. The file header is in a comment above `<template>`. When I remove that comment, I see that there is syntax highlighting for the file in GitHub.

If linguist needs to use a different grammar file or do something different from what they're doing, I'd be happy to file an issue with them.

### What is expected?

I should see syntax highlighting when I view a .vue file in GitHub.

### What is actually happening?

I do not see syntax highlighting when a .vue file starts with a comment.

### Link to minimal reproduction

https://gist.github.com/matthew-white/1ed66c54adfef4b5153cc9ed793f7529

FeBe95:
I am having the same issue with some files in our repo.

Same goes for bigger files: We have a component which contains almost 4000 lines of code (please don’t judge). Syntax highlighting is also completely absent in this file.

Orrison:
We have the same issue you can see in our public repo here: https://github.com/canyongbs/aidingapp/pull/365/files#diff-29ab79cf819c28a55805ffa38ceab6b755427c05501895fc3fc9e7499fe944d6R78-R81

JoseGoncalves:
Same issue for me at: https://github.com/JoseGoncalves/vue-google-maps/blob/main/src/App.vue

b-d-k:
Confirming same issue with all our Vue files in private (enterprise) repo.

zyoshoka:
I found the reason why this problem does not reproduce in VS Code but it does on GitHub. The problem is the following reference (`text.html.basic#comment`):

[...]

PR #5327, opened 2025-04-19
fix: inline html comment pattern in Vue syntax definition

zyoshoka:
Resolve the Linguist reference issue by inlining html comment pattern. The snapshot is also updated with similar referencing issues.

#5026 will be resolved once this is merged and the next Linguist release (in a few months) is deployed.

### DIFF
--- a/extensions/vscode/syntaxes/vue.tmLanguage.json
+++ b/extensions/vscode/syntaxes/vue.tmLanguage.json
@@ -7,9 +7,6 @@
 			"include": "#vue-comments"
 		},
 		{
-			"include": "text.html.basic#comment"
-		},
-		{
 			"include": "#self-closing-tag"
 		},
 		{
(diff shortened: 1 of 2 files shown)


extensions/vscode/syntaxes/vue.tmLanguage.json
```diff
@@ -1284,6 +1281,16 @@
...
+				{
+					"begin": "<!--",
+					"captures": {
+						"0": {
+							"name": "punctuation.definition.comment.vue"
+						}
+					},
+					"end": "-->",
+					"name": "comment.block.vue"
+				},
```
KazariEX, line 1293: Will this overwrite the `#vue-comments-key-value` behind it?

zyoshoka: Oops, you're right. Fixed it.

KazariEX (approved):

PR #5247, opened 2025-03-03
fix(language-core): prevent eager inference of slot props from generics

KazariEX:
fix #5228, fix #5246

PR #5332, opened 2025-04-22
v2.2.10

KazariEX:
Pick the 4 fixes after 2.2.8 (including the important #5247).


RayGuo-ergou:
oh okay makes sense thanks

RayGuo-ergou:
but isn't tags on commits not branches? So as long as you give the correct hash of the commit to create the hash should be fine?

KazariEX:
Added.

PR #5333, opened 2025-04-23
docs: update changelogs for v2.2.10

Eazash:
While looking into the release of `v2.2.10` via #5332, I noticed that the changelog file that's packaged with the extension has a [link to the file from the main branch](https://github.com/vuejs/language-tools/blob/424e93769af297f636e865bbaa5d730d42c88e1c/extensions/vscode/CHANGELOG.md?plain=1#L1). This can cause problems as the release was off of the v2 branch, as its effectively pointing to an incorrect log. Haven't tested on the v3 extension but I think it's safe to assume it has the same issue.

This PR symlinks `./extensions/vscode/CHANGELOG.md` to `./CHANGELOG.md`. This means the installed extension points to the correct changelog regardless of which branch was used for release. It also has the added benefit of being able to view the changelog from within vscode.

This can cause issues for extension contributors working with windows however, as you'd need to explicitly enable symlink support with `git config core.symlinks=true` ([reference](https://gitforwindows.org/symbolic-links.html)). I believe the benefits outweigh this, but still mentioning this as it can affect future/current contributors
image


KazariEX:
Thank you for bringing this up, we have decided to update change logs directly on the master branch.

Issue #445, opened 2021-09-01
Project finding should ignore irrelevant tsconfigs


johnsoncodehk:
Repro: https://github.com/johnsoncodehk/volar-starter/tree/monorepo-tsconfig-finding

PR #5346, opened 2025-05-01
fix: depend on exact volar version [v2]

tomblachut:
Patch versions of Volar are known to create incompatibilities with Vue Language Tools. This change fixes this problem for the @vue/language-server package downloaded from the npm registry.

For example something here https://github.com/volarjs/volar.js/compare/v2.4.11...v2.4.12 breaks Vue Language Server 2.2.8, and npm will download the latest patch release of dependency. This doesn't show up in VSCode Extension, because Volar 2.4.11 was specified in pnpm-lock of vue-language-tools monorepo, but as far as I know it might affect all other editors.

### DIFF
--- a/extensions/vscode/package.json
+++ b/extensions/vscode/package.json
@@ -565,7 +565,7 @@
 	"devDependencies": {
 		"@types/semver": "^7.5.3",
 		"@types/vscode": "^1.82.0",
-		"@volar/vscode": "~2.4.11",
+		"@volar/vscode": "2.4.11",
 		"@vscode/vsce": "^3.2.1",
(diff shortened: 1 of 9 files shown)


johnsoncodehk:
Thanks 👍, we should have done this earlier.

electfreak:
NB this was reverted
[#445](https://github.com/vuejs/language-tools/commit/4456033d5718660964de9a149af329309537940f)

johnsoncodehk:
We will bundle `@vue/language-server` instead.

johnsoncodehk:
We just released 2.2.12 for this.

PR #5355, opened 2025-05-08
feat: expose api via `vue-tsc`

sxzz:
Expose all APIs from `vue-tsc`, so that `rolldown-plugin-dts` can just add one peer dependency, without increasing bundle size of `rolldown-plugin-dts` itself.

Also should be ported to v2 branch.

### DIFF
--- a/packages/tsc/api.ts
+++ b/packages/tsc/api.ts
@@ -1,0 +1,2 @@
+export * from '@volar/typescript';
+export * from '@vue/language-core';


johnsoncodehk:
`vue-tsc` does not expose dependencies by design, instead you can import `vue-tsc`'s dependencies you needed from `node_modules/vue-tsc/node_modules/...`. Please reference to: https://github.com/johnsoncodehk/tsslint/blob/85adb15843ec5c20916568fd8409b2adb9016a52/packages/cli/lib/languagePlugins.ts#L19-L21

PR #1255, opened 2022-05-03
chore: publish to open-vsx

lukashass:
This adds publishing to open-vsx in a GitHub action.

### Things you would still need to do:

1. Register at https://open-vsx.org/
2. Set `OVSX_PAT` GitHub action secret
3. [Claim](https://github.com/eclipse/openvsx/wiki/Namespace-Access#how-to-claim-a-namespace) the `Vue` namespace

### Additional notes

I published https://open-vsx.org/extension/lukashass/volar for testing.

Publishing with target-platform support is blocked by:
- https://github.com/eclipse/openvsx/issues/450
- https://github.com/EclipseFdn/open-vsx.org/issues/1084.

PR #5377, opened 2025-05-24
chore: mark prereleases when publishing to Open VSX

lukashass:
Fixes the pre-release status of future versions. As discussed here: https://github.com/vuejs/language-tools/pull/1255#issuecomment-2906731482


johnsoncodehk:
Thanks! We need to wait for https://github.com/EclipseFdn/open-vsx.org/issues/4022 to be resolved before merging this.

PR #5385, opened 2025-05-28
feat(language-service): document links for template refs

alex-snezhko:
Adds LSP document links for template refs in setup script referencing the corresponding element(s) in the template.

This is a small QoL feature that is useful in my opinion as it makes it clear that the template reference is valid and traversable, in the same manner as the existing CSS class definition document link feature.

### Screenshots in VS Code

Overview:
Screenshot From 2025-05-27 20-37-34

Traversing link targets correct range in template:
Screenshot From 2025-05-27 20-38-06

Hovering the multi-element ref shows multiple links:
Screenshot From 2025-05-27 20-37-45


johnsoncodehk:
Awesome work, thanks!

PR #5405, opened 2025-06-07
docs(nvim): update readed me for neovim with vue_lsp v3

RayGuo-ergou:
Would appreciate anyone testing this with most recent commit and let me know if it works.


RayGuo-ergou:
Will write in wiki instead, also a pr upcoming for remove the outdated config.

PR #5406, opened 2025-06-07
Change "volar" to "vue_ls" in all examples

dragunovartem99:
In latest updates of lspconfig, and tooling around it (like mason), we are installing and configuring `vue_ls`. `volar` is unsupported key in such configurations

Working example with `vue_ls` from my personal config: https://github.com/dragunovartem99/nvim/blob/main/lua/config/lsps/list.lua

P.S. Also I've made small grammar changes for consistency

Edit:

`vue_ls` is the only relevant installation candidate in Mason

Related PR from `nvim-lspconfig`: https://github.com/neovim/nvim-lspconfig/pull/3843
Related PR from `mason-lspconfig`: https://github.com/mason-org/mason-lspconfig.nvim/pull/561


RayGuo-ergou:
Hi, thanks for the PR, we are aware that the lsp has renamed to `vue_lsp` in both mason and nvim-lspconfig.

However, in the next major release we will move the configuration docs into github wiki, I just did that what a timing haha. (see https://github.com/vuejs/language-tools/wiki/Neovim)

I think it still makes sense to correct the lsp name, can you change the targeting branch to v2 instead?

Thanks.

dragunovartem99:
@RayGuo-ergou yes, thank you for explanation, I will create a new PR :D

Issue #5487, opened 2025-07-09
selected html tag highlighting is incorrect


brainz-jshu:
### Vue - Official extension or vue-tsc version

3.0.1

### VSCode version

1.101.2

### Vue version

3.5.17

### TypeScript version

5.8.3

### System Info

```shell
System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 AMD Ryzen 5 7530U with Radeon Graphics
    Memory: 6.45 GB / 15.32 GB
  Binaries:
    Node: 22.13.0 - C:\Program Files\nodejs\node.EXE
    npm: 10.9.2 - C:\Program Files\nodejs\npm.CMD
  Browsers:
    Edge: Chromium (127.0.2651.74)
    Internet Explorer: 11.0.19041.5794
```

### package.json dependencies

```json
"dependencies": {
    "@tailwindcss/vite": "^4.1.10",
    "ant-design-vue": "^4.2.6",
    "axios": "^1.10.0",
    "d3": "^7.9.0",
    "d3-hexbin": "^0.2.2",
    "dayjs": "^1.11.13",
    "fluid-dnd": "^2.4.0",
    "highcharts-vue": "^2.0.1",
    "lodash": "^4.17.21",
    "nanoid": "^5.1.5",
    "pinia": "^2.3.1",
    "socket.io-client": "^4.8.1",
    "tailwindcss": "^4.1.10",
    "vue": "^3.5.17",
    "vue-router": "^4.5.1"
  },
  "devDependencies": {
    "@tsconfig/node22": "^22.0.2",
    "@types/d3": "^7.4.3",
    "@types/d3-hexbin": "^0.2.5",
    "@types/jsdom": "^21.1.7",
    "@types/lodash": "^4.17.18",
    "@types/node": "^22.15.32",
    "@vitejs/plugin-vue": "^5.2.4",
    "@vitest/eslint-plugin": "1.2.7",
    "@vue/eslint-config-prettier": "^10.2.0",
    "@vue/eslint-config-typescript": "^14.5.1",
    "@vue/test-utils": "^2.4.6",
    "@vue/tsconfig": "^0.7.0",
    "cypress": "^13.17.0",
    "eslint": "^9.29.0",
    "eslint-plugin-cypress": "^4.3.0",
    "eslint-plugin-vue": "^9.33.0",
    "jsdom": "^25.0.1",
    "npm-run-all2": "^7.0.2",
    "prettier": "^3.6.0",
[...]

PR #5490, opened 2025-07-11
fix(language-service): selected html tag highlighting is incorrect #5487

1want:


KazariEX:
The fix is incorrect. This language service plugin is designed to correctly map document highlights from virtual files back to source code. #5487 is because that we use the same prop under the `IntrinsicElementAttributes` interface for native tags with the same name.

PR #5514, opened 2025-07-17
revert: type support of slot children (#5137)

KazariEX:
After discussion, we believe the implementation of this feature is not yet mature and cannot adequately address real-world scenarios. Therefore, we revert it for now and will continue to wait for progress on [this rfc](https://github.com/vuejs/rfcs/discussions/733).


DrJume:
Oh, very sad to see this. We already used it internally and were very happy that this was possible.

Would it be possible to enable this feature under a flag of `vueCompilerOptions`?

KazariEX:
The current implementation may not be the same as the final solution, so we hope to avoid too much breaking changes caused by late modifications. Sorry for setting the false expectations.

LouisHaftmann:
Why not enable it via an experimental flag? Users should expect bugs or breaking changes from an experimental feature. Something like:

```json
{
  "vueCompilerOptions": {
    "experimental": {
      "strictSlotChildren" true
    }
  }
}
```

I would rather have breaking changes with an experimental feature than having to wait for the feature to be fully released.

Anoesj:
To be honest, I expected it to become an experimental feature too. Maybe for sponsors only?

PR #5528, opened 2025-07-18
feat(tsc): add vue-tsc/typescript entry for programmatic usage

escaton:
Moving on with https://github.com/vuejs/language-tools/pull/5517

Add `vue-tsc/typescript` entry for programmatic usage.
This change will make it possible to seamlessly use vue-tsc with
- https://github.com/typestrong/fork-ts-checker-webpack-plugin/
- https://github.com/rspack-contrib/ts-checker-rspack-plugin
- (get rid of copy-paste glue code) https://github.com/fi3ework/vite-plugin-checker/tree/main/packages/vite-plugin-checker/src/checkers/vueTsc
- any other tool that accepts path to `typescript/lib/typescript`

Not sure about file naming, but this is the least changes needed to make it work without breaking anything.

### DIFF
--- a/packages/tsc/typescript.ts
+++ b/packages/tsc/typescript.ts
@@ -1,3 +1,3 @@
-import { run } from '.'
+import { run } from '.';
 
-export default run(require.resolve('typescript/lib/typescript'))
+export = run(require.resolve('typescript/lib/typescript'))
--- a/packages/tsc/typescript.ts
+++ b/packages/tsc/typescript.ts
@@ -1,0 +1,3 @@
+import { run } from '.'
+
+export deafult run(require.resolve('typescript/lib/typescript'))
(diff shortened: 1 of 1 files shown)


escaton:
Not sure why the auto-fix job fails to commit something on a detached HEAD

johnsoncodehk:
Thanks for the PR, but this is outside the intended use case of vue-tsc.

The package name of vue-tsc implies that it is only used as a CLI tool, adding this entry is likely to mislead people into thinking this is the recommended usage for vue-tsc.

escaton:
Hi @johnsoncodehk thanks for the comment, I see your point, to some degree it makes sense.
But on the other hand for the sake of ergonomics I think it is better for that single line still be a npm package, rather then a utility file in every project.
For now I created [@esctn/vue-tsc-api](https://www.npmjs.com/package/@esctn/vue-tsc-api)
Let me know if you are interested to move it in this repo.

PR #5541, opened 2025-07-20
feat: add vue vapor completion support for script setup blocks

runyasak:
# Overview
I've added vapor completion support for Vue 3.6 in `<script setup blocks>`.

# Screenshots
https://github.com/user-attachments/assets/fd1b6d00-beb3-4feb-aa6a-64a383081e34

# Discussions
- Vue Vapor is currently in alpha (Vue 3.6.0-alpha.1), so I'm not entirely sure if we should add this feature right now or wait for a more stable release. What are your thoughts on this?
- Should we add a setting to enable/disable this feature? Since Vue Vapor is an opt-in feature and still experimental, I'm wondering if we should have a configuration option to disable vapor completions for users who don't want to see them yet.

Something like:
```json
{
  "vue.complete.vapor": true // default: true or false?
}
```

I'd appreciate your thoughts on:
- Whether it's too early to add this feature given Vapor's alpha status
- If we should include a configuration option to toggle vapor completions
- What the default value should be if we do add the config

---

If you have any feedback or suggestions, please do not hesitate to let me know.


KazariEX:
We can just write `vapor` instead of `setup vapor`, which is equivalent to writing `setup vapor` at compile time.

Since the data is automatically crawled from official document, please do not modify it manually.

runyasak:
@KazariEX  I'm sorry for updating the JSON data manually—I’ve already reverted those changes.

>We can just write vapor instead of setup vapor, which is equivalent to writing setup vapor at compile time.

Thank you for clarifying that we can just write vapor instead of setup vapor, and that it’s equivalent at compile time. As I understand, the syntax should be `<script vapor>` instead.
Should I implement completion for this feature? If so, I’d appreciate any guidance on how you’d like it handled.

If you or someone else is already working on this, please let me know and I can close this PR if needed.

KazariEX:
According to the current data generation logic, we need to wait for the official document to add the `script vapor` entry before implementing this feature.

See https://github.com/vuejs/language-tools/blob/master/packages/language-service/scripts/update-html-data.js.

runyasak:
@KazariEX I understand you now. So, I will close this PR for now and wait until the feature is ready in the official documentation. If you want me to contribute or implement this feature in the future, please let me know or feel free to re-open this PR.

Thank you for discussing this with me. 😁

PR #5562, opened 2025-07-26
feat: support tsconfig path alias resolution for document links

### LINKED ISSUE #3460: template img src path aliais jump problem in vscode
in vscode，When using path aliases for the template and img src of the Vue file, cmd+click cannot jump to the corresponding path, but the import in the script can jump using path aliases.

KazariEX:
resolve #3460


AlexVagrant:
@johnsoncodehk  Hello, do you have an estimated time for merging this pull request? I think it's excellent work.

PR #5589, opened 2025-08-05
chore: update `vue-component-type-helpers` to current version

kingyue737:
Fix unmatched peer deps warning
```
 WARN  Issues with peer dependencies found
├─┬ @nuxt/content 3.6.3
│ └─┬ nuxt-component-meta 0.12.2
│   └─┬ vue-component-meta 3.0.5
│     └── ✕ unmet peer vue-component-type-helpers@3.0.1: found 3.0.5
```

### DIFF
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -429,31 +429,26 @@ packages:
     resolution: {integrity: sha512-EL0+uMSdj/n+cZOP9ZO8ndvjmtOSWXNsMHKdAAaTG0+EjH9M9YKXD6kopP6PKOR5pJuiyHCRpVKJ4xoD4adfpQ==}
     cpu: [arm64]
     os: [linux]
-    libc: [glibc]
 
   '@dprint/linux-arm64-musl@0.50.0':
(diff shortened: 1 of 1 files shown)


johnsoncodehk:
@ghiscoding When a workspace package is added to `peerDependencies`, it seems to cause version numbers to become out of sync when running `lerna version`. Is this expected behavior?

https://github.com/vuejs/language-tools/blob/1b98031bdfdf50265f073cec54d45cd001bc6827/packages/component-meta/package.json#L20-L23

ghiscoding:
Lerna/Lerna-Lite doesn't touch peer dependencies by default because typically a package manager should never bump peer deps by itself since that is more of a user configuration and shouldn't be automated (it was explained by the original Lerna's author a while ago). But if you do want the tool to bump them, I added a new flag some time ago [`--allow-peer-dependencies-update`](https://github.com/lerna-lite/lerna-lite/tree/main/packages/version#--allow-peer-dependencies-update) which is in Lerna-Lite only

From original Lerna's author:

> Changes to a peer version range are always semver major, and should be as broad as possible. Until we can get fancier, we should never automatically modify them to match the new version being published (which is the current incorrect behavior).

johnsoncodehk:
Thank you for the clarification!

PR #5666, opened 2025-09-25
feat(language-server): add back `typescript.tsdk` initialization option

johnsoncodehk:


electfreak:
Thank you!

rchl:
I wonder what is the relevance of this setting now that the most of the typescript functionality is delegated to a separate typescript process?

Should editor packages like LSP-vue still strive to set tsdk to workspace version of typescript or rather use the version of typescript bundled with the vue language server?

johnsoncodehk:
See #5691.

PR #5678, opened 2025-09-30
feat(language-core): support configuration for language plugins

KazariEX:
This PR allows us to provide configuration for the Vue language plugins in the same way as the TypeScript language plugins.

```json
{
  "vueCompilerOptions": {
    "plugins": [
      {
        "name": "...",
        "options": { "foo": "bar" }
      }
    ]
  }
}
```


posva:
Is there a way to type the options?

currently doing

```ts
const plugin: VueLanguagePlugin = ({
  compilerOptions,
  modules: { typescript: ts },
  ...options
}) => {
  // rootDir comes from plugin options (vuejs/language-tools#5678)
  const optionsRootDir = (options as { rootDir?: string }).rootDir
  // Prioritize plugin options over tsconfig
  const rootDir = optionsRootDir ?? compilerOptions.rootDir
```

posva:
What's the syntax to make it work? It doesn't seem to recognize the name when I try it this way:

```jsonc
    "plugins": [
      "unplugin-vue-router/volar/sfc-route-blocks",
      "unplugin-vue-router/volar/sfc-typed-router"  // works
      // does not work
      // {
      //   "name": "unplugin-vue-router/volar/sfc-typed-router",
      //   "options": { "rootDir": "." }
      // }
    ]
```

KazariEX:
The value of `ctx.config` is the entire original plugin object. (which is also the behavior of ts plugins!)

posva:
ah so it would be `config.options.rootDir`? But in that case, it should still have worked because I set a fallback value of `rootDir` 🤔 

Testing this at https://github.com/vuejs/router/pull/2603

PR #5690, opened 2025-10-01
fix(typescript-plugin): place `__vue__` in project instead of program

johnsoncodehk:
Since `info.project.program` is rebuilt with each code change, the reactivity analysis plugin loses access to `__vue__`.

Which can be resolved by hoisting `__vue__` to `info.project`.


johnsoncodehk:
cc @zcf0508, this will break the current access method of `ts-type-explorer`.

KazariEX (approved):

PR #5691, opened 2025-10-01
feat(language-server): support `--tsdk` command line arg

johnsoncodehk:
`@vue/language-server`'s default TypeScript import method causes issues for WebStorm IDE when bundling the server.

Specifically, this causes the bundler to include TypeScript modules by default, significantly increasing the bundle size.

The previous PR #5666 couldn't fully resolve this issue. To maintain backward compatibility, `typescript.tsdk` was introduced as an optional parameter rather than the only working method. The code still contains TypeScript import logic, so the bundler continues attempting to bundle it.

To decouple the module dependency on TypeScript, this PR moves the language server logic to `lib/server.ts` and requires passing the TypeScript module as a parameter to `startServer()`.

Additionally, the `vue-language-server` command now recognizes the `--tsdk` parameter to handle TSDK imports outside of the LSP initialization logic.


mickverm:
This made it into the 3.1.1 release, any chance this could be added to 3.0.9, as mentioned in https://github.com/vuejs/language-tools/discussions/5455#discussioncomment-14570777?

KazariEX:
@mickverm v3.0.9 released.

Issue #4403, opened 2024-05-19
Incorrect auto-complete for tag-like string in template literals


immccn123:
I've encountered an issue with the plugin incorrectly handling tag-like strings within template literals. Below is a example that reproduces the problem:

```
<template>
  {{ "<this_is_a_tag_like string>" }}
</template>
```

https://github.com/vuejs/language-tools/assets/41335471/c7e018d6-c23f-4140-9cbe-4a642f5f6423

In this example, the plugin incorrectly interprets the string `<this_is_a_tag_like string>` as an HTML tag. This should be treated as a regular string within the template context.

## More Example

https://github.com/vuejs/language-tools/assets/41335471/e692197c-8322-4316-b261-b02c4104751e

While this issue does not significantly affect functionality, it does lead to a less smooth development experience. I would appreciate it if this issue could be addressed in a future update.

PR #5744, opened 2025-11-11
fix(language-service): prevent auto-insertion of html snippets in template interpolation

serkodev:
Fix: #4403

### Demo
https://github.com/user-attachments/assets/d596ee17-684a-4961-9a6a-c55054901a08


KazariEX:
I believe preprocessing is superior to postprocessing. Feel free to port my changes over, as I'm not good at writing tests!

PR #5745, opened 2025-11-11
fix(language-service): replace interpolations with spaces during `provideAutoInsertSnippet`

### LINKED ISSUE #4403: Incorrect auto-complete for tag-like string in template literals

KazariEX:
fix #4403, close #5744

### DIFF
--- a/packages/language-core/index.ts
+++ b/packages/language-core/index.ts
@@ -6,6 +6,7 @@ export * from './lib/parsers/scriptSetupRanges';
 export * from './lib/plugins';
 export * from './lib/types';
 export * from './lib/utils/collectBindings';
+export * from './lib/utils/forEachTemplateNode';
 export * from './lib/utils/parseSfc';
 export * from './lib/utils/shared';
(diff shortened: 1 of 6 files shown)


KazariEX:
Replacing interpolation with spaces when generating virtual document for the template will cause formatting errors.

PR #5759, opened 2025-11-17
 feat(language-service): support tsconfig path alias resolution for document links #5562 resolve conflicts

AlexVagrant:
resolve conflicts #5562


KazariEX:
The merge conflicts are not what’s preventing this PR from being merged... the real issue is whether there is a better way to implement it.

AlexVagrant:
> The merge conflicts are not what’s preventing this PR from being merged... the real issue is whether there is a better way to implement it.

So that's it. Do you have any ideas?

PR #5765, opened 2025-11-19
fix(language-core): enhance `getVIfNode` to support `v-else-if` directives

### LINKED ISSUE #5492: v-else-if and v-slot co-usage triggers false ts-plugin(6133) "unused" warning/error

3.0.1

### VSCode version

1.102.0

### Vue version

3.5.17

### TypeScript version

5.8.3

### System Info

```shell

```

### package.json dependencies

```json

```

### Steps to reproduce

1. Create a new Vue 3.5+ project using `<script setup lang="ts"`.
2. Install and use the **Vue - Official** extension in VS Code.
3. Add an imported function in `<script setup lang="ts"` that will **only be used inside** `v-else-if` block in the template.
4. Use the function in the template like this:

``` vue 3.5
<template>
  <div>
    <div v-if="func1()"></div>
    <div v-else-if="func2()></div>
    <div v-else></ div>
  </div>
</template>
<script setup lang="ts">
import { func1, func2 } from "somewhere" // `func2` is declared but its value is never read. ts-plugin(6133)
</script>
```

### What is expected?

The imported function `func2` should not trigger ts-plugin(6133) warning/error if it is used anywhere in the template, even inside a `v-else-if`.

The imported function is marked as **unused** (ts-plugin(6133) warning/error) when used **only inside a** `v-else-if` expression, despite being referenced in the template.

_No response_

### Any additional comments?

_No response_

serkodev:
Fixed #5492

### DIFF
--- a/test-workspace/tsc/passedFixtures/vue3/#5492/main.vue
+++ b/test-workspace/tsc/passedFixtures/vue3/#5492/main.vue
@@ -6,14 +6,14 @@ const cond2 = true
 </script>
 
 <template>
-    <Comp>
-        <template #default v-if="cond1">1</template>
-        <template #default v-else-if="cond2">2</template>
+	<Comp>
+		<template #default v-if="cond1">1</template>
+		<template #default v-else-if="cond2">2</template>
 
-        <div v-if="cond1 === true"></div>
-        <div v-else-if="cond1.length"></div>
+		<div v-if="cond1 === true"></div>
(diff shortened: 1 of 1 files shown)


KazariEX:
This will cause `v-else-if` to be generated as a seperate if statement, preventing type narrowing from working correctly.

serkodev:
Oh you're right. Let me think of another approach.

Issue #5810, opened 2025-12-06
Build errors after updating to 3.1.6


kubajmarek:
### Vue - Official extension or vue-tsc version

3.1.6

### VSCode version

Not used

### Vue version

3.5.25

### TypeScript version

5.8.3

### System Info

```shell
System:
    OS: Linux 6.12 EndeavourOS
    CPU: (12) x64 AMD Ryzen 5 5600X 6-Core Processor
    Memory: 11.93 GB / 31.25 GB
    Container: Yes
    Shell: 5.3.3 - /bin/bash
  Binaries:
    Node: 22.17.0 - /home/kubajmarek/.nvm/versions/node/v22.17.0/bin/node
    npm: 10.9.2 - /home/kubajmarek/.nvm/versions/node/v22.17.0/bin/npm
    bun: 1.3.3 - /home/kubajmarek/.bun/bin/bun
  Browsers:
    Chromium: 142.0.7444.175
    Firefox: 145.0.2
    Firefox Developer Edition: 145.0.2
```

### package.json dependencies

```json

```

### Steps to reproduce

Run `vue-tsc --build`

### What is expected?

Clean build without errors, as in v3.1.5

### What is actually happening?

3 errors are produced:

One:
```
TypeError: Cannot read properties of undefined (reading 'length')
    at walkObjectLiteral (/home/kubajmarek/Documents/code/PROJECT/node_modules/@vue/language-core/lib/codegen/template/element.js:307:138)
    at walkObjectLiteral.next (<anonymous>)
    at generateStyleScopedClassReferences (/home/kubajmarek/Documents/code/PROJECTm/node_modules/@vue/language-core/lib/codegen/template/element.js:281:28)
    at generateStyleScopedClassReferences.next (<anonymous>)
    at generateComponent (/home/kubajmarek/Documents/code/PROJECT/node_modules/@vue/language-core/lib/codegen/template/element.js:190:12)
    at generateComponent.next (<anonymous>)
    at generateTemplateChild (/home/kubajmarek/Documents/code/PROJECT/node_modules/@vue/language-core/lib/codegen/template/templateChild.js:48:56)
    at generateTemplateChild.next (<anonymous>)
    at generateElement (/home/kubajmarek/Documents/code/PROJECT/node_modules/@vue/language-core/lib/codegen/template/element.js:237:58)
    at generateElement.next (<anonymous>)
```

Two:
```
TypeError: Cannot read properties of undefined (reading 'length')
[...]

PR #5813, opened 2025-12-06
feat(lint): add typescript services types lint rule

serkodev:
Related #5810


johnsoncodehk:
This lint rule also uncovered another issue: `isDefaultImport` is always false in vue-tsc.

https://github.com/vuejs/language-tools/blob/d245142725c97a13d49fd4435ed56ab1d3e48245/packages/language-core/lib/parsers/utils.ts#L63

PR #5853, opened 2025-12-11
fix(vscode): vue ts highlighting when trailing type alias misses semicolon

### LINKED ISSUE #2060: Syntax highlighting requires new-line or semi-colon between script blocks.

serkodev:
Fix #2060

- Replace the `end` with `while` in patterns.
  - `end` waits for one terminating match
  - `while` re-checks per line and drops the scope as soon as its condition fails.
- Add extra rule to support single line `<script>`

### Before
before

### After
after

### DIFF
--- a/extensions/vscode/syntaxes/vue.tmLanguage.json
+++ b/extensions/vscode/syntaxes/vue.tmLanguage.json
@@ -358,6 +358,16 @@
 							"include": "#tag-stuff"
 						},
 						{
+							"begin": "(?<=>)(?=[^\\n]*<\\/script[\\s>])",
+							"end": "(?=<\\/script[\\s>])",
+							"name": "source.ts",
+							"patterns": [
+								{
+									"include": "source.ts"
+								}
+							]
+						},
+						{
 							"begin": "(?<=>)",
 							"while": "^(?!\\s*<\\/script[\\s>])",
 							"name": "source.ts",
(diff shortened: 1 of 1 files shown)


serkodev:
### Notice
- If `</script>` is intended as a closing tag but does not appear at the beginning of a line, syntax highlighting will also break (though single-line `<script>` tags are not affected):
  ```vue
  <!-- bad -->
  <script setup lang="ts">
  const a = 1;</script>

  <!-- good -->
  <script setup lang="ts">const a = 1;</script>
  <script setup lang="ts">
  const a = 1;
  </script>
  ```
- If `</script>` appears as a `string` and at the beginning of a line, syntax highlighting will break:
  ```vue
  <script setup lang="ts">
  const a = `
  </script> <- break
  `;
  </script>
  ```
  However, since `</script>` as a string currently is also not supported by @vuejs/core, so this issue has no impact.
  Ref: https://github.com/vuejs/core/issues/7391

PR #5862, opened 2025-12-12
feat(component-meta): add `tags` to slots and exposed

aj-dev:
This PR adds `tags` to slots and exposed so that jsDoc tags can be used consistently across props, events, slots and exposed.


johnsoncodehk:
LGTM, thanks for the PR!

PR #5891, opened 2025-12-20
fix(component-meta): skip schema resolution correctly when `option` is `false`

KazariEX:


johnsoncodehk:
The TS plugin has already passed the schema parameter as false; the issue appears to be that `vue-component-meta` does not correctly handle the behavior when false.

https://github.com/vuejs/language-tools/blob/880aaf0968cce612da99be66e5e591583e1315cd/packages/typescript-plugin/lib/requests/getComponentMeta.ts#L27

PR #5967, opened 2026-02-16
fix(language-service): provide data for all default html tags

liangmiQwQ:
For now, only tags `getComponentsAndElements` returned can be returned in `provideTags` fn, and it seems caused issues, like some void tags, will process incorrect document symbols without data provider returning (like this)

Screenshot 2026-02-16 at 15 50 44

To be honest, I feel quite strange when I making up this PR, I first wonder why `getComponentsAndElements` doesn't return `img` and I did some debug, found the only one tag this function return is `[ "Vue" ]` for a no-script SFC, I don't know how `tsserver` work, but I think it's really strange. And at the same time, I also wonder why `provideTags` added all tags in `builtInData` but `defaultHtmlTags` doesn't


KazariEX:
How to reproduce the situation?

liangmiQwQ:
1. Create a `.vue` file with code like this

```vue
<template>
    <img>
    <img>
    <div />
    <div />
</template>
```

2. open it, and run `User View Container: Focus on Outline View` command, then you can see that

Screenshot 2026-02-16 at 16 59 01

However, this seems to happen randomly sometimes, sometimes the outline become normal after I do some edit to the file, sometimes it still acts wrongly, sometimes the situation only shows when I set `"typescript.validate.enable": false`, while it always shows sometimes

KazariEX:
`getComponentsAndElements` should not return `["Vue"]`.

liangmiQwQ:
Idk why, but...

Screenshot 2026-02-16 at 17 28 08
Screenshot 2026-02-16 at 17 28 23

KazariEX:
I cannot reproduce this behavior, but I think that this issue is caused by us not running `provideHtmlData` in `provideDocumentSymbols`.

liangmiQwQ:
let me do some experiments after dinner.... and thank you so much!

liangmiQwQ:
> I cannot reproduce this behavior, but I think that this issue is caused by us not running `provideHtmlData` in `provideDocumentSymbols`.

Trace on #5968.

PR #5968, opened 2026-02-16
fix(language-service): use default html data provider for document symbols

liangmiQwQ:
close #5967

Now data provider run after you do hover / completion action, but not for document symbol requests, may cause make outline / breadcrumbs display differently before and after editing


liangmiQwQ:
In this branch
Screenshot 2026-02-16 at 19 16 49

PR #5983, opened 2026-03-11
fix(vscode): use regex for TS extension patching to support VS Code 1.110+

### LINKED ISSUE #4221: File Explorer "Find File References" not working for .vue SFC
### Description
When **not** in hybrid mode, when right-clicking on a `.vue` file, VSCode responds with the error shown in the picture below.
Works without an issue when in hybrid mode.

image

### Expected behaviour
Actually finding the references for the SFC component.

### Environment

- OS: `Windows 11 Version 10.0.22631 Build 22631`
- VSCode: `v1.87.2` (latest)
- Vue Extension: `v2.0.10`
- Typescript: `v5.2.2`

### Additional information
Possibly related to https://github.com/vuejs/language-tools/issues/4171, but not entirely sure, hence the new issue.

ebiryu:
## Summary

- VS Code 1.110 switched its extension build system from `tsb` to `tsgo` (microsoft/vscode#292461), changing the minified variable names in the built-in TypeScript extension's `dist/extension.js`
- The hardcoded string patterns in `patchTypeScriptExtension()` no longer matched, causing the monkey-patches to silently fail
- This broke "Find File References" and other TypeScript features for `.vue` files, because `isSupportedLanguageMode` no longer included `vue`

This PR replaces the four hardcoded string replacements with regex-based patterns that match identifier names regardless of minification, supporting both the old format (`t.typescript`, `t.jsTsLanguageModes`) and the new format (`bl`, `Oh`).

Fixes https://github.com/vuejs/language-tools/issues/4221#issuecomment-4036019402

### DIFF
--- a/extensions/vscode/src/extension.ts
+++ b/extensions/vscode/src/extension.ts
@@ -328,25 +328,43 @@ function patchTypeScriptExtension() {
 		if (args[0] === extensionJsPath) {
 			let text = readFileSync(...args) as string;
 
+			const id = String.raw`[\w$]+(?:\.[\w$]+)?`;
+
 			// patch jsTsLanguageModes
+			// before 1.110: t.jsTsLanguageModes=[t.javascript,t.javascriptreact,t.typescript,t.typescriptreact]
+			// since 1.110:  "javascriptreact",Oh=[Ma,Ua,bl,Ns]
 			text = text.replace(
-				't.jsTsLanguageModes=[t.javascript,t.javascriptreact,t.typescript,t.typescriptreact]',
-				s => s + '.concat("vue")',
+				new RegExp(
+					String
+						.raw`(\.jsTsLanguageModes=\[${id},${id},${id},${id}\])|("javascriptreact",(${id})=\[(${id},${id},${id},${id})\])`,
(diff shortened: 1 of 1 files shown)


johnsoncodehk:
LGTM, thanks!

vidal7:
Thank you. For users still using Vue 2. Does the `3.0.8` version still working with VS Code 1.110?

dadigu:
> Thank you. For users still using Vue 2. Does the `3.0.8` version still working with VS Code 1.110?

I can confirm. 3.0.8 does not work with VS Code 1.110+. `findAllFileReferences` fails with "Unsupported file type". The TS server log shows the plugin loads fine, but the command is rejected client-side before reaching the server.

We're also maintaining a Vue 2.7 app, so upgrading to 3.1+ isn't an option. Would it be possible to backport this fix to a 3.0.x release?

johnsoncodehk:
You can use `@vue/language-server` v3.0.x together with the Vue VSCode extension v3.2.6 for compatibility, see https://github.com/vuejs/language-tools/wiki/Vue-2-Compat-Guides.

PR #6025, opened 2026-04-25
fix(vscode): do not access config before activate is called

kermanx:
`resolveServerPath()` and `patchTypeScriptExtension()` were being called at module top level, causing `config.server.path` and `config.server.includeLanguages` to be read before the extension is activated.

This moves both calls inside the `defineExtension` callback so they run after `activate`.


johnsoncodehk:
`patchTypeScriptExtension()` needs to stay at the root — it has to patch `fs.readFileSync` before the TS extension loads. If it runs inside `defineExtension`, the TS extension may already be active and we'd hit the `tsExtension.isActive` branch that forces a restart.

`resolveServerPath()` is fine to move though.

PR #6049, opened 2026-05-16
feat(language-core): check required fallthrough attributes

### LINKED ISSUE #5882: fallthroughAttributes improvements: requiredness and hiding

`<!-- @fallthroughAttributes true -->` is a great tool on components meant to wrap and extend other components.

In practice, there are a few details that could be more accurate.
Let's call `O` the outside wrapper and `I` the inner component.

1. All fallthrough attributes are optional props on `O`, so unlike regular DX there's no indication when you use `<O>` and do not set a required props. 
Ideally this should take into account any prop directly set by `O` on `I`, which means it's now optional on `O` itself (or should it not be a prop at all? I'm not sure whether `$attrs` fallback or explicitly defined props take precedence).

2. In a component compiled with `fallthroughAttributes`, esp. when consider the previous point where missing mandatory props are required on wrapper component, vue-tsc should not display an error because of missing required properties on `I` (because those props will fallthrough from `O` usage). I currently use `<!-- @vue-ignore -->` to mute it but this is a heavy hammer: any other legit vue error is also ignored.

3. The type generation should omit props from `O` that hide existing props from `I`, as at runtime they will not fallthrough. It creates TS error when the types are not compatible: for example `I` has `a: number`, `O` has `a: string`, then when using `<O a="test">` TS complains that `"test"` is not an assignable to `number`.

In my opinion number (3) is the most annoying issue because it creates a lot of errors in the codebase that can't be avoided.

(1) is not actually creating TS error (although DX will not warn about potential runtime errors), and (2) can be ignored with a single comment in the wrapper template.

No visible change in api or options.
All changes impact the props type generation, as described above.

KazariEX:
resolve #5882

Consider the following structure:

```vue
<!-- basic.vue -->
<script lang="ts" setup>
  defineProps<{
    foo: string;
    bar: string;
  }>();
</script>
```

```vue
<!-- comp.vue -->
<script lang="ts" setup>
  import Basic from "./basic.vue";
</script>

<template>
  <Basic />
</template>
```

When the `fallthroughAttributes` option is enabled on `Comp`, `Comp` inherits all props from `Basic`, allowing us to get prop completions for `Basic` when using `Comp`.

By default, however, the props inherited from `Basic` are all optional. This means we cannot enforce passing those required props when using `Comp`. At the same time, even if the required props from `Basic` are passed to `Comp`, `Basic` will still report missing prop errors.

After enabling the new `checkRequiredFallthroughAttributes` option on `Comp`, `Comp` will inherit `Basic`'s props exactly as they are, and the missing prop errors on `Basic` inside `Comp` will disappear. This is because doing so is equivalent to promising that these required props should be passed when using `Comp`. Additionally, if some of the props required by `Basic` have already been passed directly to `Basic`, those props will be excluded from the inherited props.


packages/language-core/lib/codegen/template/element.ts
```diff
@@ -209,8 +209,16 @@ export function* generateComponent(
...
 	}
 
+	const shouldInheritAttrs = hasVBindAttrs(options, ctx, node);
+
 	yield `(`;
-	const token2 = yield* startBoundary('template', startTagOffset, codeFeatures.verification);
+	const token2 = yield* startBoundary(
+		'template',
+		startTagOffset,
+		shouldInheritAttrs && options.vueCompilerOptions.checkRequiredFallthroughAttributes
```
serkodev, line 220: Another issue, here disables verification for every component props call whenever `fallthroughAttributes && checkRequiredFallthroughAttributes` is true.
That may hides real missing-required-prop errors for components that are not fallthrough targets? e.g. `<div><basic /></div>` where `basic` requires bar passes typecheck.

```vue
<!-- main.vue -->
<script setup lang="ts">
import child from "./child.vue";
</script>

<template>
  <child />
</template>
```

```vue
<!-- child.vue -->
<!-- @checkRequiredFallthroughAttributes true -->

<script setup lang="ts">
import basic from "./basic.vue";
</script>

<template>
  <div>
    <!-- should error here? -->
    <basic />
  </div>
</template>

```

serkodev (commented):
In child.vue, if some props such as bar are already passed directly to the inner component, should bar still be required again when using the child component from main.vue? I would expect props already supplied by the wrapper to be excluded from the required fallthrough attrs.

```vue
<!-- child.vue -->
<!-- @checkRequiredFallthroughAttributes true -->

<script setup lang="ts">
import basic from './basic.vue';
</script>

<template>
  <basic bar="fixed" />
</template>
```

```vue
<!-- main.vue -->
<template>
  <!-- should not require bar prop here? -->
  <child />
</template>
```

packages/language-core/lib/codegen/template/element.ts
KazariEX: My mistake, now it's `shouldInheritAttrs` instead of `vueCompilerOptions.fallthroughAttributes`.

PR #6065, opened 2026-05-23
fix(vscode): do not mark trailing slash in capitalized self-closing tags as invalid

### LINKED ISSUE #6063: Fix syntax highlight for self closing components (<Head />, <Button /> etc...)

Invalid void elements have a red syntax highlighting.
The problem is for components that share the same name (but in PascalCase) also have a red syntax highlighting like if they were invalid, but they aren't. Self-closing in perfectly normal and expected for components.

Image

This has been discussed previously here: https://github.com/vuejs/language-tools/issues/1317

The regex should be case-sensitive, so only `<head />` and not `<Head />` is highlighted.

suisanka:
This PR makes syntax highlight for self-closing custom components normal like void elements.

Close #6063.


extensions/vscode/syntaxes/vue.tmLanguage.json
```diff
@@ -1176,6 +1176,28 @@
 				{
 					"include": "#template-tag"
 				},
+				{
+					"begin": "(<)([A-Z][a-zA-Z0-9]*)(?![a-zA-Z0-9.:-])(?=([^>]*/>))",
```
KazariEX:
```html
<Head a=">"/>
<!--     ^? -->
```

Issue #5341, opened 2025-04-27
[vue-component-meta] JSDoc tags in defineEmits are not included in generated EventMeta


oljc:
### Vue - Official extension or vue-tsc version

-

### VSCode version

-

### Vue version

3.5.13

### TypeScript version

5.8.3

### System Info

```shell

```

### package.json dependencies

```json

```

### Steps to reproduce

Define defineEmits in Vue’s <script setup> syntax as follows:
```ts
const emit = defineEmits<{
  /**
   * @zh 点击时触发
   * @en Emitted when the link is clicked
   */
  (e: 'click', ev: MouseEvent): void;
  /**
   * @zh 用户点击链接时触发
   * @en Triggered when the user clicks on the link
   */
  (e: 'select', hash: string | undefined, preHash: string): void;
}>();
```

<img width="539" alt="Image" src="https://github.com/user-attachments/assets/d186fd9c-9190-4ae3-aaf2-ba10fe0a37a3" />

Parse the output：
```js
[
  {
    name: 'click',
    description: '',
    tags: [],
    type: '[ev: MouseEvent]',
    rawType: undefined,
    signature: '(event: "click", ev: MouseEvent): void',
    declarations: [Getter],
    schema: [Getter]
  },
  {
    name: 'select',
    description: '',
    tags: [],
    type: '[hash: string, preHash: string]',
    rawType: undefined,
    signature: '(event: "select", hash: string, preHash: string): void',
    declarations: [Getter],
    schema: [Getter]
  }
]
```

### What is expected?

`Tags` in JSDoc comments for defineEmits should be extracted correctly, just like they are for Props.

### What is actually happening?

does not correctly extract tags from JSDoc comments when parsing components.

KazariEX:
Which version of `vue-component-meta` did you use? I cannot reproduce this.

![Image](https://github.com/user-attachments/assets/ffc5529f-6326-474c-9585-474bf547056b)

oljc:
@KazariEX I’m using the latest version.
repro:  https://stackblitz.com/edit/vitejs-vite-ajyz7qdb?file=scripts%2Fdocgen.ts
You can add Terminal run `npm run docgen` to see the issue.

KazariEX:
Yes I can reproduce it at stackblitz but cannot do it locally. I don't know how to investigate it.

![Image](https://github.com/user-attachments/assets/fdbb5c6a-7ddd-4319-a76c-3c7dace12ce9)

[...]

PR #6066, opened 2026-05-24
feat(language-service): ✨ add rich hover alternatives

jak-ch-ll:
Hello,

a while ago the Vue language tools got a new feature in form of [rich hover documentation](https://github.com/vuejs/language-tools/pull/5881). This is a pretty useful feature, but because it is based on HTML it does not work in terminal editors like Neovim. I wanted to change that and created two possible alternatives (more on those variants below).

## Notes

- I refactored the existing code structure to make it easier to extend
- While doing that I tried leaving the core table rendering logic as is
- I created a hover playground in the `test-workspace` to test the different possibilities
  - It includes a `<child>` and a `<minimal>` ; I created the latter for screenshots and intend to delete it before merge
  - Note: The description of events is not shown, but this is not a bug in the rendering, but instead in the meta (see #5341)
- The playground also helped me find a small bug in the table rendering triggered by code blocks at the end of a JSDoc comment

## Open points

- [ ] Changes compared to the table
  - Models are their own category
    - Reasoning:
      - Models are 'special' and will usually be used differently than props
      - Reduces the overall size because models aren't doubled between props and events
      - I don't think the hover docs need to 'teach' that a model is actually a prop and and event
  - Event props (props starting with `onX`) are moved into the event list
    - Reasoning:
      - The auto-complete of Vue will also complete these with `@` or `v-on`
  - @johnsoncodehk As the creator of the original table I would like to hear your opinion on these changes and if I should integrate these into the table as well
- [ ] File structure
  - Because the `vue-template.ts` file is quite long, I opted for moving most of the logic concerning this component hover into a separate file to make it a bit easier to digest
  - Does this and the resulting structure fit with the project convetions?
- [ ] Which variant to go with? Or keep both?
- [ ] How to name the settings options?
  - I don't no how clear these names are. At least "Markdown" might be a bit confusing because technically all of them are Markdown
- [ ] My intention is to cleanup the commit history before merge. Any preferences for the final PR? Everything squashed or separate commits for refactoring, bugfix and feature?

### 1. Markdown

I originally started with the markdown version and ran into the challenge with VSCode that the descriptions were slightly closer to the following meta element than to the element they belonged to. This made it confusing to see for which meta the description would actually be, especially because we are used to description on top because of JSDoc. I did not find a way with pure Markdown to create more space, which is why I decided to indent the description via blockquote (`>`).

<details><summary>Screenshot VSCode</summary>
image
</details> 

<details><summary>Details</summary>
<p>

```markdown
## Models

`modelValue?: string | undefined = "default value"`

> This is a model

## Props

`requiredProp: string`

> This is a requird prop

`withMarkdown?: string | undefined = "and default value"`

> With some markdown
> 
> # Who puts a title in a jsdoc comment?
> 
> > This is a blockquote
> 
> - This is a list item
> - This is another list item
> 
> ```ts
> const foo: string = 'bar'
> ```

~~`deprecatedProp?: string | undefined`~~

> 
> 
> ***@deprecated*** This prop is deprecated and should not be used

## Events

`@customEvent: () => any`

`@eventWithPayload: (payload: string, payload2: number) => any`

## Slots

`#default` - `{ foo: string; }`

> Default slot with props

## Exposed

`exposedMethod: () => void`

> This is a method that can be called from the parent component
```

</p>
</details> 

Upsides:
- Deprecations can be marked with ~~strikethrough~~
- Allows for Markdown rendering of the descriptions

Downsides:
- Descriptions are Markdown rendered, which can mess with the structure a bit, like with the heading in the example
- The props themselves are mono colored

### 2. JSDoc

I then realized that Markdown supports code blocks, so I put everything into TS codeblock and added the descriptions as "JSDoc comments", which resulted in the second variant.

<details><summary>Screenshot VSCode</summary>
image
</details> 

<details><summary>Resulting Markdown</summary>
<p>

```markdown
## Models

~~~ts
/**
 * This is a model
 */
modelValue?: string | undefined = "default value"
~~~

## Props

~~~ts
/**
 * This is a requird prop
 */
requiredProp: string

/**
 * With some markdown
 * 
 * # Who puts a title in a jsdoc comment?
 * 
 * > This is a blockquote
 * 
 * - This is a list item
 * - This is another list item
 * 
 * ```ts
 * const foo: string = 'bar'
 * ```
 */
withMarkdown?: string | undefined = "and default value"

/**
 * @deprecated This prop is deprecated and should not be used
 */
deprecatedProp?: string | undefined
~~~

## Events

~~~ts
@customEvent: () => any

@eventWithPayload: () => any
~~~

## Slots

~~~ts
/**
 * Default slot with props
 */
#default: { foo: string; }
~~~

## Exposed

~~~ts
/**
 * This is a method that can be called from the parent component
 */
exposedMethod: () => void
~~~
```

</p>
</details> 

Upsides:
- Props are colored (albeit not 100% correct compared to code)
- Looks and reads similar to the documentation inside of a component
- `@tags` are colored the same as in JSDoc

Downsides:
- The description is otherwise not formatted

### Conclusion

I personally have a light preference for the JSDoc version, because I think it puts a stronger emphasize on the meta information (name, type, default) instead of the description, which is what I want from this kind of documentation. To get the fully rendered description I can always open the hover docs for just that specific meta element.

But ultimately I think both variants are fine. I even asked my colleagues but there were proponents for both versions, so it might just come down to taste, which would be an argument to keep both and let the users decide themselves.


extensions/vscode/package.json
```diff
@@ -299,8 +299,14 @@
...
 				"vue.hover.rich": {
-					"type": "boolean",
-					"default": false,
+					"type": "string",
+					"enum": [
+						"off",
+						"Table",
+						"JSDoc",
+						"Markdown"
+					],
```
KazariEX, line 308: It is recommended to use lowercase.

test-workspace/tsc/hover/tsconfig.json
KazariEX, line 1: This project is not for type checking, so it doesn't need to be added.

packages/language-service/lib/plugins/vue-template/componentHoverDocs.ts
```diff
@@ -0,0 +1,413 @@
+import type { LanguageServiceContext } from '@volar/language-service';
+import type { ComponentMeta, EventMeta, ExposeMeta, PropertyMeta, SlotMeta } from '../../../../component-meta';
```
KazariEX: Should be imported from `vue-component-meta` instead.

extensions/vscode/package.json
jak-ch-ll: Changed all options to lowercase

test-workspace/tsc/hover/tsconfig.json
jak-ch-ll: Removed

packages/language-service/lib/plugins/vue-template/componentHoverDocs.ts
jak-ch-ll: And changed. Thanks for the review!

KazariEX:
I think we can keep only the `jsdoc` and `table` formats. The current `markdown` format doesn't provide the dense information display of the `table` format, nor does it feel as intuitive for js/ts as the `jsdoc` format. Rather than maintaining multiple sets of similar logic, I think it would be more valuable to focus on polishing the `jsdoc` format.

KazariEX:
For the `jsdoc` format, one idea I have is to put each kind of metadata inside a synthetic interface, as if it actually existed:

```ts
interface Props {
  /**
   * @description foo
   * @default "..."
   */
  foo?: string;
  /**
   * @deprecated
   */
  bar?: boolean;
}
```

This would make the output feel closer to how JS/TS developers already read and write type information, while still leaving room to present metadata in a structured way.

This approach would also avoid the styling friction caused by rendering markdown text and code text together.

jak-ch-ll:
> For the `jsdoc` format, one idea I have is to put each kind of metadata inside a synthetic interface, as if it actually existed:
> 
> ```ts
> interface Props {
>   /**
>    * @description foo
>    * @default "..."
>    */
>   foo?: string;
>   /**
>    * @deprecated
>    */
>   bar?: boolean;
> }
> ```
> 
> This would make the output feel closer to how JS/TS developers already read and write type information, while still leaving room to present metadata in a structured way.
> 
> This approach would also avoid the styling friction caused by rendering markdown text and code text together.

I played around with that idea. The first version just wraps everything in interfaces, but keeps the default and prefixes (`@` and `#`) as is:

<details>
  <summary>Fake interfaces</summary>
  image
</details>

The second version uses fully legal interfaces and the `@default` tag:

<details>
  <summary>Valid interfaces</summary>
  image
</details>

I kinda like `@` and `#` because they immediately make it clear that these are events/slots. But I can live without them if we want to keep the syntax valid.

For default I can see arguments for both sides. The one downside of `@default` that I see is that it requires more space, especially if there would otherwise not be a JSDoc comment. But on the plus side it makes the interfaces valid.

I pushed my changes, if you wanna try it. The code needs some clean up; I will do that once we decided with which solution to move forward.

KazariEX:
I think using these template syntax specific characters to distinguish them is indeed a good idea, but I still lean toward using fully valid jsdoc to reduce potential cognitive friction.

A few other minor changes:
- Use named tuples to represent emits
- Remove blank lines between properties

PR #6088, opened 2026-06-04
fix(language-core): only exclude already-set props from inherited attrs when `checkRequiredFallthroughAttributes` is enabled

### LINKED ISSUE #6082: Regression: `fallthroughAttributes` now excludes attributes inherited from a child component with props set

3.3.3

### VSCode version

1.122.0

### Vue version

3.5.35

### TypeScript version

6.0.3

### System Info

```shell
  System:
    OS: Windows 10 10.0.19045
    CPU: (12) x64 Intel(R) Core(TM) i7-9750H CPU @ 2.60GHz
    Memory: 3.00 GB / 15.85 GB
  Binaries:
    Node: 26.2.0 - C:\Program Files\nodejs\node.EXE
    npm: 11.13.0 - C:\Program Files\nodejs\npm.CMD
    pnpm: 11.1.3 - C:\Users\grant\AppData\Roaming\npm\pnpm.CMD
  Browsers:
    Chrome: 148.0.7778.179
    Edge: Chromium (140.0.3485.66)
    Firefox: 150.0.1 - C:\Program Files\Mozilla Firefox\firefox.exe
    Internet Explorer: 11.0.19041.5794
```

### package.json dependencies

```json
{
  "devDependencies": {
    "typescript": "^6.0.3",
    "vue": "^3.5.35",
    "vue-tsc": "^3.3.3"
  }
}
```

### Steps to reproduce

1. Clone the minimal reproduction linked below.
2. Run `npm i`.
3. Run `npm run check`, or open `App.vue` in VS Code with the Vue extension installed.

### What is expected?

There should not be any type errors.

```
> vue-tsc -b --noEmit

App.vue:6:11 - error TS2353: Object literal may only specify known properties, and 'optional' does not exist in type '{ readonly style?: unknown; readonly key?: PropertyKey | undefined; readonly ref?: VNodeRef | undefined; readonly class?: unknown; readonly ref_for?: boolean | undefined; ... 6 more ...; readonly onVnodeUnmounted?: VNodeMountHook | ... 1 more ... | undefined; } & VNodeProps & AllowedComponentProps & ComponentCustomP...'.

6   <Parent optional="override" />
            ~~~~~~~~

App.vue:7:11 - error TS2353: Object literal may only specify known properties, and 'required' does not exist in type '{ readonly style?: unknown; readonly key?: PropertyKey | undefined; readonly ref?: VNodeRef | undefined; readonly class?: unknown; readonly ref_for?: boolean | undefined; ... 6 more ...; readonly onVnodeUnmounted?: VNodeMountHook | ... 1 more ... | undefined; } & VNodeProps & AllowedComponentProps & ComponentCustomP...'.

7   <Parent required="override" />
            ~~~~~~~~


Found 2 errors.
```

https://github.com/GrantGryczan/vue-language-tools-issue

### Any additional comments?

I suspect this regression is related to the addition of the `checkRequiredFallthroughAttributes` Vue compiler option, but that's just a guess. That option doesn't need to be enabled for the bug to occur. This wasn't an issue a few months ago.

KazariEX:
fix #6082

### DIFF
--- a/packages/language-core/lib/codegen/template/element.ts
+++ b/packages/language-core/lib/codegen/template/element.ts
@@ -262,9 +262,14 @@ export function* generateComponent(
 	}
 
 	if (shouldInheritAttrs) {
-		const restsVar = ctx.getInternalVariable();
-		yield `var ${restsVar} = ${names.omit}(${getPropsVar()}, {\n${propsStr}})${endOfLine}`;
-		ctx.inheritedAttrVars.add(restsVar);
+		if (options.vueCompilerOptions.checkRequiredFallthroughAttributes) {
+			const restsVar = ctx.getInternalVariable();
+			yield `var ${restsVar} = ${names.omit}(${getPropsVar()}, {\n${propsStr}})${endOfLine}`;
+			ctx.inheritedAttrVars.add(restsVar);
+		}
+		else {
+			ctx.inheritedAttrVars.add(getPropsVar());
+		}
 	}
 
 	yield* generateStyleScopedClassReferences(options, node);


GrantGryczan:
Does this really fix the issue? It looks like if `checkRequiredFallthroughAttributes` is enabled, the same bug would occur, since the code is the same. The behavior described in #6082 is not the correct behavior with `checkRequiredFallthroughAttributes` enabled either.

KazariEX:
When this option is enabled, inheriting all props from the inner component unchanged can still report required props as missing when the wrapper component is used, even if those required props have already been passed through. Excluding them is therefore reasonable.

If you still need these props to be externally overridable, define them explicitly.

GrantGryczan:
I thought they would simply be changed the optional props, no? That was the impression I got from the documentation.

KazariEX:
If it becomes optional and a prop with the same name is also explicitly defined, the two prop types will be merged into a union. This behavior was mentioned in https://github.com/vuejs/language-tools/issues/5882#issuecomment-4464054704.

GrantGryczan:
I see. I'm not sure that's ideal, but I suppose I'll find out if it actually causes any problems once this patch releases, since I do have that option enabled in my project. As long as the errors I indicated in my issue still go away, it's probably fine, at least for now.

In any case, thanks so much for the quick response! I really appreciate how well-maintained you keep this library. :)

KazariEX:
This is an artifact of the current codegen colliding with a TS quirk. It would of course be nice to avoid it, but doing so could introduce significantly more complex codegen logic and potential TS performance issues, so I chose to sidestep it.

KazariEX:
Thanks for your understanding :)

GrantGryczan:
Yeah, I updated and it looks like I was right. The error still occurs when `checkRequiredFallthroughAttributes` is enabled. Can you not somehow use `Omit<X, keyof Y> & Y`, where `X` is the inherited props and `Y` is the explicit props? That way props with the same name wouldn't merge. I'm not familiar with the implementation. But at the very least, I suspect a non-asymptotic performance degradation isn't a huge concern since this is an opt-in feature, and performance was already the documented reason you wouldn't want to enable `fallthroughAttributes` by default in the first place.

I can open a new issue for this if desirable.

KazariEX:
Feel free to open a new one.

PR #6126, opened 2026-07-20
feat(language-core): support vapor directives

liangmiQwQ:
The current template codegen treats every custom directive as a VDOM ObjectDirective. In Vue 3.6 Vapor mode, custom directives instead receive the element, a value getter, the argument, and modifiers, which causes vue-tsc to type-check an invalid generated call.

This PR generates the Vapor call shape for custom directives while preserving the existing VDOM path for built-in directives. Vapor mode is resolved from SFC vapor attributes or the new global `vueCompilerOptions.vapor` option. (Vapor mode can be enabled by configuring it in `vite.config.ts`.)

🤖 Generated with Codex


liangmiQwQ:
@codex review

test-workspace/tsc/directives/vapor.vue
```diff
@@ -0,0 +1,17 @@
+<script vapor lang="ts">
```
KazariEX: It is enough to use a single test sfc with a `<!-- @vapor true -->` top-level comment.

packages/language-core/lib/codegen/template/index.ts
```diff
@@ -19,6 +19,7 @@ export interface TemplateCodegenOptions {
 	slotsAssignName?: string;
 	inheritAttrs: boolean;
 	componentName: string;
+	isVapor: boolean;
```
KazariEX: Put these after the `template` property.

liangmiQwQ: Handled in 0dd0aa8

test-workspace/tsc/directives/vapor.vue
liangmiQwQ: Handled in 70589c5

packages/language-core/lib/codegen/template/elementDirectives.ts
```diff
@@ -100,25 +119,34 @@ function* generateArg(
 			`)`,
 		);
 	}
-	yield `, `;
+	if (asBindingProperty) {
+		yield `, `;
```
KazariEX: Put the generation of `, ` outside.

packages/language-core/lib/codegen/template/elementDirectives.ts
```diff
@@ -29,13 +29,26 @@ export function* generateElementDirectives(
...
+			yield `(null!, `;
+			yield* generateValue(options, ctx, prop, false);
+			yield `, `;
+			yield* generateArg(options, ctx, prop, false);
+			yield `, `;
+			yield* generateModifiers(options, ctx, prop, 'modifiers', false);
+			yield `)`;
+		}
+		else {
+			yield `${names.asFunctionalDirective}(`;
```
KazariEX, line 42: I'm wondering if we can modify `asFunctionalDirective` to change its return value into the same form as `VaporDirective` 🧐

PR #6127, opened 2026-07-21
fix(language-service): prevent crash on malformed html.customData configuration

WangKepeople:
## Description
This change makes the handling of `html.customData` more defensive by validating the value returned from `getConfiguration()` before iterating over it.

Instead of assuming the returned value is always a `string[]`, the code now validates it with `Array.isArray()`, falls back to an empty array when the returned value is not an array, and filters out falsy entries before iterating over the paths.

## Background
While investigating a crash reported as

```
TypeError: paths is not iterable
```

I found that the LSP client replies to `workspace/configuration` with:

```json
[null]
```

However, inside `vue-template.ts`, `context.env.getConfiguration?.("html.customData")` unexpectedly returns an empty object (`{}`) instead of `null` or an array.

The exact source of this conversion is still under investigation, but this PR prevents the language service from crashing regardless of where the malformed value originates.

## Changes
- Validate the result using `Array.isArray()`
- Fall back to `[]` when the returned value is not an array
- Filter out falsy entries with `.filter(Boolean)` before iteration

### DIFF
--- a/packages/language-service/lib/plugins/vue-template.ts
+++ b/packages/language-service/lib/plugins/vue-template.ts
@@ -867,8 +867,9 @@ export function create(
 			}
 
 			async function getCustomData() {
-			  const rawPaths = await context.env.getConfiguration?.('html.customData');
-			  const paths: string[] = Array.isArray(rawPaths) ? rawPaths.filter(Boolean) : [];
+				// Some LSP clients represent unset configuration as `[null]`
+				const raw = await context.env.getConfiguration?.('html.customData');
+				const paths: string[] = Array.isArray(raw) ? raw.filter(Boolean) : [];
 				const customData: html.IHTMLDataProvider[] = [];
 				for (const path of paths) {
 					for (const workspaceFolder of context.env.workspaceFolders) {
--- a/packages/language-service/lib/plugins/vue-template.ts
+++ b/packages/language-service/lib/plugins/vue-template.ts
@@ -867,7 +867,8 @@ export function create(
 			}
 
 			async function getCustomData() {
-				const paths: string[] = await context.env.getConfiguration?.('html.customData') ?? [];
+			  const rawPaths = await context.env.getConfiguration?.('html.customData');
+			  const paths: string[] = Array.isArray(rawPaths) ? rawPaths.filter(Boolean) : [];
 				const customData: html.IHTMLDataProvider[] = [];
 				for (const path of paths) {
 					for (const workspaceFolder of context.env.workspaceFolders) {
--- a/packages/language-service/lib/plugins/vue-template.ts
+++ b/packages/language-service/lib/plugins/vue-template.ts
@@ -867,7 +867,7 @@ export function create(
 			}
 
 			async function getCustomData() {
-				// Some LSP clients represent unset configuration as `[null]`
+				// some LSP clients represent unset configuration as `[null]`
 				const raw = await context.env.getConfiguration?.('html.customData');
 				const paths: string[] = Array.isArray(raw) ? raw.filter(Boolean) : [];
 				const customData: html.IHTMLDataProvider[] = [];


packages/language-service/lib/plugins/vue-template.ts
```diff
@@ -867,7 +867,8 @@ export function create(
 			}
 
 			async function getCustomData() {
-				const paths: string[] = await context.env.getConfiguration?.('html.customData') ?? [];
+			  const rawPaths = await context.env.getConfiguration?.('html.customData');
+			  const paths: string[] = Array.isArray(rawPaths) ? rawPaths.filter(Boolean) : [];
```
KazariEX:
```suggestion
				// Some LSP clients represent unset configuration as `[null]`
				const raw = await context.env.getConfiguration?.('html.customData');
				const paths: string[] = Array.isArray(raw) ? raw.filter(Boolean) : [];
```

packages/language-service/lib/plugins/vue-template.ts
```diff
@@ -867,7 +867,9 @@ export function create(
 			}
 
 			async function getCustomData() {
-				const paths: string[] = await context.env.getConfiguration?.('html.customData') ?? [];
+				// Some LSP clients represent unset configuration as `[null]`
```
KazariEX:
```suggestion
				// some LSP clients represent unset configuration as `[null]`
```

KazariEX:
The error report seems to indicate that `paths` itself is not iterable. If it were `[null]`, it should not produce `TypeError: paths is not iterable`.

WangKepeople:
Thanks for pointing that out — you're right.

If the value were actually `[null]`, it would still be iterable and wouldn't explain a `TypeError: paths is not iterable`.

This change is intended to make the code more robust against unexpected values returned by `getConfiguration()`.

The `Array.isArray(rawPaths)` check guarantees that `paths` is always an array before iteration, while `filter(Boolean)` additionally filters out falsy entries (such as `null`) that could otherwise be passed to `getCustomData()`.

I'll update the PR description to better reflect this distinction.

KazariEX:
I need to know under what situations `html.customData` will return `[null]` (reproduction steps), otherwise I have no way of knowing what actual scenario this PR solves.

WangKepeople:
Thanks for pointing this out.

I investigated this further and found that the actual issue is not `[null]`.

From the LSP trace, the client replies to

```json
workspace/configuration
```

with

```json
[null]
```

which is expected.

However, inside `vue-template.ts`, the value returned by

```ts
await context.env.getConfiguration?.("html.customData")
```

is actually an empty object (`{}`), rather than `null` or an array.

I verified this by instrumenting both `vue-template.ts` and Volar's configuration layer. The value reaching `getCustomData()` has:

```
typeof === "object"
constructor === Object
Array.isArray(...) === false
Object.keys(...) === []
```

which eventually causes

```
TypeError: paths is not iterable
```

when the code attempts to iterate over it.

At the moment, I haven't yet determined where this `{}` is introduced, since the LSP response itself is `[null]`. It may be caused by an interaction between the client and the configuration abstraction, and I'm still investigating that part.

Regardless of the exact origin, the language service currently assumes the returned value is always iterable. This PR simply guards against unexpected values so that the language server won't crash.

I'm still investigating why `getConfiguration()` returns `{}` instead of reflecting the LSP response, and if the root cause turns out to be elsewhere, I'd be happy to submit a fix there instead.

KazariEX:
What lsp client are you using?

WangKepeople:
I'm using Eglot (GNU Emacs) with `eglot-typescript-preset`.

Eglot replies to `workspace/configuration` with `[null]`, but `context.env.getConfiguration?.("html.customData")` eventually becomes `{}` in `vue-template.ts`.

I'm still investigating where that conversion happens.

PR #6131, opened 2026-07-24
feat(component-meta): expose runtime values of enum members in schema

valentinpalkovic:
TS enum members currently resolve to their qualified name only (`"MyEnum.Small"`), because `resolveSchema` falls through to `typeToString`. The runtime value is lost, even though the type checker has it at that exact spot (`subtype.value` on the literal type).

Consumers cannot recover the values without running their own type checker over the user's project. For docs tooling this matters: Storybook, for example, generates a select control from enum schemas, and with only the member names available it either injects the literal string `"MyEnum.Small"` into the component (wrong value) or has to drop the control entirely (storybookjs/storybook#35565).

This PR adds a `literal` schema node for enum member types:

```ts
// enum MyStringEnum { Small = 'small', ... }  →  prop: MyStringEnum
{
  kind: 'enum',
  type: 'MyStringEnum',
  schema: [
    { kind: 'literal', type: 'MyStringEnum.Small', value: '"small"' },
    { kind: 'literal', type: 'MyStringEnum.Medium', value: '"medium"' },
    { kind: 'literal', type: 'MyStringEnum.Large', value: '"large"' },
  ],
}
```

- `value` is printed like literal-union members already are: strings quoted, numbers bare.
- Scoped to `TypeFlags.EnumLiteral` literal types only: Plain literal unions (`'a' | 'b'`) keep their existing string form, so existing consumers are unaffected there.
- `PropertyMetaSchema` gains one variant: `{ kind: 'literal'; type: string; value: string }`.
- Tests: string enum added to the `reference-type-props` fixture with a new assertion; the numeric enum snapshot and the tsc dts snapshot are updated accordingly.

### DIFF
--- a/packages/component-meta/lib/schemaResolvers.ts
+++ b/packages/component-meta/lib/schemaResolvers.ts
@@ -335,11 +335,14 @@ export function createSchemaResolvers(
 			// enum members stringify to their qualified name (e.g. "MyEnum.Small"), which
 			// hides the runtime value from consumers such as docs generators; expose the
 			// value alongside, printed like any other literal
-			return {
-				kind: 'literal',
-				type,
-				value: typeof subtype.value === 'string' ? `"${subtype.value}"` : String(subtype.value),
-			};
+			const { value } = subtype;
+			if (typeof value === 'string' || typeof value === 'number') {
+				return {
+					kind: 'literal',
+					type,
+					value: typeof value === 'string' ? `"${value}"` : String(value),
+				};
+			}
 		}
 
 		return type;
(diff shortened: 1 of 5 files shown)


KazariEX (approved):

packages/component-meta/lib/schemaResolvers.ts
```diff
@@ -335,11 +335,14 @@ export function createSchemaResolvers(
...
-				kind: 'literal',
-				type,
-				value: typeof subtype.value === 'string' ? `"${subtype.value}"` : String(subtype.value),
-			};
+			const { value } = subtype;
+			if (typeof value === 'string' || typeof value === 'number') {
+				return {
+					kind: 'literal',
+					type,
+					value: typeof value === 'string' ? `"${value}"` : String(value),
```
KazariEX: 🧐

```suggestion
					value: JSON.stringify(value),
```

PR #6135, opened 2026-07-27
feat(language-core): suppress tsgo's TS6196 on generated virtual declarations

johnsoncodehk:
tsgo reports unused type parameters as **TS6196** where stock TS reports **TS6133** (verified: tsgo intentionally re-purposed 6196 for unused type parameters — microsoft/typescript-go#3046). Volar's SFC diagnostic suppression was keyed on 6133 only, so every unused generic type param in a .vue SFC surfaced as a TS6196 error under the tsgo-backed engine (e.g. typescript-native-bridge), while stock never reports them — the same false-positive class doNotReportTs6133 already covers.

Two suppression points, matching the existing doNotReportTs6133 mechanism:

- **doNotReportTs6133 boundary** (codeFeatures.ts): also exclude 6196
- **Generated generic function's type parameter** (scriptSetup.ts): the declaration volar emits for the type machinery gets a verification feature suppressing 6133/6196 — stock never reports it, so tsgo's counterpart is suppressed identically

Verified: full suite 205/209 green against typescript-native-bridge master (which reverted its TS6196→TS6133 conversion to pristine tsgo behavior), including the vue-tsc typecheck snapshot.


packages/language-core/lib/codegen/codeFeatures.ts
```diff
@@ -114,7 +114,9 @@ export const codeFeatures = defineCodeFeatures({
 	doNotReportTs6133: {
```
KazariEX, line 114:
```suggestion
	doNotReportTs6133AndTs6196: {
```

test-workspace/tsc/#2472/generic.vue
```diff
@@ -2,6 +2,7 @@
 defineSlots<{
   activator?: (props: { isActive: boolean }) => void
 }>();
+null as unknown as T;
```
KazariEX, line 5: Just use `generic` instead of `generic="T"` + `null as unknown as T`.

packages/language-server/tests/completions.spec.ts
```diff
@@ -1120,6 +1113,10 @@ test('#5847', async () => {
 		  ],
 		}
 	`);
+	}
```
KazariEX, line 1116: The indentation is misaligned.

KazariEX:
This PR does not look correct 🤔. Tsgo still reports unused type parameters as TS6133 and reports unused class declarations as TS6196.

PR #6136, opened 2026-07-27
fix(component-meta): keep getProgramAndFile free of side effects

### LINKED ISSUE #6134: component-meta: a query for a file outside the project rebuilds the program each time
### What we ran into

We upgraded a Vue design system's Storybook from 10.4.6 to 10.5.4 and the build went from 35 seconds to just over 24 minutes. It took a while to narrow down, and it lands in `component-meta`, so I wanted to write up what I found and sanity check my reading of it with you.

The short of it: asking the checker about a file that isn't already in the TypeScript program adds that file to the project's root file set and rebuilds the program. `getExportNames` and `getComponentMeta` both go through that path, so a read ends up mutating the project, once per file, and the project only ever grows.

`packages/component-meta/lib/checker.ts`:

```ts
function getProgramAndFile(componentPath: string) {
	let program = tsLs.getProgram()!;
	let sourceFile = program.getSourceFile(componentPath);
	if (!sourceFile) {
		fileNamesSet.add(componentPath);   // permanent
		projectVersion++;                  // forces a full program rebuild
		program = tsLs.getProgram()!;
		sourceFile = program.getSourceFile(componentPath)!;
	}
	return [program, sourceFile] as const;
}
```

In 2.x there was no equivalent path, since meta came from a synthesised `<file>.meta.ts` twin registered for every file in the tsconfig. A query for anything outside the project threw `Could not find main source file` straight away and cost nothing.

### Reproduction

Self contained, nothing to clone. `npm i vue-component-meta typescript`, then `node repro.mjs [projectSize]`.

```js
import { mkdirSync, rmSync, writeFileSync } from 'node:fs';
import { join } from 'node:path';
import { createChecker } from 'vue-component-meta';

const SIZE = Number(process.argv[2] ?? 400);
const ROOT = join(process.cwd(), '.repro');

rmSync(ROOT, { recursive: true, force: true });
mkdirSync(join(ROOT, 'src'), { recursive: true });
mkdirSync(join(ROOT, 'outside'), { recursive: true });

for (let i = 0; i < SIZE; i++) {
  writeFileSync(
    join(ROOT, 'src', `C${i}.vue`),
    `<script setup lang="ts">defineProps<{ a: string; b?: number }>();</script>\n<template><div /></template>\n`
  );
}

// Not covered by the tsconfig include, exactly like a dependency's compiled output.
for (let i = 0; i < 20; i++) {
  writeFileSync(join(ROOT, 'outside', `D${i}.js`), `export default { name: 'D${i}' };\n`);
}

writeFileSync(
  join(ROOT, 'tsconfig.json'),
  JSON.stringify({ compilerOptions: { strict: true }, include: ['src/**/*'] }, null, 2)
);

const checker = createChecker(join(ROOT, 'tsconfig.json'), {});

function time(fn) {
  const t = process.hrtime.bigint();
  try { fn(); } catch {}
  return Number(Number(process.hrtime.bigint() - t) / 1e6).toFixed(1);
}

const roots0 = checker.getProgram().getRootFileNames().length;
console.log(`project size: ${SIZE} components, ${roots0} program root files\n`);

console.log('queries for files INSIDE the project:');
for (let i = 0; i < 3; i++) {
  console.log(`  C${i}.vue  ${time(() => checker.getExportNames(join(ROOT, 'src', `C${i}.vue`))).padStart(8)} ms`);
}

console.log('\nqueries for files OUTSIDE the project:');
for (let i = 0; i < 8; i++) {
  const ms = time(() => checker.getExportNames(join(ROOT, 'outside', `D${i}.js`)));
  const roots = checker.getProgram().getRootFileNames().length;
  console.log(`  D${i}.js  ${ms.padStart(8)} ms   program roots now ${roots} (+${roots - roots0})`);
}

console.log('\nrepeating the FIRST outside file (now already added):');
console.log(`  D0.js  ${time(() => checker.getExportNames(join(ROOT, 'outside', 'D0.js'))).padStart(8)} ms`);

rmSync(ROOT, { recursive: true, force: true });
```

Output on `vue-component-meta` 3.3.8, TypeScript 5.9.3, node 24.16.0:

```
project size: 400 components, 400 program root files

queries for files INSIDE the project:
  C0.vue       0.4 ms
  C1.vue       0.0 ms
  C2.vue       0.0 ms

queries for files OUTSIDE the project:
  D0.js       21.4 ms   program roots now 401 (+1)
  D1.js       19.7 ms   program roots now 402 (+2)
  D2.js       22.4 ms   program roots now 403 (+3)
  D3.js       17.1 ms   program roots now 404 (+4)
  D4.js       15.9 ms   program roots now 405 (+5)
  D5.js       18.8 ms   program roots now 406 (+6)
  D6.js       15.5 ms   program roots now 407 (+7)
  D7.js       15.2 ms   program roots now 408 (+8)

repeating the FIRST outside file (now already added):
  D0.js        0.0 ms
```

The root count climbs by one per query and never comes back down. Re-querying a file that's already been added is free, so the cost is the add and the rebuild rather than the query.

That cost scales with the size of the program, so bigger projects fare worse:

| project size | cost of one query for a file outside the project |
|---|---|
| 400 roots | about 18 ms |
| 1000 roots | about 40 ms |
| 2500 roots | about 90 ms |
| 5230 roots, real project with real dependency types | about 1030 ms |

### Where it bites us

Bundler integrations hand the checker every module in the build graph, because they can't know up front which ones are components. That's what turns a per query cost into a build time problem.

Ours is a 271 component design system. `@storybook/vue3-vite` passes 1179 distinct module ids to the checker, and 208 of them are program source files. The other 971 aren't, and 836 of those are compiled `.js` inside `node_modules`, which no tsconfig include picks up. At about a second each that's roughly 16 minutes of nothing but program rebuilds.

Same project, same machine, same module set:

| setup | build time |
|---|---|
| Storybook 10.4.6, `vue-component-meta` 2.2.12 | 35 s |
| Storybook 10.5.4, `vue-component-meta` 3.3.7 | 24 min 24 s |
| 10.5.4 with the expansion patched out locally | 41 s |

I got the last row by patching `getProgramAndFile` to throw instead of expanding, which puts the 2.x semantics back. Docgen coverage held: the same 141 components as 10.4.6, with none lost.

### Was the expansion deliberate?

This is the part I'm unsure about, and you'll know far better than I would.

#5876 looks like a good change from the outside: dropping the `component-type-helpers` dependency and moving inference onto `TypeChecker` so the core can run from the TypeScript plugin for #4931. My guess is the expansion came along as a fallback for the case where the queried file isn't in the program, since the synthesised twin used to answer that question implicitly and suddenly didn't.

As far as I can tell the plugin path doesn't go through it. `packages/typescript-plugin/lib/requests/getComponentMeta.ts` imports the low level `getComponentMeta` from `vue-component-meta/lib/componentMeta` and gets `program` and `sourceFile` handed to it by its caller, so #4931 looks unaffected either way. I might be missing a case there.

There's also `checker.updateFile(path, content)`, documented for virtual files and live editing, which does the same `fileNamesSet.add` plus `projectVersion++`. #5438 added it for the not in project case, so a caller who wants to query outside the tsconfig already has a route they control explicitly.

The thing that made me ask rather than just send a patch is your comment on #3896, which I read as the checker's file scope being the tsconfig's scope, and a miss meaning the caller passed the wrong tsconfig. If that's still how you see it, then the expansion quietly works around it and `include` no longer bounds what gets pulled in. Have I read that right?

### One possible shape

If you do want a miss to stay cheap, this puts the 2.x behaviour back and leaves `updateFile` as the way to add a file on purpose:

```diff
 function getProgramAndFile(componentPath: string) {
-	let program = tsLs.getProgram()!;
-	let sourceFile = program.getSourceFile(componentPath);
-	if (!sourceFile) {
-		fileNamesSet.add(componentPath);
-		projectVersion++;
-		program = tsLs.getProgram()!;
-		sourceFile = program.getSourceFile(componentPath)!;
-	}
+	const program = tsLs.getProgram()!;
+	const sourceFile = program.getSourceFile(componentPath);
+	if (!sourceFile) {
+		throw new Error(
+			`'${componentPath}' is not part of this checker's project. `
+			+ `Pass a tsconfig that includes it, or call updateFile() first.`
+		);
+	}
 	return [program, sourceFile] as const;
 }
```

Keeping the current default and putting the expansion behind a `MetaCheckerOptions` flag would work for bundler callers too, though it leaves the surprise in place for anyone who hasn't found the flag.

Happy to be told this is working as intended and that we should be scoping our tsconfig better. Either way, I'll gladly put a PR together for whichever shape you prefer, with a test that pins the root count across a query.

### Versions

- `vue-component-meta` 3.3.8, also confirmed on 3.3.7. Introduced in 3.2.0 by #5876.
- TypeScript 5.9.3
- node 24.16.0, macOS 26.5.2 arm64

seanogdev:
Closes #6134

Thanks for the quick reply on the issue. Here's the change you green-lit.

### What this does

`getProgramAndFile` no longer mutates the project when the queried file isn't in the program. Previously a miss added the path to `fileNamesSet`, bumped `projectVersion`, and rebuilt the program, so a read quietly grew the project and every subsequent read got slower. It now throws instead, which is the outcome 2.x produced via its synthesised `.meta.ts` twin.

```diff
 function getProgramAndFile(componentPath: string) {
-	let program = tsLs.getProgram()!;
-	let sourceFile = program.getSourceFile(componentPath);
+	const program = tsLs.getProgram()!;
+	const sourceFile = program.getSourceFile(componentPath);
 	if (!sourceFile) {
-		fileNamesSet.add(componentPath);
-		projectVersion++;
-		program = tsLs.getProgram()!;
-		sourceFile = program.getSourceFile(componentPath)!;
+		throw new Error(
+			`'${componentPath}' is not part of the project. `
+				+ `Use a tsconfig that includes it, or call \`updateFile()\` to add it.`,
+		);
 	}
 	return [program, sourceFile] as const;
 }
```

`updateFile()` still does the `fileNamesSet.add` + `projectVersion++` pair, so callers who genuinely want to query outside the tsconfig keep a route they control, which is what #5438 added it for.

### Heads up on a behaviour change

Anyone currently querying a file their tsconfig doesn't include gets an error where it used to work. That is the point of the change, but it is worth calling out in release notes. The migration is a one-liner: either widen the tsconfig, or call `updateFile(path, text)` first.

### The one thing in this repo that relied on it

Before sending this I went looking for anything depending on the old behaviour, and found exactly one spot: `test-workspace/component-meta/ts-component/component.tsx`. It sits beside a `component.ts`, and TypeScript's wildcard resolution keeps only the higher-priority extension when two files share a basename, so `include: ["**/*"]` never picked the `.tsx` up. Both `ts-component.tsx` tests were reaching it through the expansion.

I've added it via `files` in the two fixture configs, since `files` entries skip that dedup while `include` entries don't. The fixture is now genuinely part of the project, which seems closer to what the test intended anyway. Happy to take a different approach here if you'd rather not touch the fixtures.

### Tests

Three cases in a new `project scope` block, using a checker scoped to `empty-component/**/*` so a sibling fixture is legitimately out of scope:

- a miss throws, from both `getExportNames` and `getComponentMeta`
- the program's root file names are unchanged across a miss, which is the regression guard
- `updateFile()` then makes the same file queryable, so the escape hatch is covered

The first two fail without the change; the third passes either way by design. Full suite is green (209 passed, 4 skipped), and `dprint check` is clean.


packages/component-meta/tests/index.spec.ts
```diff
@@ -1829,9 +1830,50 @@ const noTsConfigChecker = createCheckerByJson(
 		'include': [
 			'**/*',
 		],
+		// `files` because a wildcard drops this in favour of its sibling `component.ts`,
+		// as `.ts` outranks `.tsx`
+		'files': [
+			'ts-component/component.tsx',
```
KazariEX: I think it could be renamed to `component-tsx.tsx` to avoid maintaining two `files`.

seanogdev: Good call, done in 04e50f0. `component.tsx` is now `component-tsx.tsx`, so the wildcard picks it up on its own and both `files` arrays are gone.

It turned out to fix more than the duplication. The two fixtures were also colliding on their declaration output, since `component.ts` and `component.tsx` both emit to `component.d.ts`, so TypeScript was suppressing emit for both. That is why the `packages/tsc` dts snapshot changes in this commit: those two entries were `undefined` and now hold real declarations.

```diff
-exports[`Input: ts-component/component.ts, Output: ts-component/component.d.ts 1`] = `undefined`;
+exports[`Input: ts-component/component.ts, Output: ts-component/component.d.ts 1`] = `
+"import { type MyProps } from './PropDefinitions';
+declare const _default: import("vue").DefineSetupFnComponent<MyProps, string[], {}, ...
```

Worth a sanity check from you that the new snapshot content is what you would expect there, since it is the first time that fixture has emitted anything.

One more upside: the two `ts-component.tsx` tests now pass against the unmodified checker as well, so they no longer depend on the behaviour this PR removes. Only the new `project scope` guards fail without it, which is what they are for.

packages/component-meta/tests/index.spec.ts
```diff
@@ -1835,3 +1836,39 @@ const noTsConfigChecker = createCheckerByJson(
...
+		checkerOptions,
+	);
+	const outsidePath = path.resolve(
+		__dirname,
+		'../../../test-workspace/component-meta/reference-type-props/component.vue',
+	);
+
+	test('throws for a component the project does not include', () => {
+		expect(() => scopedChecker.getExportNames(outsidePath)).toThrow(/is not part of the project/);
+		expect(() => scopedChecker.getComponentMeta(outsidePath)).toThrow(/is not part of the project/);
```
KazariEX:
```suggestion
		expect(() => scopedChecker.getExportNames(outsidePath)).toThrow();
		expect(() => scopedChecker.getComponentMeta(outsidePath)).toThrow();
```

packages/component-meta/tests/index.spec.ts
```diff
@@ -1835,3 +1836,39 @@ const noTsConfigChecker = createCheckerByJson(
...
+	});
+
+	test('leaves the program untouched when a component is not found', () => {
+		const before = scopedChecker.getProgram()!.getRootFileNames();
+
+		expect(() => scopedChecker.getExportNames(outsidePath)).toThrow();
+
```
KazariEX:
```suggestion
```

KazariEX (approved):

Issue #6020, opened 2026-04-22
vue-tsc fails to parse nested template literals in script setup inside a VitePress Markdown file


HowieHz:
### Vue - Official extension or vue-tsc version

vue-tsc 3.2.7

### VSCode version

N/A. Reproduced with vue-tsc CLI.

### Vue version

3.5.32

### TypeScript version

6.0.3

### System Info

```shell
System:
  OS: Windows 11 10.0.26200
  CPU: (24) x64 AMD Ryzen AI 9 HX 370 w/ Radeon 890M
  Memory: 12.04 GB / 31.12 GB
Binaries:
  Node: 25.9.0
  npm: 11.12.1
  pnpm: 10.33.0
Browsers:
  Edge: Chromium (140.0.3485.54)
```

### package.json dependencies

```json
{
  "dependencies": {
    "vue": "^3.5.32"
  },
  "devDependencies": {
    "vitepress": "2.0.0-alpha.17",
    "vue-tsc": "^3.2.7"
  }
}
```

### Steps to reproduce

1. Open this reproduction project.
2. Run `pnpm install --ignore-workspace`.
3. Run `pnpm run typecheck`.
4. Inspect `docs/index.md`.

### What is expected?

`vue-tsc` should parse this valid nested template literal without syntax errors:

```ts
const label = `Zodiac ${zodiac}${zodiacEmoji ? ` ${zodiacEmoji}` : ""}`;
```

### What is actually happening?

`vue-tsc` reports parse errors:

```text
docs/index.md(4,50): error TS1005: ',' expected.
docs/index.md(4,51): error TS1005: ',' expected.
docs/index.md(4,64): error TS1005: ',' expected.
```

### Link to minimal reproduction

https://github.com/HowieHz/vue-tsc-error/tree/main/vitepress-vue-tsc-template-string-locals

PR #6145, opened 2026-08-04
fix(language-core): do not extract sfc blocks across inline markdown codes

### LINKED ISSUE #6142: Inline Markdown `<script>` parsed as SFC by vue-tsc

vue-tsc 3.3.9 (CLI)

### VSCode version

N/A

### Vue version

3.5.40

### TypeScript version

6.0.3

### System Info

```shell
System:
  OS: Windows 11 10.0.26200
  CPU: (24) x64 AMD Ryzen AI 9 HX 370
  Memory: 9.81 GB / 31.12 GB
Binaries:
  Node: 26.5.0
  npm: 11.17.0
  pnpm: 11.18.0
Browsers:
  Edge: Chromium (151.0.4129.59)
```

### package.json dependencies

```json
{
  "dependencies": {
    "vitepress": "2.0.0-alpha.19",
    "vue": "3.5.40"
  },
  "devDependencies": {
    "typescript": "6.0.3",
    "vue-tsc": "3.3.9"
  }
}
```

### Steps to reproduce

1. Add ".md" to tsconfig include.
2. Set:
   ```json
   {
     "vueCompilerOptions": {
       "vitePressExtensions": [".md"]
     }
   }
   ```
3. Add this Markdown content:
   ```md
   Inline code: `<script type="module">`
   Closing tag: `src="/x"></script>`
   ```
4. Run:
   pnpm exec vue-tsc -b

### What is expected?

Inline HTML inside Markdown code spans remains Markdown.
vue-tsc reports no TypeScript syntax errors.

vue-tsc treats `<script>` and `</script>` inside inline Markdown code as a real SFC script block.

Example errors:
docs/example.md(...): error TS1005
docs/example.md(...): error TS1443

Plain `tsc -b` passes.

https://github.com/HowieHz/vue-tsc-error2

### Any additional comments?

Likely regression in @vue/language-core Markdown parser.

v3.2.7 masked inline code before SFC extraction.
v3.2.8+ extracts SFC blocks before masking inline code.

This fixed nested template literals in real <script setup> blocks, but causes inline `<script>` examples to be parsed as SFC.

Related:
https://github.com/vuejs/language-tools/issues/6020
https://github.com/vuejs/language-tools/pull/6024
https://github.com/vuejs/language-tools/pull/6038

lazerg:
A `<script>` tag written inside an inline Markdown code span gets picked up by SFC block extraction. So an inline code span like `<script type="module">` on one line and `</script>` on another is parsed as one real script block, and vue-tsc reports syntax errors on the Markdown file.

Inline code spans are now blanked in a scratch copy of the content, used only to locate the blocks. The block text itself still comes from the original content, so template literals inside a real `<script setup>` stay unaffected (#6020).

Fixes #6142

### DIFF
--- a/packages/language-core/lib/plugins/file-md.ts
+++ b/packages/language-core/lib/plugins/file-md.ts
@@ -11,7 +11,7 @@ const codeSnippetImportRE = /^\s*<<<\s*.+/gm;
 const sfcBlockRE = /<(script|style)\b[^>]*>([\s\S]*?)<\/\1>/g;
 const htmlTagRE = /(?<=<\/?)([a-z][a-z0-9-]*)\b[^>]*(?=>)/gi;
 const interpolationRE = /(?<=\{\{)[\s\S]*?(?=\}\})/g;
-const inlineCodeRE = /(`{1,2})[^`]+\1/g;
+const inlineCodeRE = /(`{1,2})[^`\n]+\1/g;
 const angleBracketRE = /<[^\s:]*:\S*>/g;
 
 const plugin: VueLanguagePlugin = ({ vueCompilerOptions }) => {
--- a/packages/language-core/lib/plugins/file-md.ts
+++ b/packages/language-core/lib/plugins/file-md.ts
@@ -37,9 +37,16 @@ const plugin: VueLanguagePlugin = ({ vueCompilerOptions }) => {
 				content = content.replace(pattern, match => ' '.repeat(match.length));
 			}
 
+			let maskedContent = content;
+			for (const { 0: text, index } of content.matchAll(inlineCodeRE)) {
+				maskedContent = maskedContent.slice(0, index) + ' '.repeat(text.length)
+					+ maskedContent.slice(index + text.length);
+			}
+
 			const codes: Segment[] = [];
 
-			for (const { 0: text, index } of content.matchAll(sfcBlockRE)) {
+			for (const { 0: masked, index } of maskedContent.matchAll(sfcBlockRE)) {
+				const text = content.slice(index, index + masked.length);
 				codes.push([text, undefined, index]);
 				codes.push('\n\n');
 				content = content.slice(0, index) + ' '.repeat(text.length) + content.slice(index + text.length);


packages/language-core/lib/plugins/file-md.ts
```diff
@@ -37,9 +37,16 @@ const plugin: VueLanguagePlugin = ({ vueCompilerOptions }) => {
 				content = content.replace(pattern, match => ' '.repeat(match.length));
 			}
 
+			let maskedContent = content;
+			for (const { 0: text, index } of content.matchAll(inlineCodeRE)) {
+				maskedContent = maskedContent.slice(0, index) + ' '.repeat(text.length)
+					+ maskedContent.slice(index + text.length);
+			}
+
```
KazariEX, line 45: This causes template interpolations like this to be stripped out:

```ts
`...${foo + bar}...`;
```

Also, the processing of inline codes has been deferred until after the SFC blocks have been extracted: https://github.com/vuejs/language-tools/blob/8e78c5d/packages/language-core/lib/plugins/file-md.ts#L55-L62

lazerg: Good catch. It was worse than just interpolations: since the early mask has no block boundaries yet, any unmatched backtick earlier in the doc (e.g. a contraction like "it`s") could pair with the next backtick anywhere later in the file, including across an entire <script setup> block, causing it to disappear from parsing entirely.

Restricted the mask to single-line spans (backticks can't span a newline), so it can't reach past a block boundary anymore, while still catching the inline `<script>` case this PR targets. Pushed as e7e94d7.

KazariEX: Excluding newlines does not address the issue I raised above, nor does it handle script blocks inside backticks that contain no newline characters.

lazerg: I've tried to reproduce a remaining failure and can't, so I might be missing your exact case. What I've verified against the current branch:

- The original repro from #6142 (two separate single-line inline code spans, each holding half a split script tag across two lines): still correctly produces no script block.
- Your nested-template-literal example (#6020/#6021 style, inside a real `<script setup>`): script content comes through intact, including the nested backticks.
- The stray-backtick-swallows-a-whole-block case I found while investigating your first comment: now fixed.

If interpolation stripping is about inline code in prose getting blanked from the type-checked template output, that happens in the second masking pass (`inlineCodeRE` applied to `content` after extraction) — that's pre-existing behavior from #6038, not something my early pass touches or changes.

Could you share the specific snippet that's still broken? I'd rather fix the actual case than keep guessing.

KazariEX: You are right, the first issue has indeed been resolved, but the second one remains.

image

lazerg: Ran vue-tsc directly against that exact content (script setup with the interpolation, blank line, then the single-line inline `<script> [] + {}; </script>`) on this branch and got zero diagnostics. So I'm not seeing what's underlined in your screenshot.

Could you paste the actual error message from hovering over the squiggle? Also worth knowing: extension or vue-tsc CLI, and did you rebuild after pulling the latest commit?

KazariEX: I mistakenly added `.md` to `vueCompilerOptions.extensions` 😅 It’s working now.

KazariEX (approved):

PR #6062, opened 2026-05-20
feat(vscode): transform tsserver content

KazariEX:
close #6060

PR #6149, opened 2026-08-07
feat(vscode): patch `typescript.js` via `--require` CLI option

serkodev:
This PR fixes the original `// FIXME: cannot work on read-only file system` issue from 7aeb5f3a (#6062).

Instead of modifying `typescript.js` on disk, we now use Node.js's `--require` CLI option to inject `tsserverPatch.js` into the child process. The patch is applied to `typescript.js` before it is loaded, so this also works on read-only file systems.

## Additional

This also fixes a pre-existing race condition.

Previously, the extension generated a proxy file that was later read by the forked tsserver process. If window A wrote its proxy and window B using a different TypeScript SDK, for example via `typescript.tsdk` overwrote that file during the ~20-100 ms before A's child process read it, window A's tsserver could end up running window B's TypeScript build.


extensions/vscode/rolldown.config.ts
```diff
@@ -6,6 +6,7 @@ export default defineConfig({
 	input: {
 		'extension': './src/extension.ts',
 		'reactivity-analysis-plugin': './src/reactivityAnalysisPlugin.ts',
+		'tsserverPatch': './src/tsserverPatch.ts',
```
KazariEX: Use kebab case key for case-insensitive systems.


```suggestion
		'patch-tsserver': './src/patchTsserver.ts',
```

extensions/vscode/src/patch-tsserver.ts
KazariEX, line 1: Rename to `patchTsserver.ts`.

KazariEX (approved):

PR #5769, opened 2025-11-23
perf(language-core): rewrite a subset of template node transforms

KazariEX:
fix #4539, fix #5492, fix #5899, fix vuejs/core#12258, close #5765

This change simplifies the transforms of the template AST, in particular by removing all `codegenNode` that existed only for runtime codegen purpose. This also results in an AST shape that is more favorable for language tools integration.

PR #6155, opened 2026-08-11
fix(language-core): generate full fragment props for typecheck

serkodev:
Related #5769

This PR fixed expected type check for `<template>` props.

- Fixed: `__VLS_asFunctionalElement` is split since #5845
- New feature: Generate failed expressions for handling props like `@[x.bar]="x.baz()"`

### DIFF
--- a/packages/language-core/lib/codegen/template/element.ts
+++ b/packages/language-core/lib/codegen/template/element.ts
@@ -380,6 +380,7 @@ export function* generateFragment(
 	node: CompilerDOM.ElementNode,
 ): Generator<Code> {
 	const [startTagOffset] = getElementTagOffsets(node, options.template);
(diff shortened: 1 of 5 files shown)


KazariEX (approved):

KazariEX:
I'm not sure if we should check these failed properties since they are never consumed by the runtime.

serkodev:
I think it makes sense to keep this behavior from the language tools perspective, for a few reasons:

* Before #5769, `generateElement` would also check failed expressions.
* We still perform the same checks in other dead-code scenarios, for example after `v-if="false"`.
* If a user changes `<div v-for>` to `<template v-for>`, we probably shouldn’t suddenly silently ignore type errors.

Alternatively, another approach would be to emit an error or warning telling the user that these props will never be evaluated.

Issue #5899, opened 2025-12-22
Getting error TS1005 after upgrading from v3.1.5


pablo-olmedo-poex:
### Vue - Official extension or vue-tsc version

3.2.1

### VSCode version

1.105.1

### Vue version

3.5.26

### TypeScript version

5.9.3

### System Info

```shell
System:
  OS: Windows 11 10.0.26100
  CPU: (8) x64 11th Gen Intel(R) Core(TM) i7-1165G7 @ 2.80GHz
  Memory: 2.57 GB / 15.69 GB
Binaries:
  Node: 22.17.1 - C:\Program Files\nodejs\node.EXE
  npm: 10.9.2 - C:\Program Files\nodejs\npm.CMD
Browsers:
  Chrome: 143.0.7499.146
  Edge: Chromium (140.0.3485.54)
  Firefox: 146.0 - C:\Program Files\Mozilla Firefox\firefox.exe
  Internet Explorer: 11.0.26100.1882
```

### package.json dependencies

```json
{
  "dependencies": {
    "awesome-phonenumber": "^7.2.0",
    "mitt": "^3.0.1",
    "quill": "^2.0.3",
    "vue-quilly": "^1.1.5"
  },
  "devDependencies": {
    "@eslint/compat": "^2.0.0",
    "@testing-library/jest-dom": "^6.9.1",
    "@testing-library/user-event": "^14.6.1",
    "@testing-library/vue": "^8.1.0",
    "@types/google.maps": "^3.58.1",
    "@types/jsdom": "^27.0.0",
    "@types/node": "^22.18.13",
    "@vitejs/plugin-vue": "^6.0.3",
    "@vue/test-utils": "^2.4.6",
    "eslint": "^9.39.2",
    "eslint-import-resolver-typescript": "^4.4.4",
    "eslint-plugin-import-x": "^4.16.1",
    "jsdom": "^27.3.0",
    "npm-run-all2": "^8.0.4",
    "prettier": "^3.7.4",
    "pug": "^3.0.3",
    "rimraf": "^6.0.1",
    "sass": "~1.97.1",
    "stylelint": "^16.26.1",
    "typescript": "^5.9.3",
    "vitest": "^4.0.16",
    "vue-tsc": "^3.2.1"
  },
  "peerDependencies": {
    "vue": "^3.5.26",
    "vue-i18n": "^11.2.7",
    "vue-router": "^4.6.4"
  }
}
```

### Steps to reproduce

[...]

Issue #5840, opened 2025-12-09
[Bug]: Using generics in defineProps breaks useTemplateRef return type


ayrtonandino:
### Vue - Official extension or vue-tsc version

3.1.8, 3.1.7, 3.1.6, (3.1.5 works)

### VSCode version

1.106.3

### Vue version

3.5.25

### TypeScript version

5.9.3

### System Info

```shell
System:
    OS: macOS 13.7.8
    CPU: (4) x64 Intel(R) Core(TM) i5-7360U CPU @ 2.30GHz
    Memory: 102.95 MB / 8.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Yarn: 4.12.0 - /usr/local/bin/yarn
    npm: 11.6.2 - /Users/ayrtonandinobonazzola/Library/Application Support/Herd/config/nvm/versions/node/v24.11.1/bin/npm
    pnpm: 10.17.1 - /usr/local/bin/pnpm
  Browsers:
    Edge: 143.0.3650.66
    Firefox Developer Edition: 146.0
    Safari: 18.6
```

### package.json dependencies

```json
"dependencies": {
    "vue": "3.5.25"
  },
  "devDependencies": {
    "@types/node": "^24.10.2",
    "@vitejs/plugin-vue": "^6.0.2",
    "unplugin-vue-components": "^30.0.0",
    "vite": "^7.2.7",
    "vue-tsc": "^3.1.8"
  }
```

### Steps to reproduce

To start:

   - install dependencies

```sh
npm install
```

  - run vue-tsc

```sh
npm run test
```

  - observe error on line 12 in `/src/components/Comp2.vue`

  - Comment out line 5 (`generic?: T`)

  - run test again and observe error on line 12 disappears

### What is expected?

to work

### What is actually happening?

throws error TS2339 on `src/components/Comp2.vue:12:20` 

### Link to minimal reproduction

https://github.com/ayrtonandino/vue-tsc-bug-templateRef-generics

### Any additional comments?

This bugs appears only on auto imported components, (from unplugin-vue-components), that use generics in defineProps, if manually imported it work fine.

Also this bug was not pressent in `vue-tsc <= 3.1.5`, only `3.1.6` onwards

serkodev:
Thanks for the report, the regression was introduced by #5803

jods4:
is there a nightly available somewhere? I would like to test the fix.
[...]

Issue #4822, opened 2024-09-06
Suspected TypeScript bug when globally registering component that uses script tag generics with defineModel


TristanMNorton:
### Vue - Official extension or vue-tsc version

2.1.6

### VSCode version

1.93.0

### Vue version

3.5.3

### TypeScript version

5.5.3

### System Info

```shell
System:
    OS: macOS 14.5
    CPU: (14) arm64 Apple M3 Max
    Memory: 642.34 MB / 36.00 GB
    Shell: 5.9 - /bin/zsh
  Binaries:
    Node: 20.15.1 - ~/.nvm/versions/node/v20.15.1/bin/node
    npm: 10.7.0 - ~/.nvm/versions/node/v20.15.1/bin/npm
```

### package.json dependencies

```json
{
  "name": "generics-test",
  "private": true,
  "version": "0.0.0",
  "type": "module",
  "scripts": {
    "dev": "vite",
    "build": "vue-tsc -b && vite build",
    "preview": "vite preview"
  },
  "dependencies": {
    "vue": "^3.5.3"
  },
  "devDependencies": {
    "@vitejs/plugin-vue": "^5.1.2",
    "typescript": "^5.5.3",
    "vite": "^5.4.1",
    "vue-tsc": "^2.1.6"
  }
}
```

### Main.ts
```ts
import { createApp } from 'vue'
import App from './App.vue'
import HelloWorld from './components/HelloWorld.vue'

const app = createApp(App)

app.mount('#app')

app.component('HelloWorld', HelloWorld) // Typescript error
```

### HelloWorld.vue
```vue
<script setup lang="ts" generic="T">
const modelValue = defineModel<T>()
</script>

<template>
  <pre>
    {{ modelValue }}
  </pre>
</template>
```

### What is expected?

The SFC internally is type safe and should allow for usage of generics when using defineModel.

### What is actually happening?

```
[...]

Issue #4785, opened 2024-09-02
Attrs type error


teleskop150750:
### Vue - Official extension or vue-tsc version

v2.1.4

### VSCode version

1.92.2

### Vue version

3.4.38

### TypeScript version

5.5.4

### System Info

```shell
System:
  OS: Linux 6.8 Ubuntu 24.04.1 LTS 24.04.1 LTS (Noble Numbat)
  CPU: (12) x64 12th Gen Intel(R) Core(TM) i5-12400
  Memory: 8.11 GB / 15.38 GB
  Container: Yes
  Shell: 3.7.1 - /usr/bin/fish
Binaries:
  Node: 20.15.1 - /run/user/1000/fnm_multishells/38089_1725257095122/bin/node
  npm: 10.7.0 - /run/user/1000/fnm_multishells/38089_1725257095122/bin/npm
  pnpm: 9.9.0 - /run/user/1000/fnm_multishells/38089_1725257095122/bin/pnpm
Browsers:
  Chrome: 128.0.6613.84
```

### Steps to reproduce

https://github.com/teleskop150750/vue-language-generic

pnpm type-check

### What is expected?

success

### What is actually happening?

❯ pnpm type-check

> vue-lang@0.0.0 type-check /home/teleskop150750/Documents/examples/vue/vue-lang
> vue-tsc --build --force

src/App.vue:6:36 - error TS2353: Object literal may only specify known properties, and '"aria-label"' does not exist in type '{ type: "multiple"; } & VNodeProps & AllowedComponentProps & ComponentCustomProps'.

6   <ToggleGroupRoot type="multiple" aria-label="Foo">
                                     ~~~~~~~~~~


Found 1 error.

 ELIFECYCLE  Command failed with exit code 2

### Link to minimal reproduction

https://github.com/teleskop150750/vue-language-generic

### Any additional comments?

_No response_

dten:
Also get this going from `2.0.29` to `2.1.2`

PR #6170, opened 2026-08-20
feat: content mapper (v4 alpha)

johnsoncodehk:
## Summary

Vue Language Tools **v4** type-checks `.vue` files with `tsc` itself through a **content mapper** — no separate `vue-tsc` binary.

## Quick start

```jsonc
// tsconfig.json
{
  "compilerOptions": { "noEmit": true },
  "contentMappers": [
    {
      "package": "@vue/content-mapper",
      "extensions": [".vue"],
      "options": {
        "target": 3.5
      }
    },
    // VitePress markdown, when `@vue/content-mapper-vitepress` is installed
    {
      "package": "@vue/content-mapper-vitepress",
      "extensions": [".md"]
    }
  ]
}
```

```sh
tsc --runExternalCode   # replaces `vue-tsc --noEmit`
```

Vue compiler options go directly under `options` (the top-level `vueCompilerOptions` field is gone). Add `"languageFeatures": false` under `options` for CLI-only type-checking.

`contentMappers` and `--runExternalCode` are the TypeScript content-mapper hook (microsoft/typescript-go#4712, merged). It ships in the `typescript@next` nightlies (`7.1.0-dev`) — `npm i -D typescript@next` is all you need.

## Breaking changes

- **`strictTemplates`, `strictVModel`, and all `checkUnknown*` options are removed.** Their permissive mode worked by dropping specific TypeScript error codes (e.g. TS2339) at specific generated locations. The content-mapper protocol cannot express that: `diagnosticDirectives` are `(range, policy)` tuples with no error-code field, and span maps only map-or-drop whole locations — per-code filtering is a capability the protocol deliberately omits, not one that is merely unimplemented. So the knobs are dropped, and template checking aligns with TypeScript's own behavior — `lang="ts"` is strict, `lang="js"` is non-strict unless `checkJs` / `@ts-check`.
- **Unknown components / directives are now always reported**, so they must be declared:

  ```ts
  declare module 'vue' {
    interface GlobalComponents { Foo: typeof import('./Foo.vue').default }
    interface GlobalDirectives { vTooltip: typeof import('./v-tooltip').default }
  }
  ```

  `GlobalComponents` / `GlobalDirectives` were added in **Vue 3.3**, so the minimum `target` is now 3.3 (`target: 3` is dropped).
- **VitePress markdown needs `@vue/content-mapper-vitepress`, and petite-vue is dropped.** Markdown handling left language-core with the `vitePressExtensions` option — register the new mapper for `.md` instead. Petite-vue's `.html` support (`petiteVueExtensions`) is removed rather than ported; a standalone `.html` mapper can be added the same way as the markdown one if there is demand.

## Diagnostics: from virtual comments to structured directives

vue-tsc suppressed diagnostics partly through mapping capabilities (`verification` / `shouldReport` per error code) and partly by *writing comments into the virtual file* — `@vue-expect-error` was lowered to a real `// @ts-expect-error` line. Under the content mapper the virtual file is what tsgo parses and checks, so suppression is expressed as data instead of text:

- **`@ts-ignore` / `@ts-expect-error` in `<script>` need no machinery.** The script block is embedded verbatim, so these comments are ordinary TS directives in the virtual file: tsgo applies them natively (including unused `@ts-expect-error` reporting), and surviving diagnostics map back through the span map.
- **`@vue-ignore` / `@vue-expect-error` in `<template>` are carried by `diagnosticDirectives`.** Codegen tags emitted mappings with marker phases (`anchor` / `content` / `end`); the mapper aggregates them into `[originalStart, originalLength, virtualStart, virtualEnd, policy]` tuples handed to tsgo with the transform result. tsgo drops any diagnostic whose position falls inside a directive's virtual range, and for an `Expect` directive that swallowed nothing it reports `Unused '@ts-expect-error' directive.` at the original comment in the `.vue` file. The legacy `// @ts-expect-error` emission is removed — directives are the only mechanism.
- **Codegen-owned suppressions** (the remaining synthesized `// @ts-ignore` lines listed under Status) still work natively; each is tracked for removal at the codegen source.

The protocol deliberately has no per-error-code filtering: a directive covers a span, not a code. That is why the permissive `checkUnknown*` options could not be carried over (see Breaking changes).

#### Foundation

- [x] `vue-tsc` removed
- [ ] retire `@vue/typescript-plugin` (the TS 5/6 plugin; superseded by the content mapper)
- [x] content-mapper protocol (`initialize` / `openProject` / `transform` / `closeProject`)
- [x] built-in directives typed in codegen (`vHtml` / `vText` / `vCloak` / `vMemo`; `vOnce` / `vSlot` fixed upstream in Vue 3.6)
- [x] span mappings + diagnostic directives + worker pool
- [x] unit tests (`spanMappings`, `diagnosticDirectives`, `project`, `workerPool`, `localization`)
- [x] integration tests against the `typescript@next` nightly (`build`, `dts`, `fixtures`, `corpus`)
  - `dts` covers `component-meta` + `dts/jsx-slots`, and rechecks every emitted `.d.vue.ts` as plain TS so no global `__VLS_` helper dangles (#6196, ported to the content-mapper harness)
- [x] CI runs the content-mapper tests against the npm nightly (no tsgo build)

#### Diagnostic model

- [x] unmapped codegen diagnostics eliminated at the codegen source (#6179)
- [x] code-level diagnostic filtering removed (synthetic vars consumed with `void`; unknown props/events always checked)
- [x] permissive options removed (`strictTemplates`, `strictVModel`, `checkUnknownComponents`, `checkUnknownDirectives`) → template checking follows TS/JS defaults
- [x] `@vue-expect-error` / `@vue-ignore` → `diagnosticDirectives` (legacy `// @ts-expect-error` emission removed; expect/ignore semantics are marker-driven only)
- [x] slot / template-ref synthesized ignore eliminated at source (#6190, #6191)
- [x] truncated-script parse errors resolve to the original block — the synthesized `;` terminator is mapped to the section end, so TS1109 reports at the same positions as vue-tsc (#3632)
- [x] functional directives widened at the type level (`(...args: [...P, ...any[]]) => R`) — v-cloak no longer trips a spurious TS2554, generic function directives now infer and check (#4682)
- [ ] remaining synthesized `@ts-ignore` (semantic diagnostics, exposed under content mapper):
  - [ ] auto-import candidate identifiers (`context.ts`) — TS2304, structural (auto-import needs unresolved names)
  - [ ] `__VLS_withDotValue` assertions (`template.ts`) — TS2454, needs restructure to map real uninitialized errors
  - [ ] synthetic trailing args (`elementDirectives.ts`) — TS2345 on object-form directive hooks, structural (runtime 4-arg contract); functional-branch TS2554 eliminated at the type level (`__VLS_asFunctionalDirective` widens to `(...args: [...P, ...any[]]) => R`)
     #6197
  - [ ] compound event handlers (`elementEvents.ts`) 
    - [x] TS6133 (fixable via `void $event`)
      #6193
    - [ ] TS7031/TS2493 structural
  - [ ] first component instantiation (`element.ts`) — TS2304/TS7006/TS2345, structural (instance/generic inference)
  - [ ] JS typedef guards (`localTypes.ts`) — TS2503, structural (global `JSX` namespace); #6195 closed — moving `__VLS_PropsChildren` into global types leaks into the emitted dts, needs a different approach
  - [x] scoped classes type alias (`scopedClasses.ts`) — TS6196, fixable via a non-JSDoc reference
     #6194
- [x] duplicate CSS module names reported as TS2300/TS2717 at the `module` attribute (#6198) — corpus asserts the same 9 diagnostics (`bare` / `empty` / `kebab`) at the same positions as vue-tsc
- [x] `__combineToken` chunks merged into single spans in the mapper (#6170) — camelized segments and boundary-bracketed content become one logical span, so token-wide diagnostics map back instead of resolving to a gap; corpus reaches vue-tsc parity for unknown components / props / events / directives (#2225, #2468, #2588, #2629, #2646, #2678, #2725, #3122, #3123, #3718, #4646, #4785, #4822, #5840, #5899, events, noPropertyAccessFromIndexSignature, templateRef, unknownComponent)

#### Mapping & features

- [x] `linkedCodeMap` removed (#6175)
- [x] setup bindings unwrapped by value (#6184)
- [x] template token boundary double-mapping fixed (#6178)
- [x] `shouldHighlight` / `shouldRename` → static `SpanMapFeature` bits
- [x] content mapper round 2 alignment ([TypeScript#63936](https://github.com/microsoft/TypeScript/pull/63936)): `SpanMapFeature` bits pinned to `spanmap.go`, `inferredProject` → `inferredProjectContribution`, zero-length navigation anchor on synthetic `export default` (the unmappable-navigation fallback was removed upstream)
- [x] `SpanMapKind.Alias` evaluated — not adopted: it only affects diagnostic *message* rendering (hover/navigation/codefix never read it), substituting the original name when the virtual name is mapper-private (upstream's example: lisp `add` → `+`). Every Vue respelling is a real contract name users act on (`GlobalComponents` / `GlobalDirectives` keys, the `onXxx` emit-prop convention), so diagnostics should display it verbatim
- [ ] Vue-specific completion markers — not yet carried into `SpanMapping`
- [ ] replace `@vue/typescript-plugin` request forwarding (TS 7 / content mapper has no such semantic). 17 `_vue:` requests to migrate:
  - project routing — `projectInfo`
  - native LSP (`SpanMapFeature`) — `documentHighlights-full`, `encodedSemanticClassifications-full`
  - twoslash query — `quickinfo` (InlayHints, not Hover)
  - completion (markers / completion data) — `getAutoImportSuggestions`, `getComponentProps`, `resolveAutoImportCompletionEntry`
  - template/component type-level metadata (consumed by language-server VLS) — `getComponentMeta`, `getComponentSlots`, `getComponentDirectives`, `getElementAttrs`, `getElementNames`, `getComponentNames`
  - helpers — `collectExtractProps`, `resolveModuleName`, `getImportPathForFile`, `isRefAtPosition`

#### Config & emit

- [x] mapper options flattened (Vue options live directly under `options`, no `vueCompilerOptions` layer)
- [x] minimum `target` raised to 3.3
- [x] `configIdentity` / `watchedFiles` — inferred projects and plugins declared under a tsconfig's `vueCompilerOptions.plugins` are watched and fingerprinted too
- [x] VitePress markdown decoupled from language-core into `@vue/content-mapper-vitepress` — the markdown front-end left language-core (`file-md`, `vitePressExtensions`) and is registered as its own content mapper for `.md`; the shared mapper core is exported from `@vue/content-mapper`
- [x] petite-vue support dropped (`file-html`, `petiteVueExtensions`, the `globalThis` template ctx branch) — no standalone `.html` mapper unless someone asks for one
- [ ] language server loads the language plugin of a registered content-mapper package — until then `.md` no longer resolves in the editor, since the markdown front-end is no longer a built-in language-core plugin (type checking through the mapper is unaffected)
- [ ] language-core drops file-extension matching entirely — `extensions` leaves `VueCompilerOptions`; which files are Vue is decided by the host (`contentMappers[].extensions`), which also resolves the old `options.extensions` drift coupling
  - [ ] `extensions` cannot come from `openProject` yet: the protocol passes only `{ configFileName, projectHandle, options, compilerOptions }`. Reading the tsconfig ourselves works for tsconfig projects (nearest `contentMappers` entry wins, the array is replaced wholesale through `extends`), but inferred projects have no file to read, so upstream would need to pass the entry's `extensions`

## Implementation notes

Classic `typescript` stays pinned `<7` (via `typescript-native-bridge`) to keep `tsc -b` / `tsslint` working; content-mapper tests run against the `typescript-7` npm alias (`typescript@next`, 7.1.0-dev). Upstream content-mapper development moved from `microsoft/typescript-go` (archived staging repo) to `microsoft/TypeScript` (`tsc/internal/contentmapper`).

Vue is pinned to `3.6.0-rc.6` — the rc that carries the `defineComponent({ __typeProps, __typeEmits, __typeRefs, __typeEl })` + extended `DefineComponent` API (the intermediate `DefineComponent2` is gone), and the `vOnce` / `vSlot` casing fix.

## Testing

```sh
pnpm test content-mapper
```

## References

- TypeScript PR: https://github.com/microsoft/typescript-go/pull/4712


escaton:
Just in case, I tested the tsgo + content-mapper in my private project (29367 lines of vue) last weekend before it was merged.

https://github.com/andrewbranch/language-tools/commit/76d07d927e085bb0a0e222ff5bb8cc8665d70e8a + https://github.com/andrewbranch/typescript-go/commit/d07c1fff6efd364533b7073dd87b39aaf03029c8

I found just two inconsistencies in content-mapper version vs vue-tsc@3.3.9 + typescript@6.0.3:

- v-html is not recognized. Both vue-tsc (ts6 and tnb) do not report that.
```
test.vue:8:10 - error TS2339: Property 'vHtml' does not exist on type 'GlobalDirectives'.

8     <div v-html="'foo'" />
           ~~~~~~
```
And that seems a correct error, there is indeed no `vHtml` in https://github.com/vuejs/core/blob/a2b40db9a83b36ed9da3a16403cf8f040262d73f/packages/runtime-dom/src/index.ts#L62 but somehow `vue-tsc` makes it any.

- errors on two style blocks with same `module` attrubute. This is a good catch that current vue-tsc does not report.
```html
<!-- test.vue -->
<style module="foo" critical>
.foo {
}
</style>
<style module="foo" critical>
.foo {
}
</style>
```

```html
test.vue:73:16 - error TS2300: Duplicate identifier 'foo'.

73 <style module="foo" critical>
                  ~
```

Hope that helps

escaton:
Btw typescript@next haven't yet been released after content-mappers merge. Today was skipped, probably due to typescript-go repo migration.

escaton:
Hi @johnsoncodehk would you mind publishing this as next, or v4-alpha version on npm?

johnsoncodehk:
@escaton Yes, I plan to do that, but before that, the removal of @vue/typescript-plugin needs to be completed.

Issue #6176, opened 2026-08-21
Renaming a useTemplateRef binding in :[foo] also renames the synthesized `.value`


johnsoncodehk:
## Reproduction

```vue
<template>
	<div :[foo]="123"></div>
</template>

<script setup lang="ts">
import { useTemplateRef } from 'vue'
const foo = useTemplateRef('foo')
</script>
```

Steps:

1. Put the cursor on `foo` in `:[foo]`, right at the `foo`/`]` boundary.
2. Trigger rename (F2) and rename it.

## Expected

Only the `foo` binding is renamed.

## Actual

The rename also reports a location for the `value` property — the synthesized `.value` in the generated `foo.value` — which resolves to `Ref.value` in `@vue/reactivity`.

## Root cause

For a `useTemplateRef` binding, the generated virtual code is:

```ts
[foo.value]: (123),
```

where `.value` is synthesized and emitted as a plain (unmapped) string:

```ts
if (setupRefs.has(name)) {
	yield [name, block.name, start + offset, data]; // "foo" -> source foo
	yield `.value`;                                  // synthesized, no source mapping
}
```

The closing `]` maps to the source `]` right after `foo`. Because the 6-char `.value` sits between `foo` and `]`, that `]` mapping lands in the generated code immediately after `value`:

```
source:   [ foo  ]
          |  \   \
gen:      [ foo .value ]
                      \
                       `]` maps here — right after `value`
```

When the cursor is exactly at the `foo`/`]` boundary, `getTouchingPropertyName` sees the generated position right after `value` and resolves it to the `value` identifier, so the rename follows `Ref.value` instead of `foo`.

PR #6177, opened 2026-08-21
test(language-server): add #6176 rename test

lazerg:
## Summary

Adds a failing test for #6176: renaming a `useTemplateRef` binding used as a dynamic prop name (`:[foo]`) also renames the synthesized `.value`.

The test documents the expected behavior — only the `foo` binding should be renamed — and currently fails because the rename also reports `Ref.value` from `@vue/reactivity`.

## Why test-only

No fix is included here; the intent is to capture the bug first so a future fix has a red test to turn green.

### DIFF
--- a/packages/language-server/tests/renaming.spec.ts
+++ b/packages/language-server/tests/renaming.spec.ts
@@ -647,6 +647,72 @@ test('Component dynamic props', async () => {
 	`);
 });
 
+test('#6176', async () => {
+	expect(
+		await requestRenameToTsServer(
+			'tsconfigProject/fixture.vue',
+			'vue',
+			`
+			<template>
+				<div :[foo|]="123"></div>
+			</template>
+
+			<script lang="ts" setup>
+			import { useTemplateRef } from 'vue';
+			const foo = useTemplateRef('foo');
+			</script>
+		`,
+		),
+	).toMatchInlineSnapshot(`
+		{
+		  "info": {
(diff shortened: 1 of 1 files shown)


lazerg:
The auto-fix check failure is unrelated to this change. It can't fetch the branch by name from this repo since the PR comes from a fork; the test matrix (macos/ubuntu/windows) all passed.

johnsoncodehk:
I changed this PR to be **test-only**.

The trailing-space approach has a regression: it makes the synthesized `foo.value` (9 chars) unmapped, so TypeScript silently drops the `TS18047` error.

Repro (`useTemplateRef` + `strictNullChecks`):

```vue
<template>
	<input ref="foo" />
	{{ foo.bar }}
</template>

<script setup lang="ts">
import { useTemplateRef } from 'vue'
const foo = useTemplateRef('foo')
</script>
```

Generated: `( foo.value .bar )`.

- without the trailing space: `TS18047 'foo.value' is possibly 'null'` + `TS2339` are both reported
- with the trailing space: `TS18047` disappears, because `toSourceRange(foo.value)` returns empty

So I removed the fix and kept the `#6176` test as a failing test that documents the rename bug, ready for a future fix to turn green. I also added two typecheck regression cases on `master` (`useTemplateRef-dynamic-arg` for the `TS2464` case, and `useTemplateRef-nullable` for the `TS18047` case).

PR #6178, opened 2026-08-22
fix(language-core): prevent template token boundary double-mapping

### LINKED ISSUE #6176: Renaming a useTemplateRef binding in :[foo] also renames the synthesized `.value`

johnsoncodehk:
Closes #6176

## Problem

Renaming `foo` at the `foo`/`]` boundary of `:[foo]` (with `useTemplateRef('foo')`) also renamed the synthesized `.value`, producing a spurious `Ref.value` reference in `@vue/reactivity`.

## Why

`foo` is a ref, so the template compiles `foo` to `foo.value` (`.value` is synthesized, no source mapping).

Rename requires a **unique** source→generated mapping — each source position maps to exactly one generated position. The `.value` gap breaks that: the cursor at `:[foo|]` maps to two generated positions, and rename runs at both:

```
[foo|.value]    → renames "foo"
[foo.value|]    → renames "value"   (spurious Ref.value)
```

## Fix

Rename is content-sensitive, so it must only fire on a **contiguous** mapped range (volarjs/volar.js#243); error mapping is position-based and may span discontiguous mappings. The synthesized `.value` is an unmapped gap that breaks that contiguity (and the uniqueness above), so the boundary character is downgraded to verification-only — it keeps error mapping but stops claiming the boundary for rename:

```ts
yield [code.slice(0, 1), source, offset, { verification: data.verification }];
```

The cut only happens where a previous token exists (`shouldCut = prevEnd > 0`), so the expression's leading character is preserved. `foo.bar` is unaffected — only the leading `.` is cut, `bar` stays renameable.

## After

Renaming `foo` now returns only `foo` (template + script).

### DIFF
--- a/packages/language-core/lib/codegen/template/interpolation.ts
+++ b/packages/language-core/lib/codegen/template/interpolation.ts
@@ -46,6 +46,7 @@ export function* generateInterpolation(
 				block.name,
 				start + prevEnd,
 				data,
+				prevEnd > 0,
 			);
 			yield `: `;
 		}
@@ -55,6 +56,7 @@ export function* generateInterpolation(
 				block.name,
 				start + prevEnd,
 				data,
+				prevEnd > 0,
 			);
 		}
 
@@ -103,6 +105,7 @@ export function* generateInterpolation(
 			block.name,
 			start + prevEnd,
 			data,
+			prevEnd > 0,
 		);
 	}
 
@@ -124,8 +127,13 @@ function* generateNonIdentifierCode(
 	source: string,
 	offset: number,
 	data: VueCodeInformation,
+	shouldCut = true,
 ): Generator<Code> {
 	if (!code.length) {
+		return;
+	}
+	if (!shouldCut) {
+		yield [code, source, offset, data];
 		return;
 	}
 	yield [code.slice(0, 1), source, offset, { verification: data.verification }];
--- a/packages/language-core/lib/codegen/template/interpolation.ts
+++ b/packages/language-core/lib/codegen/template/interpolation.ts
@@ -55,6 +55,7 @@ export function* generateInterpolation(
 				block.name,
 				start + prevEnd,
 				data,
+				true,
 			);
 		}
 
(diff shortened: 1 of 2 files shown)


packages/language-core/lib/codegen/template/interpolation.ts
```diff
@@ -98,19 +98,42 @@ export function* generateInterpolation(
...
 
+/**
+ * Yield a code chunk, cutting the first character off as verification-only.
+ *
+ * Adjacent mappings share the boundary offset (closed interval), so both the
+ * previous token's end and this chunk's start claim the same source offset.
+ * Downgrading the first character to verification-only keeps content-sensitive
+ * features (rename / navigation) from firing on the following chunk's boundary.
+ */
+function* yieldNonIdentifierCode(
```
KazariEX: For consistency.

```suggestion
function* generateNonIdentifierCode(
```

packages/language-core/lib/codegen/template/interpolation.ts
```diff
@@ -98,19 +99,51 @@ export function* generateInterpolation(
...
+	cutLast = false,
+): Generator<Code> {
+	if (!code.length) {
+		return;
+	}
+	if (cutLast) {
+		if (code.length > 1) {
+			yield [code.slice(0, -1), source, offset, data];
+		}
+		yield [code.slice(-1), source, offset + code.length - 1, { verification: data.verification }];
```
KazariEX:
```suggestion
		yield [code.slice(-1), source, offset + code.length - 1, codeFeatures.verification];
```

johnsoncodehk: What if data.verification === false?

KazariEX: My mistake, but it can be `data.verification ? codeFeatures.verification : { ... }` for less allocation.

PR #6195, opened 2026-08-30
refactor(language-core): move `__VLS_PropsChildren` into global template helpers

serkodev:
Related #6170

Fix for "remaining synthesized @ts-ignore" > JS typedef guards TS2503

The error is structural, but the `@ts-ignore` can live in the static `template-helpers.d.ts` instead of generated code. Behavior stays the same, and the content mapper never sees it.


KazariEX (approved):

serkodev:
Closing this PR: when `__VLS_PropsChildren` is moved into the global types, it leaks into the emitted dts.

PR #6198, opened 2026-08-31
feat(language-core): report duplicate CSS module names

serkodev:
Related #6170

Fix for "duplicate CSS module names report TS2300"


packages/language-core/lib/utils/parseSfc.ts
```diff
@@ -142,6 +142,7 @@ function createBlock(node: ElementNode, source: string) {
 				}
 				else if (p.name === 'module') {
 					block.__module = parseAttr(p, node);
+					block.__moduleAttrOffset = p.loc.start.offset - node.loc.start.offset;
```
KazariEX, line 145: Use `IRAttr#offset` instead.

